PR #133, opened 2025-12-05
Fix wizard transition race condition from double-click

pythagorakase:
## Summary

- **Guard in handleTraitConfirm**: Added early-return when `isLoading` is true to prevent duplicate API calls from double-clicking the Confirm button
- **Phase validation in handleArtifactConfirm**: Added check to ensure artifact type matches expected phase before processing, ignoring stale artifacts from previous phases

## Root Cause

Double-clicking the TraitSelector's Confirm button fired two simultaneous API calls with `current_phase="character"`. The first response advanced the wizard to the "seed" phase, but when the second response arrived (still with a character artifact type), the code incorrectly tried to process it as a seed artifact, calling transition with incomplete data.

Error: `"Missing: selected_seed, layer_draft, zone_draft, initial_location"`

## Test plan

- [ ] Complete wizard flow normally (single clicks)
- [ ] Attempt double-click on TraitSelector Confirm - should not produce duplicate API calls
- [ ] Verify transition completes successfully with seed data

### DIFF
--- a/ui/client/src/components/NewStoryWizard/InteractiveWizard.tsx
+++ b/ui/client/src/components/NewStoryWizard/InteractiveWizard.tsx
@@ -263,6 +263,7 @@ export function InteractiveWizard({
     };
 
     const triggerSubphaseContinuation = async (artifactType: string, contextData: any) => {
+        if (isLoading) return;
         setIsLoading(true);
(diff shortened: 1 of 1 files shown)


pythagorakase:
@claude please review this PR

PR #185, opened 2026-05-13
Route runtime API keys through macOS Keychain

pythagorakase:
## Summary

Migrates NEXUS runtime API key retrieval from `op` (1Password CLI subprocess) to **macOS Keychain via the `security` CLI**, eliminating biometric / session-timeout prompts during agentic workflows. 1Password remains the canonical source of truth — only `scripts/sync_secrets.py` invokes `op`, and only at bootstrap or rotation.

## Architecture

Every runtime caller (3 provider classes, the cloud router, 4 auxiliary scripts) now funnels through `nexus.util.secret_manager.get_secret(<provider>)`, which on macOS subprocesses to the Apple-signed `security` binary (no ACL prompts) and on other platforms falls back to the `keyring` Python library. Lookup order: `NEXUS_KEYRING_DISABLE` env-only escape hatch → Keychain → `<PROVIDER>_API_KEY` env var → `MissingSecretError`.

## :rotating_light: Action needed before merging

The branch is committed and the Anthropic key is already bootstrapped in Keychain (verified end-to-end via a live API call). **OpenAI and OpenRouter still need a one-time bootstrap**, which requires an interactive ``op signin``:

\`\`\`bash
op signin
poetry run python scripts/sync_secrets.py --provider openai openrouter
poetry run python scripts/sync_secrets.py --verify
\`\`\`

After that, the migration is fully active and runtime never touches \`op\` again.

## What changed

**New**
- \`nexus/util/secret_manager.py\` — single entry point: \`get_secret(provider) -> str\`.
- \`scripts/sync_secrets.py\` — bootstrap / rotation helper. Only place that invokes \`op\`.
- \`nexus.toml\` — new \`[secrets.providers]\` table mapping provider → \`op-item:...\` or \`op-read:...\` reference.

**Refactored** (\`_get_api_key\` bodies collapsed to \`return get_secret(\"...\")\`)
- \`scripts/api_openai.py\`, \`scripts/api_anthropic.py\`, \`scripts/api_openrouter.py\` — 30-line subprocess blocks → 2-line helper call each.
- \`nexus/llm/cloud.py\` — \`_resolve_api_key\` now goes through \`get_secret\`; \`is_available\` checks key presence via the helper instead of \`os.environ.get(api_key_env)\`. Fixes a latent bug where this path had no 1Password fallback at all.
- \`scripts/batch_metadata_processing.py\`, \`scripts/estimate_time_delta.py\` — dropped ~80 lines of \`.env\`-file and \`~/.zshrc\` parsing helpers.
- \`scripts/faction_former.py\`, \`scripts/freestyle_api_query.py\` — \`os.environ.get(\"OPENAI_API_KEY\")\` → \`get_secret(\"openai\")\`.

**Schema** (Pydantic)
- \`LLMCloudConfig.api_key_env\` removed (caller-side too, in \`router.py\`).
- \`SecretsConfig\` added at the root with \`providers: Dict[str, str]\`.

**Docs**
- \`CLAUDE.md\` "API Key Management" section rewritten.
- \`AGENTS.md\` "Security & Configuration Tips" line 19 rewritten.
- \`.claude/skills/update-1password-keys\` → \`.claude/skills/manage-api-keys\` with rewritten SKILL.md covering both Keychain (runtime) and 1Password (rotation) flows.

**Deleted** (obsolete workarounds)
- \`api_key_daemon.sh\` (60-min /tmp cache)
- \`extend_1password_session.sh\` (session extender)
- \`keep_1password_alive.sh\` (alive-pinger)

## Why the helper subprocesses to \`security\` rather than using the \`keyring\` Python library

The Python interpreter is ad-hoc-signed on macOS, so it isn't in any Keychain item's partition-list grant by default. When \`keyring.get_password()\` is invoked from Python, macOS issues a GUI ACL prompt that blocks unattended runs (this is what was causing the \"python3.11 wants to access key 'nexus-api'\" dialogs during development). Subprocessing to the Apple-signed \`/usr/bin/security\` bypasses the entire problem because that binary is unconditionally trusted by Keychain Services.

The \`keyring\` library is kept as a dependency only for the non-Darwin fallback path.

## Test plan

- [x] Real Anthropic API call via the full stack (\`AnthropicProvider\` → \`_get_api_key\` → \`get_secret\` → \`security\` subprocess → \`api.anthropic.com\` HTTP 200, masked log: \`sk-ant-...NQAA (len=108)\`).
- [x] \`NEXUS_KEYRING_DISABLE=1 OPENAI_API_KEY=dummy ... get_secret('openai')\` returns \`dummy\` without touching Keychain.
- [x] \`MissingSecretError\` raised with actionable remediation when no key exists.
- [x] Pydantic Settings validates the new \`[secrets.providers]\` block and the schema without \`api_key_env\`.
- [x] Silent retrieval timing — sub-second, no GUI prompt path.
- [ ] OpenAI / OpenRouter live calls (blocked on the one-time bootstrap above).
- [ ] \`nexus continue --slot N\` multi-turn — wait for OpenAI bootstrap first.

## Caveats discovered during the run

1. **One pre-existing test failure**: \`tests/test_llm_integration.py::test_router_with_real_backends\` fails when LM Studio isn't running AND no cloud key is reachable. Verified to fail identically on \`main\` (stashed changes, re-ran, same failure) — not a regression introduced by this PR. The test design treats backend exhaustion as failure rather than skip.
2. **\"python3.11 wants to access key nexus-api\" prompts during development**: my first iteration of \`secret_manager.py\` used \`keyring.get_password()\` directly, which triggers the Python-binary ACL prompt. I rewrote to subprocess \`security\` before committing — but stale \`__pycache__\` bytecode from the earlier iteration kept spawning prompts until I cleared the caches. The committed code path **does not** call \`keyring.get_password\` on macOS. If a prompt ever reappears, that's a regression / bypass bug.
3. **macOS login-keychain password mismatch**: separately surfaced during testing — the login keychain password and the macOS account password have drifted on your machine, so neither Touch ID nor your account password is accepted at Keychain dialogs. Unrelated to this PR but worth resyncing via Keychain Access.app → Edit → Change Password for Keychain \"login\" at your convenience.


pythagorakase:
@claude please review this PR. No need to flag issues already flagged by Codex's review. Target audience for your review is a local coding agent, not a human.

PR #188, opened 2026-05-13
Fix async structured LOGON generation

### LINKED ISSUE #187: Bootstrap fails: pydantic_ai.Agent.run_sync() in FastAPI's running event loop
## Summary

Wizard→narrative bootstrap fails reliably with a Pydantic validation error, but the validation error is a *secondary* symptom. Root cause is that `LogonUtility.generate_narrative()` (called from a FastAPI async route) invokes `pydantic_ai.Agent.run_sync()`, which calls `asyncio.run()` / `loop.run_until_complete()` internally — illegal inside an already-running event loop.

## Reproduction

Run on a fresh clean of slot 5, with the daemon up:

```bash
poetry run nexus clear --slot 5
poetry run nexus model --slot 5 --set gpt-5.5
poetry run nexus continue --slot 5                  # canned greeting
poetry run nexus continue --slot 5 --choice 1       # sci-fi
poetry run nexus continue --slot 5 --accept-fate    # subgenre
poetry run nexus continue --slot 5 --choice 4       # character archetype
poetry run nexus continue --slot 5 --accept-fate    # auto-name + traits
poetry run nexus continue --slot 5 --accept-fate    # confirm traits
poetry run nexus continue --slot 5 --choice 1       # wildcard
poetry run nexus continue --slot 5 --choice 3       # story seed
poetry run nexus continue --slot 5 --accept-fate    # ← bootstrap attempt fails here
```

The last call returns:

```
Error: 500: Failed to generate bootstrap narrative: 1 validation error for StorytellerResponseMinimal
choices
  Field required [type=missing, ...]
```

## Real chain of failures

Daemon log shows two errors per bootstrap attempt, ~38 s apart:

```
nexus.lore.logon - ERROR - Failed to get structured response: This event loop is already running
[... ~38 seconds of plain-text completion ...]
nexus.api.narrative_generation - ERROR - Bootstrap generation failed: 1 validation error for StorytellerResponseMinimal
```

1. **Root cause** — `scripts/api_openai.py:328-440` `OpenAIProvider.get_structured_completion()` uses `pydantic_ai.Agent` (model-agnostic harness, intentional design choice for the model picker) and calls `agent.run_sync(prompt)` at line 404. Inside FastAPI's async route, the event loop is already running, so `run_sync` raises `This event loop is already running`.

2. **Broken fallback** — `nexus/agents/lore/logon_utility.py:194-206`:

   ```python
   try:
       parsed_response, llm_response = self.provider.get_structured_completion(
           prompt, StorytellerResponseExtended
       )
       return parsed_response
   except Exception as e:
       logger.error(f"Failed to get structured response: {e}")
       response = self.provider.get_completion(prompt)  # plain text — ~38 s of LLM time
       return create_minimal_response(response.content)
   ```

3. **Invalid-by-construction fallback object** — `nexus/agents/logon/apex_schema.py:870-881`:

   ```python
   def create_minimal_response(narrative_text: str) -> StorytellerResponseMinimal:
       return StorytellerResponseMinimal(narrative=narrative_text)
   ```

   But `StorytellerResponseMinimal.choices` is `List[str]` with `min_length=2, max_length=4` — a **required** field. So this constructor is invalid the moment it's called. The Pydantic error is what surfaces to the user.

## Recommended fix

Tactical (1 line, smallest blast radius):
- Wrap `agent.run_sync()` in `asyncio.to_thread()` (and make `get_structured_completion` `async`), OR use `anyio.from_thread.run_sync` if we want to stay sync at the call site
- Effectively: run the synchronous pydantic_ai call on a worker thread so it gets its own event loop

Strategic (correct long-term):
- Add an async `agent.run()` path and make `LogonUtility.generate_narrative` async, propagating up through `narrative_generation.generate_bootstrap_narrative` (already async — good) and the LORE turn cycle.

Secondary cleanup either way:
- Fix or remove `create_minimal_response()`. As written it always produces an invalid Pydantic object. If we keep the fallback, it must populate `choices` with at least 2 strings (e.g. a generic `["Continue.", "Wait and observe."]`). Better: drop the fallback entirely and let the asyncio error surface — there's no upside to wasting ~30 s of LLM time on a path that always returns garbage.

## Context

- Surfaced during the keychain-secret-manager live verification — see PRs #185, #186.
- Migration from `responses.parse()` (OpenAI-specific) to `pydantic_ai.Agent` (model-agnostic) is correct and should be preserved. The bug is purely in the sync/async glue.

pythagorakase:
## Summary

Fixes #187 by moving LOGON structured narrative generation onto native async Pydantic AI calls instead of invoking `Agent.run_sync()` inside FastAPI's running event loop.

## What changed

- Added async structured completion APIs for both OpenAI and Anthropic providers using `await agent.run(...)`.
- Kept the existing sync structured APIs for script callers, but added running-loop guards that point callers to the async method.
- Added `LogonUtility.generate_narrative_async(...)` and routed bootstrap plus normal LORE Apex generation through it.
- Removed LOGON's structured-to-plain-text fallback so structured failures fail fast instead of doing a second doomed LLM call.
- Made `create_minimal_response(...)` schema-valid by adding two generic choices for non-LOGON coercion paths.

## Root cause

`LogonUtility.generate_narrative()` was called from async FastAPI/LORE paths but delegated to provider `get_structured_completion()`, which used `pydantic_ai.Agent.run_sync()`. That attempts to manage an event loop internally and fails when FastAPI already has one running. The old fallback then called plain-text completion and attempted to construct `StorytellerResponseMinimal` without required `choices`, surfacing a misleading Pydantic validation error.

## Validation

- `poetry run pytest tests/test_lore/test_logon_lazy_init.py tests/test_lore/test_apex_schema.py tests/test_lore/test_structured_provider_guards.py` -> 4 passed, 3 skipped
- `poetry run pytest tests/test_logon_mock_integration.py` -> 2 skipped
- `poetry run pytest tests/test_lore` -> 78 passed, 21 skipped

Manual slot-5 bootstrap acceptance was not run because it begins with `nexus clear --slot 5`, which would wipe live save-slot state.

### DIFF
--- a/nexus/agents/logon/apex_schema.py
+++ b/nexus/agents/logon/apex_schema.py
@@ -869,16 +869,22 @@ def validate_story_turn_response(data: Dict[str, Any]) -> StoryTurnResponse:
 
 def create_minimal_response(narrative_text: str) -> StorytellerResponseMinimal:
     """
-    Create a minimal response with just narrative text.
+    Create a minimal response from narrative text.
     Useful for fallback scenarios.
 
     Args:
         narrative_text: The narrative prose
 
     Returns:
-        StorytellerResponseMinimal with just narrative
+        StorytellerResponseMinimal with generic choices
     """
-    return StorytellerResponseMinimal(narrative=narrative_text)
+    return StorytellerResponseMinimal(
+        narrative=narrative_text,
+        choices=[
+            "Continue.",
+            "Wait and observe.",
+        ],
+    )
 
 
 def extract_narrative_text(response: StoryTurnResponse) -> str:
--- a/nexus/agents/lore/logon_utility.py
+++ b/nexus/agents/lore/logon_utility.py
@@ -15,12 +15,11 @@ import psycopg2
 # Add scripts directory to path for API imports
 sys.path.append(str(Path(__file__).parent.parent.parent.parent))
 
-from scripts.api_openai import OpenAIProvider, LLMResponse
+from scripts.api_openai import OpenAIProvider
 from scripts.api_anthropic import AnthropicProvider
 from nexus.agents.logon.apex_schema import (
     StoryTurnResponse,
     StorytellerResponseExtended,
-    create_minimal_response,
 )
 from nexus.config.loader import get_provider_for_model
 
@@ -183,8 +182,10 @@ class LogonUtility:
         """Public wrapper for provider initialization."""
         self._ensure_provider()
     
-    def generate_narrative(self, context_payload: Dict) -> StoryTurnResponse:
-        """Generate narrative from context payload with structured output"""
+    def generate_narrative(
+        self, context_payload: Dict[str, Any]
+    ) -> StoryTurnResponse:
+        """Generate narrative from context payload with structured output."""
         self._ensure_provider()
         # Format the context into a prompt
         prompt = self._format_context_prompt(context_payload)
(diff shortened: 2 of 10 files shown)


pythagorakase:
@claude please review this PR. The target audience for your review is a local coding agent, not a human.

Issue #194, opened 2026-05-14
ModelManager.load_model() crashes LM Studio model on verification (lmstudio==1.5.0)


pythagorakase:
## Summary

`ModelManager.load_model()` in `nexus/llm/model_manager.py` reliably crashes the LM Studio model process when called against `nexveridian/gpt-oss-120b` (the configured `default_model`). The crash happens during the post-load verification chat at lines 338-342. The same model loaded *manually* via LM Studio's UI works perfectly, including under the same 65,536 context. The same model serves inference fine via LM Studio's REST API. The early-return path in `ensure_default_model()` (when the model is already loaded externally) also works fine.

So the regression is narrow: **the `lms.llm(model_id, config={"contextLength": ...})` + post-load verification path in `load_model()` is what's crashing the model**, not anything about the model itself or about NEXUS's later use of the handle.

### NEXUS path fails

`model_manager.py:323-348`:

```python
self.current_model = lms.llm(model_id, config={"contextLength": context_window})
logger.info(f"Loaded with contextLength: {context_window}")
...
time.sleep(5)

logger.info("Verifying model...")
chat = lms.Chat("You are a test assistant.")
chat.add_user_message("Respond with 'OK'")

result = self.current_model.respond(chat, config={"maxTokens": 10})
```

Daemon log (chunk-2 test, 2026-05-13):

```
15:40:51,955 - model_manager - INFO - Loading model: nexveridian/gpt-oss-120b
15:40:51,955 - model_manager - INFO - Loading with context_window: 65536
15:41:12,269 - model_manager - INFO - Loaded with contextLength: 65536
15:41:17,275 - model_manager - INFO - Verifying model...
15:41:22,346 - model_manager - ERROR - Failed to load model nexveridian/gpt-oss-120b:
  Chat response error: The model has crashed without additional information. (Exit code: null)
15:41:22,347 - narrative_generation - ERROR - Error generating narrative:
  FATAL: Cannot initialize LM Studio SDK!
  ACTION REQUIRED: Start LM Studio and ensure a model is available
```

5 seconds between "Verifying model..." and "model has crashed" — the verification `respond()` call kills the LM Studio model process.

### REST API path works

Same model, same machine, same time:

```bash
curl -s -X POST http://localhost:1234/v1/chat/completions \
  -H "Content-Type: application/json" \
[...]

Issue #193, opened 2026-05-14
MEMNON._fetch_chunk_by_id references nonexistent chunk_metadata.place column


pythagorakase:
## Summary

`MEMNON._fetch_chunk_by_id` (at `nexus/agents/memnon/memnon.py:1444-1463`) executes SQL that references `chunk_metadata.place`, a column that does not exist in the current schema. Result: chunk fetching always fails on any slot, which kills `WARM_ANALYSIS` (and therefore any chunk-2+ generation).

This was hidden until #192 (the `self.logger` typo) is fixed; the broken exception handler turned every chunk-fetch failure into "AttributeError: 'MEMNON' object has no attribute 'logger'" instead of the real SQL error.

## The broken query

`memnon.py:1446-1461`:

```python
query = text("""
    SELECT
        nc.id,
        nc.raw_text,
        cm.season,
        cm.episode,
        cm.scene,
        cm.place,                                    -- ← does not exist
        nv.world_time,
        p.name as place_name
    FROM narrative_chunks nc
    LEFT JOIN chunk_metadata cm ON nc.id = cm.chunk_id
    LEFT JOIN narrative_view nv ON nc.id = nv.id
    LEFT JOIN places p ON cm.place = p.id            -- ← does not exist
    WHERE nc.id = :chunk_id
""")
```

Runtime error:

```
psycopg2.errors.UndefinedColumn: column cm.place does not exist
LINE 14:    LEFT JOIN places p ON cm.place = p.id
                                   ^
```

## Schema facts

`chunk_metadata` actual columns:

```
id, chunk_id, season, episode, scene, world_layer, time_delta, generation_date, slug
```

No `place` column. Never has been (or was removed long enough ago that no migration in `migrations/` reverts it).

Place ↔ chunk associations actually live in **`place_chunk_references`** (base table):

```
place_id, chunk_id, reference_type, evidence
```

And aggregated in **`chunk_places_view`** (per-chunk):

```
chunk_id, season, episode, scene, place_references
```

## Suggested fix

Decide which semantics are intended:

1. **One canonical place per chunk** — join through `place_chunk_references` filtered by some "primary" reference_type, or just pick the first one:

   ```sql
   LEFT JOIN place_chunk_references pcr ON nc.id = pcr.chunk_id
   LEFT JOIN places p ON pcr.place_id = p.id
   ```

2. **All places for the chunk** — use `chunk_places_view.place_references` and adjust the return shape.

[...]

PR #195, opened 2026-05-14
Fix narrative chunk continuation

### LINKED ISSUE #192: MEMNON.fetch_chunk typo + misleading wrapper error hide real turn-cycle failures
## Summary

Two related error-handling bugs in the narrative turn cycle compound to hide the real cause of chunk-2 generation failures from anyone debugging via CLI output:

1. **`nexus/agents/memnon/memnon.py:1483`** — exception handler references nonexistent `self.logger`, raising `AttributeError: 'MEMNON' object has no attribute 'logger'` *instead of* logging the actual fetch error.
2. **`nexus/api/narrative_generation.py`** — wraps any LORE-side failure as `"Error generating narrative: No narrative text in LORE response"`, which mis-attributes generation to LORE (LORE orchestrates; APEX generates) and obscures the real failure phase.

## Bug 1: `self.logger` typo

`memnon/memnon.py:107` defines a module-level logger:

```python
logger = logging.getLogger("nexus.memnon")
```

`memnon/memnon.py:1483` references it as an instance attribute that was never set:

```python
except Exception as e:
    self.logger.error(f"Error fetching chunk {chunk_id}: {str(e)}")
    return None
```

**Result**: when chunk fetching has any real error, the exception handler raises a *new* `AttributeError`, which propagates up through `turn_cycle.py` as "Failed to load requested chunk N: 'MEMNON' object has no attribute 'logger'". The actual underlying error is never logged.

**Fix**: drop `self.` — should be `logger.error(...)`.

Confirmed via `grep -rn "self\.logger" nexus/agents/memnon/`: this is the *only* offending reference; all other MEMNON code uses the module-level logger correctly.

## Bug 2: Misleading wrapper error

`narrative_generation.py` reports:

```
ERROR - Error generating narrative: No narrative text in LORE response
```

This is technically true (no narrative comes out of the LORE call) but semantically wrong because:
- LORE doesn't generate narrative; APEX does (via LOGON).
- The actual failure is usually in an earlier turn-cycle phase (WARM_ANALYSIS, MEMORY_RETRIEVAL, etc.) before APEX is reached.

Anyone debugging without the daemon log will think LORE is the problem; the real failure is somewhere upstream.

**Suggested fix**: include the actual turn-cycle phase / phase-error in the wrapper. The phase information is already available in `LORE.turn_context.error_log`.

## Reproduction

After PR #191 merges, on a fresh slot 5 with the model pre-loaded in LM Studio:

```bash
poetry run nexus clear --slot 5
poetry run nexus model --slot 5 --set gpt-5.5
# ... walk the wizard to bootstrap (see #189 repro) ...
poetry run nexus continue --slot 5 --user-text "anything"
# Error: No narrative text in LORE response
```

Daemon log:

```
nexus.lore.turn_cycle - ERROR - Failed to load requested chunk N:
  'MEMNON' object has no attribute 'logger'
nexus.lore - ERROR - Error in turn cycle phase TurnPhase.WARM_ANALYSIS:
  FATAL: No warm slice chunks retrieved.
nexus.api.narrative_generation - ERROR - Error generating narrative:
  No narrative text in LORE response
```

The *real* error (whatever made the `try` block in `_fetch_chunk_by_id` raise) is silently discarded.

## Context

- Surfaced during the chunk-2 live test on PR #191's branch (`fix-issue-189`).
- The bootstrap fix from #189 is what made this reachable — chunk 2 is the first time MEMNON's warm-slice retrieval activates in a fresh story.
- Once Bug 1 is fixed, we'll know what the *actual* underlying chunk-fetch error is and can file/fix that separately. Will update this issue with that finding.

pythagorakase:
## Summary
- fix MEMNON chunk fetch to use place references and preserve the real fetch error
- pass the active slot into LORE generation and surface phase failures before adapter coercion
- allow structured LOGON to emit partial new entity references, normalize descriptive place types, and retry one structured validation repair
- update incubator/commit conversion for the current reference schema, partial entity JSONB, and faction IDs without defaults
- avoid LM Studio chat verification for already-loaded models and extend CLI polling for long reasoning runs

Fixes #192, #193, #194.

## Validation
- `poetry run pytest tests/test_lore tests/test_api tests/test_llm_model_manager.py tests/test_cli.py`
- `poetry run pytest tests/test_db_converters.py tests/test_logon_mock_integration.py tests/test_new_story_schemas.py tests/test_summary_triggers.py`
- `poetry run python -m py_compile nexus/agents/logon/apex_schema.py nexus/agents/memnon/memnon.py nexus/api/narrative_generation.py nexus/api/lore_adapter.py nexus/api/commit_handler_sync.py nexus/api/db_converters.py nexus/cli.py nexus/llm/model_manager.py scripts/api_openai.py scripts/api_anthropic.py`
- `git diff --check`
- Live slot 5: generated chunk 2 from chunk 1 with `I cross the street and confront the watcher.`, then approved it; API committed narrative chunk 2 with metadata and entity/place/faction references.

### DIFF
--- a/nexus/agents/logon/apex_schema.py
+++ b/nexus/agents/logon/apex_schema.py
@@ -304,12 +304,14 @@ class NewPlace(BaseModel):
         if not isinstance(raw_type, str):
             return data
 
-        normalized_type = raw_type.strip().lower().replace("-", "_").replace(" ", "_")
-        if normalized_type in PlaceType._value2member_map_:
-            data["type"] = normalized_type
-            return data
-
         data = dict(data)
+        normalized_type = raw_type.strip().lower().replace("-", "_").replace(" ", "_")
+        try:
+            data["type"] = PlaceType(normalized_type).value
+            return data
+        except ValueError:
+            pass
+
         vehicle_terms = {"vehicle", "train", "car", "bus", "tram", "ship", "boat"}
         virtual_terms = {"virtual", "digital", "online", "network", "simulation"}
         if normalized_type in vehicle_terms:
@@ -681,6 +683,10 @@ class FactionStateUpdate(BaseModel):
     faction_name: Optional[str] = Field(
         default=None,
         description="Faction name (if ID unknown)"
+    )
+    current_activity: Optional[str] = Field(
+        default=None,
+        description="Current faction activity or operational focus"
     )
     recent_actions: Optional[List[str]] = Field(
         default_factory=list,
--- a/nexus/agents/logon/apex_schema.py
+++ b/nexus/agents/logon/apex_schema.py
@@ -114,13 +114,17 @@ class CharacterTraits(BaseModel):
         default=None, description="Debts, oaths, or duties you must honor"
     )
 
-    # Required wildcard - unique trait that sets this character apart
-    wildcard_name: str = Field(description="Name of the unique custom trait")
-    wildcard_description: str = Field(
+    # Optional wildcard - unique trait that sets this character apart
+    wildcard_name: Optional[str] = Field(
+        default=None,
+        description="Name of the unique custom trait"
+    )
+    wildcard_description: Optional[str] = Field(
+        default=None,
         description="What this trait means - capability, possession, relationship, or curse"
     )
 
-    model_config = ConfigDict(extra="forbid")
+    model_config = ConfigDict(extra="allow")
 
 
 class PlaceDetails(BaseModel):
@@ -175,7 +179,7 @@ class PlaceDetails(BaseModel):
         default_factory=list, description="Current rumors or gossip (max 3)"
     )
 
-    model_config = ConfigDict(extra="forbid")
+    model_config = ConfigDict(extra="allow")
 
 
 class FactionDetails(BaseModel):
@@ -201,7 +205,7 @@ class FactionDetails(BaseModel):
         default=None, description="Key traditions or rituals"
     )
 
-    model_config = ConfigDict(extra="forbid")
+    model_config = ConfigDict(extra="allow")
 
 
 # ============================================================================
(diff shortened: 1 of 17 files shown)


pythagorakase:
@claude please review this PR. No need to flag issues already flagged in Codex's review. The target audience for your review is a coding agent, not a human.

pythagorakase:
Spotter review note — single small thing flagged in an otherwise excellent PR:

**`nexus/llm/model_manager.py:169-172`** — the conditional after `_extract_loaded_identifiers` is a no-op:

\`\`\`python
loaded_identifiers = self._extract_loaded_identifiers(entries)
if loaded_identifiers:
    return loaded_identifiers
return loaded_identifiers
\`\`\`

Both branches return the same value, so the `if` does nothing. Looks like an unfinished edit — probably started with the intent to add divergent behavior (logging, a different fallback) and then thought better of it but didn't collapse the structure back down. Functionally correct, but visually misleading; a future reader will burn time wondering what the conditional was supposed to mean.

Suggested cleanup (one-line revert to pre-PR behavior at this site):

\`\`\`python
return self._extract_loaded_identifiers(entries)
\`\`\`

— or, if there *was* intended logging/handling that got dropped, just collapse to that.

Everything else looks good — verified per-issue:

| Filed | Outcome |
|---|---|
| #192 (logger + wrapper) | Logger typo fixed; `_describe_lore_failure()` surfaces real phase errors from `lore.turn_context.error_log` ✓ |
| #193 (SQL drift) | Properly joins through `place_chunk_references` with `string_agg(DISTINCT ... FILTER ...)` for the M:N case — better than my issue-body sketch ✓ |
| #194 (LM Studio crash) | Chat verification dropped; pre-load short-circuit added inside `load_model`; non-inference HTTP probe replaces inference verification ✓ |

Emergent scope (schema relaxation + place-type normalizer + `retries=0` → `retries=1` + adapter rework) all looks sound. Tests are stub/spy patterns rather than LLM-output mocks — `HandleThatMustNotVerify.respond()` raising `AssertionError` is a particularly nice negative-spy lock-in for the "no chat verification on load" invariant.

Approve once the dead-code blip is addressed (or with a one-sentence explanation that I missed the intent).

Issue #197, opened 2026-05-15
Chunk promotion discards user response: choice_text NULL, selected null, storyteller_text == raw_text


pythagorakase:
## Summary

When a chunk transitions from `incubator` to `narrative_chunks` (committed/finalized), three pieces of state that should be persisted are not:

1. `choice_text` is **NULL** — should hold the user's response text (selected choice text, or freeform `user_text`)
2. `choice_object.selected` is **`null`** — should hold the index of the selected structured choice (`1..N`)
3. `storyteller_text` equals `raw_text` — should be a strict subset (storyteller portion only)

These are almost certainly facets of one root cause in the chunk-promotion path.

## Design contract

For chunks generated in-system (slot 5 era, not legacy imports):

- `storyteller_text` — authoritative: storyteller's narrative portion only.
- `choice_text` — authoritative: user's response text (selected choice text, or freeform).
- `choice_object` — authoritative: `{"presented": [str, ...], "selected": <int 1..N | null>}`. Non-selected choices preserved underground for undo/audit; surface displays only the selected outcome.
- `raw_text` — **derived / surface only**: `storyteller_text || choice_text`. No unique information.

## Reproduction (slot 5, current state)

After the prior session generated chunk 1 (bootstrap) and chunk 2 (first storyteller turn) via the wizard → narrative transition, with the user responding to chunk 1 via `--choice 1` ("Step out from under the awning and confront the watcher across the street."):

```sql
SELECT id,
       length(raw_text) AS raw_len,
       length(storyteller_text) AS storyteller_len,
       length(choice_text) AS choice_len,
       choice_object->'selected' AS selected
FROM narrative_chunks
ORDER BY id;
```

```
 id | raw_len | storyteller_len | choice_len | selected
----+---------+-----------------+------------+----------
  1 |    1142 |            1142 |            | null
  2 |    7193 |            7193 |            | null
```

**Expected for chunk 1** (user answered with `--choice 1`):
- `raw_len` ≈ `storyteller_len + choice_len`
- `choice_len` ≈ length of "Step out from under the awning and confront the watcher across the street." (~73 chars)
- `selected` = `1`

## Impact

- Audit and undo cannot recover what the user picked
- The "surface entropy reduction" design (raw_text omitting non-selected choices) is broken because the decomposition (`storyteller_text` ≠ `raw_text`) was never actually persisted
- Display works after the companion slot_state fix lands, but only shows the *current* chunk's offered choices, not a history

## Suggested investigation

[...]

PR #198, opened 2026-05-15
Fix slot_state: surface choices on committed narrative chunks

pythagorakase:
## Summary

`slot_state._get_narrative_state` hardcoded `choices=[]` for committed chunks, even though `narrative_chunks.choice_object` actually preserves the structured choices on commit. This made `nexus load` (and the `GET /api/slot/{slot}/state` payload) show empty choices after any chunk acceptance, forcing users into `--user-text` instead of `--choice N`.

The fix: include `choice_object` in the SELECT and run it through `extract_presented_choices()` — the same helper the wizard path already uses (`slot_state.py:226`).

3 insertions, 4 deletions. Read-path only.

## How discovered

Prepping the embedding-pipeline smoke test (Phase A) on slot 5. The slot state API returned `"choices": []` for chunk 2 despite the underlying column holding all four structured options exactly as you'd expect.

## Related (NOT addressed here)

Issue #197 captures an adjacent write-path bug: `choice_object.selected` and `choice_text` are NULL on committed chunks. This PR only fixes the read path. Once #197 lands, full audit/undo becomes possible.

## Test plan

- [x] `psql -d save_05 -c "SELECT choice_object FROM narrative_chunks WHERE id=2;"` shows 4 presented choices
- [x] `curl http://127.0.0.1:8002/api/slot/5/state` returns the 4 choices (was `[]`)
- [x] `nexus load --slot 5` displays numbered choices block at end of output (was missing)

### DIFF
--- a/nexus/api/slot_state.py
+++ b/nexus/api/slot_state.py
@@ -301,7 +301,7 @@ def _get_narrative_state(cur) -> NarrativeState:
     # No pending content - get latest committed chunk
     cur.execute(
         """
-        SELECT nc.id, nc.raw_text
+        SELECT nc.id, nc.raw_text, nc.choice_object
         FROM narrative_chunks nc
         ORDER BY nc.id DESC
         LIMIT 1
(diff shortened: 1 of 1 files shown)


pythagorakase:

PR #204, opened 2026-05-15
Add nexus regenerate CLI verb with --note for soft steering

pythagorakase:
## Summary

- Adds `nexus regenerate --slot N` for re-rolling the last storyteller turn while it sits in the incubator. Discharges a deferred Phase A out-of-scope item.
- Adds optional `--note "..."` for soft out-of-character steering (e.g., `"darker, plz"`, `"continuity correction: the artifact was in Vienna, not Prague"`). Threaded all the way to the Apex prompt as a new `=== AUTHOR'S NOTE ===` section. Does **not** replace `user_text` — that's what `undo` is for.

## Why a new backend endpoint

The original plan pointed at `POST /api/story/regenerate` in `nexus/api/storyteller.py:575`, but that route lives on a separate FastAPI app that isn't deployed (the production daemon is `nexus.api.narrative:app` on port 8002, which never includes the storyteller app's routes). So this PR wires a real implementation:

- `POST /api/narrative/regenerate` looks up the singleton incubator row, captures `parent_chunk_id` and `user_text`, deletes the row, and re-invokes `generate_narrative_async` with the **same** `session_id` for replace-in-place semantics.
- The CLI then polls `/api/narrative/status/{session_id}` exactly like `run_continue` does.

## Soft steering channel (`note`)

A note is a single soft suggestion the player attaches to a regen — *not* a directive. Examples the user proposed: `"darker, plz"`, `"I want to win the fight despite my poor choices"`, `"hey you made a continuity error, the thing from 500 chunks ago was this, not that"`.

Plumbing:
- `RegenerateNarrativeRequest.note` → endpoint passes to background task → `generate_narrative_async(..., note=note)` → `lore.process_turn(..., note=note)` → stashed on `TurnContext.note` → included in `context_payload["note"]` → rendered as `=== AUTHOR'S NOTE ===` section in the Apex prompt via `logon_utility._format_context_prompt`.

Kept **separate** from the existing `authorial_directives` plumbing because that's a heavier mechanism — each directive triggers an extra MEMNON retrieval pass (`memory/manager.py:689`). A `note` is a single soft suggestion that bypasses retrieval entirely.

## Incidental fix

Surfaces `session_id` on `/api/slot/{slot}/state`. The field was already on `NarrativeState` (populated only while an incubator row exists — i.e., exactly the lifecycle window where regenerate is meaningful) but never made it into the API response schema. Independently useful for any slot-scoped CLI that needs the live session.

## Test plan

- [x] CLI `nexus regenerate --slot 5` succeeds end-to-end against an incubator-pending turn.
- [x] `nexus regenerate --slot 5 --note "darker, plz"` produces demonstrably darker output (body horror, uncanny callbacks, patient menace) while preserving the established narrative arc.
- [x] Same `chunk_id` and `session_id` after regen (replace-in-place verified).
- [x] Polling loop tolerates transient status-GET hangs without aborting the whole regen.

## Out of scope (queued)

- Narrative-mode `nexus undo` extension + drop of `chunk_embeddings_0384d` — separate follow-up PR (see plan file).
- Async-ifying the embedding subprocess (existing TODO at `chunk_workflow.py:388`).

### DIFF
--- a/nexus/agents/lore/logon_utility.py
+++ b/nexus/agents/lore/logon_utility.py
@@ -260,17 +260,6 @@ class LogonUtility:
         sections.append("\n=== USER INPUT ===")
         sections.append(context.get("user_input", ""))
 
-        # Add author's note (soft out-of-character suggestion, used by regenerate)
-        note = context.get("note")
-        if note:
-            sections.append("\n=== AUTHOR'S NOTE ===")
-            sections.append(
-                "The player is also leaving a soft, out-of-character suggestion for this "
-                "generation — treat it as authorial intent, not a hard constraint. It may "
-                "be a tonal nudge, a continuity correction, or an outcome preference:"
-            )
-            sections.append(note)
-
         # Add entity data with hierarchical support
         entity_data = context.get("entity_data", {})
         if entity_data:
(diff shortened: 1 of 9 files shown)


pythagorakase:

PR #205, opened 2026-05-16
Extend nexus undo for narrative grace window + drop chunk_embeddings_0384d

pythagorakase:
## Summary

Discharges Part 2 of the Phase A out-of-scope queue:

1. **`nexus undo` now closes the grace-window loop in narrative mode.** Previously it only deleted the incubator row, leaving the parent chunk's `choice_object.selected` and `choice_text` populated — the chunk recorded a "selected" choice the user had effectively un-made. Now the parent's choice fields are reset atomically alongside the incubator deletion.
2. **Drops the vestigial `chunk_embeddings_0384d` table.** Relic of an abandoned BGE-Small fine-tune. No production code reads from or writes to it (`DIMENSION_TABLE_MAP` has not contained `384` for some time), and it was the only embedding-shard table missing `ON DELETE CASCADE` on its FK — a sharp edge that has surfaced in three separate features.
3. **Pre-existing CLI bug fix:** `run_undo` was dropping the API's `message` field on failures, so the closed-grace-window case (and any "nothing to undo" from wizard mode) showed `"Error: Unknown error"`. Now mapped to `error` so users see the actual reason.

## Lifecycle context

The user-stated design: when chunk M sits in the incubator, M-1 is the mutable parent — committed but not yet ironman. The lifecycle places M-1 at `state=FINALIZED` with `embedding_generated_at IS NULL` (per memory: `accept_chunk(N)` finalizes N and fires embedding on N-1, so M-1's embedding only fires when chunk M is accepted).

The new `revert_pending_choice` enforces this invariant: it refuses to operate on an embedded chunk because mutating `raw_text` would invalidate the embedding's source text. If we ever hit that case, the design has drifted and should be surfaced rather than silently corrupting state.

## What changed

- `nexus/api/chunk_workflow.py`: New `revert_pending_choice(parent_chunk_id)` method. Single-transaction reset of `choice_text → NULL`, `choice_object.selected → JSON null` (preserves `presented`), `raw_text → storyteller_text`. Refuses on already-embedded chunks.
- `nexus/api/slot_endpoints.py`: `slot_undo_endpoint` narrative branch captures `parent_chunk_id` from the incubator row before deletion, then calls `revert_pending_choice` on it. Skipped for `parent_chunk_id == 0` (bootstrap chunk has no preceding choice).
- `nexus/cli.py`: `run_undo` surfaces failure messages via `error` field so they reach the user.
- `migrations/020_drop_chunk_embeddings_0384d.sql`: `DROP TABLE IF EXISTS chunk_embeddings_0384d;` (idempotent, applied to save_05 + NEXUS_template + save_02; saves 03/04 had pre-existing 018 drift unrelated to this PR).

## Test plan

- [x] **Open grace window**: With chunk 3 in incubator + chunk 2 carrying a selected choice (raw_text=7281, choice_text=86, selected=1), `nexus undo --slot 5` deletes incubator and resets chunk 2 (raw_text=7193 == storyteller_text exactly, choice_text NULL, selected null). CLI prints "Deleted pending content and reverted choice on chunk 2".
- [x] **Closed grace window**: Re-running `nexus undo --slot 5` against the now-empty incubator surfaces "Cannot undo: no live turn. The last response is already committed (the parent chunk is ironman / embedded)."
- [x] **Migration 020**: `chunk_embeddings_0384d` confirmed absent from save_05 and NEXUS_template. Migration is idempotent (`DROP TABLE IF EXISTS`).
- [x] All modified modules import clean.

## Out of scope (queued)

- Async-ifying the embedding subprocess (existing TODO at `chunk_workflow.py:388`).
- Reconciling the user's "FINALIZED and EMBEDDED are redundant" preference with the runtime usage at `chunk_workflow.py:244` — small follow-up.

### DIFF
--- a/migrations/020_drop_chunk_embeddings_0384d.sql
+++ b/migrations/020_drop_chunk_embeddings_0384d.sql
@@ -1,0 +1,17 @@
+-- 020_drop_chunk_embeddings_0384d.sql
+--
+-- Drop the vestigial 384-dimension embedding table. It is a relic of an
+-- abandoned BGE-Small fine-tune attempt; no production code reads from or
+-- writes to it. nexus.agents.memnon.utils.embedding_tables.DIMENSION_TABLE_MAP
+-- has not contained a 384 entry for some time. The table is also missing
+-- ON DELETE CASCADE on its FK to narrative_chunks(id) — the only such gap
+-- among the dimension-sharded tables — which would matter if any future
+-- feature ever deletes from narrative_chunks.
+--
+-- Discharges a recurring sharp edge that has surfaced in three separate
+-- features (issue #175 single-model consolidation, narrative undo design,
+-- and PR review of the embedding-pipeline overhaul).
+--
+-- Idempotent via DROP TABLE IF EXISTS — safe to re-run.
+
+DROP TABLE IF EXISTS chunk_embeddings_0384d;
--- a/nexus/api/chunk_workflow.py
+++ b/nexus/api/chunk_workflow.py
@@ -394,6 +394,72 @@ class ChunkWorkflow:
                     new_generation_triggered=True,
                 )
 
+    def revert_pending_choice(self, parent_chunk_id: int) -> None:
+        """
+        Reset the parent chunk's choice fields after the user undoes a pending turn.
+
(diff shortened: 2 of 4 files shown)


pythagorakase:

PR #207, opened 2026-05-16
Queue embedding generation after chunk acceptance

### LINKED ISSUE #206: Async-ify the synchronous embedding subprocess (chunk_workflow.py:542)
## Context

`ChunkWorkflow._trigger_embedding_generation` at `nexus/api/chunk_workflow.py:542` runs the embedding-generation worker as a synchronous subprocess with a 5-minute timeout:

```python
# TODO: Make async with FastAPI BackgroundTasks or Celery to avoid blocking
result = subprocess.run(
    [sys.executable, "scripts/regenerate_embeddings.py", "--chunk", ...],
    capture_output=True, text=True, timeout=300,
)
```

This is called from `accept_chunk(N)` (chunk_workflow.py:234), which fires embedding on N-1. The whole call chain blocks the async event loop for the duration of the subprocess — which, for Octen-Embedding-4B on a slow load, can be tens of seconds; worst case 5 minutes.

## Why it matters under the current lifecycle

The user's stated chunk lifecycle (see `project_embedding_lifecycle.md` in memory) treats the incubator as a one-chunk mutability buffer:
- Chunk M (just generated, in incubator) — mutable
- Chunk M-1 (one back) — committed but NOT yet ironman; embedding fires when M is accepted
- Chunk M-2+ — ironman, embedded

This timing presupposes that embedding fires *asynchronously* so the user isn't sitting on a frozen UI while M-2's embedding runs. Today, every `accept` blocks the next turn from being read. The sync subprocess is the bottleneck.

It also has secondary impact:
- Status-polling GETs from CLI's `run_continue` / `run_regenerate` hang while the event loop is busy with `subprocess.run` (PR #204 had to add per-iteration exception tolerance to the polling loop to survive this).
- Any other request hitting the daemon during embedding waits in line.

## Proposed shape

**Option A — FastAPI BackgroundTasks:** Move the embedding trigger to `background_tasks.add_task(...)` in the route handler that calls `accept_chunk`. The route returns immediately; the subprocess runs after. Single-process, no new infra. Risk: BackgroundTasks runs in the same uvicorn worker; a heavy embedding still ties up that worker (though it doesn't block the event loop because subprocess.run releases the GIL).

**Option B — Celery / RQ:** Push embedding work to a separate worker process via a queue. Cleanest separation; tolerates daemon restarts. New infra burden (broker, worker process).

Option A is the minimum-viable change and matches the existing TODO's framing. Option B is a richer follow-up if/when scaling matters.

## Out of scope

- Re-architecting the embedding pipeline itself (model selection, retries, etc.).
- Changing the lifecycle invariant — that's settled.
- The synchronous behavior is *not currently incorrect* (it produces correct results). This is a latency/concurrency issue, not a correctness one.

## Acceptance criteria

- `accept_chunk` returns to the caller before the embedding subprocess completes.
- A failed embedding doesn't leave the parent chunk in a broken state (clear the timestamp, log loudly).
- Existing CLI flows (`nexus continue`, `nexus regenerate`, `nexus undo`) work without timing assumptions changing visibly.
- The plan file's "Out of scope (queue for follow-up)" entry referencing this TODO can be deleted.

## References

- TODO: `nexus/api/chunk_workflow.py:542`
- Lifecycle memory: `project_embedding_lifecycle.md`
- Polling-loop workaround in PR #204: `nexus/cli.py:634-641`

pythagorakase:
## Summary
- queue chunk-acceptance embedding work through FastAPI BackgroundTasks instead of running the subprocess inline
- keep the existing synchronous workflow path for non-API callers, with stable job IDs for scheduled tasks
- clear incomplete embedding state on worker failure and cover the scheduler/failure behavior with unit tests

Fixes #206

## Verification
- poetry run python -m py_compile nexus/api/chunk_workflow.py nexus/api/narrative.py nexus/api/storyteller.py tests/test_api/test_chunk_workflow.py tests/test_api/test_narrative_generation.py
- poetry run python -m pytest tests/test_api tests/test_cli.py


pythagorakase:

PR #208, opened 2026-05-16
Retire ChunkState.EMBEDDED — embedding_generated_at is the ironman flag

pythagorakase:
## Summary

Drops the redundant `'embedded'` value from `narrative_chunks.state`. The timestamp `embedding_generated_at` was already the source of truth for ironman status; the state value duplicated it without carrying any unique information. Closes the user's "FINALIZED and EMBEDDED are redundant" cleanup item from the Phase A queue.

## What was redundant

- **Two writes** of `state='embedded'`: in `accept_chunk` (anticipatorily, after triggering embedding) and in `_trigger_embedding_generation`'s success path (after subprocess completion). Both happened in lockstep with `embedding_generated_at` being set.
- **Zero reads** of `state == 'embedded'`. The only state check on the previous chunk in `accept_chunk` was `prev_state == FINALIZED` as a "fire embedding on this one" gate — semantically equivalent to `embedding_generated_at IS NULL`.

## Changes

- `nexus/api/chunk_workflow.py`:
  - `ChunkState` enum: dropped `EMBEDDED`. Three values remain: `DRAFT`, `PENDING_REVIEW`, `FINALIZED`.
  - `accept_chunk`: previous-chunk SELECT now reads `(id, embedding_generated_at)`. The gate is `prev_embedded_at IS NULL`. The redundant `UPDATE state = 'embedded'` is gone.
  - `_trigger_embedding_generation` success path: UPDATE now only writes `embedding_generated_at`, not `state`.
- `ui/client/src/components/NarrativeTab.tsx`:
  - Dropped the `state === "embedded"` branch from the className merger — same Tailwind classes as `"finalized"`, so visually identical.
  - Removed `"embedded"` from the `ChunkWithMetadata.state` TypeScript union.
- `migrations/021_dedup_embedded_state.sql`:
  - Backfills existing rows: `UPDATE narrative_chunks SET state='finalized' WHERE state='embedded'`.
  - Refreshes the `state` column comment to drop the "synonymous with 'embedded' in practice" parenthetical (added in migration 018).

## Test plan

- [x] `ChunkState` enum no longer contains `EMBEDDED`; import smoke test passes.
- [x] Migration 021 applied to `save_05` + `NEXUS_template`: chunk 1 in save_05 correctly migrated from `state='embedded'` → `state='finalized'`, `embedding_generated_at` preserved.
- [x] Column comment refreshed.
- [x] `nexus continue --slot 5 --choice 1` exercises `accept_chunk`'s new embedding-gate predicate end-to-end. Post-state: zero chunks with `state='embedded'`; chunk 3 generated into incubator; chunk 1 still embedded via timestamp.

## Companion issue

[#206](https://github.com/pythagorakase/nexus/issues/206) — async-ifying the embedding subprocess. Filed separately; that work and this dedup are independent (this PR ships the dedup against today's sync subprocess; async-ify becomes interesting precisely because it opens a real `FINALIZED-but-subprocess-running` intermediate state worth tracking, which the dedup doesn't preclude).

### DIFF
--- a/migrations/021_dedup_embedded_state.sql
+++ b/migrations/021_dedup_embedded_state.sql
@@ -1,0 +1,23 @@
+-- 021_dedup_embedded_state.sql
+--
+-- Retire the redundant narrative_chunks.state value 'embedded'. The state
+-- 'embedded' carried the same information as embedding_generated_at IS NOT
+-- NULL — the timestamp is the single source of truth for ironman status,
+-- so the state value duplicated it.
+--
+-- This migration backfills existing rows (state='embedded' → 'finalized')
+-- and refreshes the column comment from migration 018 to drop the
+-- "synonymous with 'embedded' in practice" parenthetical. Application
+-- code (ChunkState enum + accept_chunk's prev-chunk guard) has been
+-- updated in the same PR to write 'finalized' and to check
+-- embedding_generated_at IS NULL instead of state == 'finalized'.
+--
+-- Idempotent: re-running on a database where no 'embedded' rows remain
+-- is a no-op UPDATE; COMMENT ON overwrites the prior comment.
+
+UPDATE narrative_chunks
+SET state = 'finalized'
+WHERE state = 'embedded';
+
+COMMENT ON COLUMN narrative_chunks.state IS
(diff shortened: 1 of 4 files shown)


pythagorakase:

PR #209, opened 2026-05-16
[codex] Use compound embedding table keys

### LINKED ISSUE #203: Use compound primary keys for chunk embedding tables
## Summary

Follow-up from PR #202: the dimension-specific embedding tables (`chunk_embeddings_1024d`, `chunk_embeddings_1536d`, `chunk_embeddings_2560d`, `chunk_embeddings_4096d`) should use the natural embedding identity as the primary key:

```sql
PRIMARY KEY (chunk_id, model)
```

The current `id SERIAL PRIMARY KEY` is functionally unnecessary and creates distracting sequence drift. In the live slot 5 test, the first `chunk_embeddings_2560d` row had `id = 2` and `chunk_id = 1`; that is harmless for retrieval, but it is a symptom that the schema is modeling the wrong identity.

## Why

An embedding row is uniquely identified by the narrative chunk plus the embedding model. The retrieval code joins on `chunk_id` and filters on `model`; it should not need a dedicated embedding-row surrogate ID.

Using a compound primary key would:

- remove sequence drift entirely
- make `ON CONFLICT (chunk_id, model)` the primary write path instead of a secondary unique constraint
- support multiple models per dimension table without pretending `id` can stay aligned with `chunk_id`
- let PR #202 drop the targeted row-id/sequence-alignment workaround

## Proposed Scope

- Add a migration that converts existing dimension tables to `PRIMARY KEY (chunk_id, model)`.
- Remove `id SERIAL PRIMARY KEY` from dynamic table creation in `scripts/regenerate_embeddings.py`.
- Update existing migration/schema definitions for 2560d/4096d tables so new DBs start with the compound key.
- Update SQLAlchemy embedding table models to use composite primary keys where they still declare an `id` column.
- Replace any embedding-table null checks like `e.id IS NULL` with `e.chunk_id IS NULL` or equivalent.
- Remove the sequence/id-alignment logic added for the targeted embedding path in PR #202.

## Migration Notes

Before adding the compound primary key, the migration should either assert or clean up duplicate `(chunk_id, model)` rows. Existing tables already have a unique `(chunk_id, model)` constraint in the generation path, so this should be low risk, but the migration should still be defensive.

If any old table has dependent constraints or indexes on `id`, drop those before dropping the column. Retrieval should continue to join via `chunk_id`.

## Acceptance Criteria

- No dimension embedding table requires a standalone `id` surrogate key.
- Targeted embedding upserts work with `ON CONFLICT (chunk_id, model)`.
- Missing-embedding detection no longer relies on `e.id`.
- Existing vector search / hybrid search still returns the same chunk IDs and metadata.
- Live targeted embedding on slot 5 inserts/updates `chunk_embeddings_2560d` without creating or relying on an ID sequence.

## Related

- Follow-up to PR #202.

pythagorakase:
## Summary

Fixes #203.

This migrates dimension-specific chunk embedding tables from a singleton `id` primary key plus unique `(chunk_id, model)` constraint to `(chunk_id, model)` as the primary key. It also makes embedding tables lazy-created by active embedding write paths, so fresh slots do not pre-create unused dimension tables.

## What changed

- Added migration `022_compound_embedding_pk_lazy_tables.sql`:
  - preserves non-empty embedding tables
  - fails fast on duplicate `(chunk_id, model)` pairs
  - drops `id` and old id sequences
  - drops empty pre-created embedding tables
- Added shared embedding table helpers for canonical naming, existence checks, listing, and lazy creation.
- Removed static MEMNON ORM embedding classes that were recreating `1024d`/`1536d` tables.
- Updated embedding regeneration/upsert code to use `(chunk_id, model)` only.
- Made `--only-indexes` inspect existing tables without creating missing ones.
- Fixed `scripts/propagate_schema.py` to match the live per-database `schema_migrations(version, name)` tracker.
- Made migrations `018` and `021` safe on older empty slots missing chunk-workflow columns.
- Refreshed embedding docs/tests for the lazy table model.

## Live migration notes

- Cloned slot 1 into slot 2 before migration.
- Preserved a backup of old slot 2 as `save_02_pre203_20260516_122902`.
- Explicitly unlocked slot 1 for migration, then relocked it afterward.
- Confirmed all slots and `NEXUS_template` are now migrated through `022`.
- Final live shape:
  - `NEXUS_template`, `save_03`, `save_04`: no embedding tables
  - `save_01`, `save_02`: `chunk_embeddings_1024d`, `chunk_embeddings_1536d`, no `id`, PK `(chunk_id, model)`
  - `save_05`: `chunk_embeddings_2560d`, no `id`, PK `(chunk_id, model)`

## Validation

- `poetry run python -m py_compile nexus/agents/memnon/utils/embedding_tables.py nexus/agents/memnon/utils/content_processor.py nexus/agents/memnon/utils/db_access.py nexus/agents/memnon/utils/alias_search.py nexus/agents/memnon/memnon.py nexus/agents/memnon/utils/db_schema.py scripts/regenerate_embeddings.py scripts/create_vector_index.py scripts/propagate_schema.py scripts/utils/embedding_utils.py scripts/query_narratives_vector.py`
- `poetry run pytest tests/test_ir_eval_v2/test_embedding_tables.py tests/test_api tests/test_cli.py`
- `NEXUS_RUN_POSTGRES=1 poetry run pytest tests/test_api/test_chunk_workflow_integration.py`
- Live lazy-create check from `NEXUS_template`
- `--only-indexes` missing-table check confirms no table is created by index maintenance

### DIFF
--- a/docs/vector_embeddings.md
+++ b/docs/vector_embeddings.md
@@ -22,25 +22,26 @@ NEXUS uses a multi-model ensemble approach with weighted combinations of differe
 
 ## Database Storage Strategy
 
-To handle different vector dimensions efficiently, we use a **dimension-specific tables approach**:
+To handle different vector dimensions efficiently, we use a **lazy
+dimension-specific tables approach**. Tables are named
+`chunk_embeddings_<dimensions>d` with four-digit padding for current model
+families, such as `chunk_embeddings_1536d`.
 
-1. **chunk_embeddings_1536d**
-   - Table for 1536-dimensional vectors (inf-retriever-v1-1.5b)
-   - Column type: `embedding Vector(1536)`
+Each table has the same logical shape:
 
-2. **chunk_embeddings_1024d**
-   - Table for 1024-dimensional vectors (E5-Large-V2, BGE-Large-EN)
-   - Column type: `embedding Vector(1024)`
-
-3. **chunk_embeddings_2560d**
-   - Table for 2560-dimensional vectors (Octen-Embedding-4B)
-   - Column type: `embedding Vector(2560)`
-
-4. **chunk_embeddings_4096d**
-   - Table for 4096-dimensional vectors (Octen-Embedding-8B)
-   - Column type: `embedding Vector(4096)`
+```sql
+chunk_id BIGINT NOT NULL REFERENCES narrative_chunks(id) ON DELETE CASCADE,
+model TEXT NOT NULL,
+embedding vector(<dimensions>) NOT NULL,
+created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+PRIMARY KEY (chunk_id, model)
+```
 
 This approach was chosen because pgvector requires fixed dimensions at table creation time, and mixing vectors of different dimensions in the same table would require significant padding and dimensionality management.
+
+Fresh slots do not pre-create every possible embedding table. Write paths call
+`ensure_embedding_table(...)` for the active model dimension when embeddings are
+generated. Read paths inspect existing tables and never create them.
 
 ## Query Routing System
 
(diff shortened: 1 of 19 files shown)


pythagorakase:

PR #231, opened 2026-05-17
Fix Claude review orchestration dispatch

pythagorakase:
## Summary
- Allow the dispatched Claude review workflow to run when initiated by `github-actions` from the orchestrator.
- Capture the dispatched Claude workflow run id so the orchestrator can verify downstream completion.
- Fail the orchestrator if the Claude workflow fails or is still incomplete instead of posting a misleading ready marker.

## Validation
- `git diff --check`
- `ruby -e 'require "yaml"; ARGV.each { |f| YAML.load_file(f); puts "ok #{f}" }' .github/workflows/agent-review-orchestration.yml .github/workflows/claude-code-review.yml`

Context: PR #229 did dispatch Claude, but the downstream Claude workflow failed because the action rejected the `github-actions` bot actor.

### DIFF
--- a/.github/workflows/agent-review-orchestration.yml
+++ b/.github/workflows/agent-review-orchestration.yml
@@ -75,11 +75,26 @@ jobs:
             exit 0
           fi
 
-          gh workflow run claude-code-review.yml \
-            --ref "$BASE_REF" \
-            -f pr_number="$PR_NUMBER" \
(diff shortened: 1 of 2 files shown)


pythagorakase:
@claude please review this PR. Don't flag issues already flagged by Codex. The target audience for your review is a coding agent, not a human.

PR #238, opened 2026-05-17
Skip review orchestration after initial request

pythagorakase:
## Summary
- treat an existing `nexus-agent-review:claude-request` marker as proof that the initial review window has started
- make `synchronize` events after that point exit as a fast no-op instead of waiting or dispatching Claude again
- keep `synchronize` useful before any review has started, so early pre-review pushes can still get orchestration

## Validation
- `actionlint .github/workflows/agent-review-orchestration.yml`
- `git diff --check`
- replayed the PR #237 comment-marker query and confirmed it would skip the post-review synchronize path

### DIFF
--- a/.github/workflows/agent-review-orchestration.yml
+++ b/.github/workflows/agent-review-orchestration.yml
@@ -86,13 +86,13 @@ jobs:
           fi
 
           if [[ "$EVENT_NAME" == "pull_request_target" && "$EVENT_ACTION" == "synchronize" ]]; then
-            ready_markers="$(
+            review_window_markers="$(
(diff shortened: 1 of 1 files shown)


pythagorakase:
Claude feedback note: the failed/cancelled downstream Claude run case is intentional for this PR. The invariant we want is one agent-review window per PR, not a second automatic review window after later synchronize/fix commits. If the dispatched Claude workflow fails, the orchestrate check should fail or withhold the ready marker, and a human/local agent can inspect the linked run and manually rerun the workflow/comment-trigger path. This keeps fix commits as fast no-ops while still making a failed initial review visible instead of silently trusting it.

PR #278, opened 2026-05-19
[codex] Add reranker tokenizer dependencies

### LINKED ISSUE #277: Fix MEMNON cross-encoder protobuf dependency failure
## Summary

During live Orrery/LORE validation, MEMNON hybrid retrieval completed but cross-encoder reranking failed and fell back to the original result order because the local environment is missing protobuf support for the configured reranker tokenizer/model path.

## Evidence

Observed during live slot 2 and slot 5 narrative turns on 2026-05-18:

```text
nexus.memnon.cross_encoder - ERROR - Error loading cross-encoder model:
 requires the protobuf library but it was not found in your environment.
```

The fallback continued cleanly:

```text
nexus.memnon.cross_encoder - WARNING - Returning original results due to reranking failure
```

The underlying traceback also included a tokenizer conversion failure:

```text
ValueError: Converting from SentencePiece and Tiktoken failed, if a converter for SentencePiece is available, provide a model path with a SentencePiece tokenizer.model file.
```

## Impact

- Not blocking narrative generation or Orrery validation.
- MEMNON retrieval still returns results.
- Cross-encoder reranking is effectively disabled in this environment, so result ordering may be lower quality than intended.

## Suggested next steps

1. Confirm whether `protobuf` should be a Poetry dependency for the current cross-encoder stack.
2. If adding `protobuf` is sufficient, add it and run the relevant MEMNON retrieval/reranker tests.
3. If tokenizer assets are also incomplete, document or fix the expected local cross-encoder model files, especially `tokenizer.model` for SentencePiece-backed models.
4. Consider degrading the log level after first failure per process if fallback is expected, so live runs are less noisy.

pythagorakase:
## Summary
- Add explicit `protobuf` and `sentencepiece` runtime dependencies for MEMNON's local DeBERTa/SentencePiece cross-encoder reranker.
- Add a dependency smoke test so the tokenizer extras fail loudly before a live rerank path falls back.

## Validation
- `poetry run pytest tests/test_memnon_cross_encoder_dependencies.py tests/test_memnon_db_access.py tests/test_lore/test_structured_provider_guards.py`
- Direct local reranker probe against `/Users/pythagor/nexus/models/naver-trecdl22-crossencoder-debertav3` on CPU; the model loaded and reranked the Pontchartrain passage above the intentionally higher-scored irrelevant passage.

## Notes
- No schema, settings, or data migration impact.
- The existing Transformers warning about SentencePiece byte fallback in the fast tokenizer is still emitted, but it is non-fatal and does not block model load/scoring.

Fixes #277.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1976,6 +1976,26 @@ pyyaml = ">=5.1"
 virtualenv = ">=20.10.0"
 
 [[package]]
+name = "protobuf"
+version = "6.33.6"
+description = ""
+optional = false
+python-versions = ">=3.9"
+groups = ["main"]
+files = [
+    {file = "protobuf-6.33.6-cp310-abi3-win32.whl", hash = "sha256:7d29d9b65f8afef196f8334e80d6bc1d5d4adedb449971fefd3723824e6e77d3"},
+    {file = "protobuf-6.33.6-cp310-abi3-win_amd64.whl", hash = "sha256:0cd27b587afca21b7cfa59a74dcbd48a50f0a6400cfb59391340ad729d91d326"},
+    {file = "protobuf-6.33.6-cp39-abi3-macosx_10_9_universal2.whl", hash = "sha256:9720e6961b251bde64edfdab7d500725a2af5280f3f4c87e57c0208376aa8c3a"},
+    {file = "protobuf-6.33.6-cp39-abi3-manylinux2014_aarch64.whl", hash = "sha256:e2afbae9b8e1825e3529f88d514754e094278bb95eadc0e199751cdd9a2e82a2"},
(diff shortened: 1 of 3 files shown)


pythagorakase:
Addressed Claude's actionable feedback in `efdd65ab`: split the dependency imports into independent tests without no-op asserts, tightened `protobuf` to `>=6.0.0,<7.0.0`, and added an optional local-model rerank path test that skips when the DeBERTa model directory is absent. Local validation: `poetry run pytest tests/test_memnon_cross_encoder_dependencies.py tests/test_memnon_db_access.py tests/test_lore/test_structured_provider_guards.py` -> 8 passed.

Issue #282, opened 2026-05-20
[Design] Package self-awareness: branch selection reads acting entity's own properties


pythagorakase:
## Concept

Packages today (and as described in the in-progress role taxonomy work) gate on tag conditions for **entry qualification** — "who can attempt this action." Branch selection within a firing package generally reads additional tag conditions of the same shape.

What's missing is a deliberate pattern for branch selection to read the **acting entity's own properties** as parameters to the package's decision-making. The acting entity is both the *subject* of the action AND a parameter influencing *which sub-action it chooses*.

Calling this **package self-awareness** — a package's stage-2 branch selection treats its own actor as input.

## Three Distinct Concerns (corrected)

When designing fame-dynamic packages, three stages should be kept architecturally separate:

| Stage | What it decides | What feeds it |
|---|---|---|
| **Stage 1: Entry gating** | Whether the package fires at all | Disposition + circumstance tags. **Fame is NOT a stage-1 factor.** A `cautious` NPC with `pursued_by(other)` qualifies for HIDE; a `reckless` one may not. |
| **Stage 2: Branch selection** | Which sub-action within the firing package | Acting entity's *own* properties — fame, disposition, resources. **This is the package-self-awareness layer.** |
| **Stage 3: Outcome / failure** | Whether the package succeeds | Per-tick encounter rolls weighted by ambient fame AND active pursuing-tag detection bonuses. PRNG, chunk-seeded for reproducibility. |

## Motivating Example: HIDE

Consider an NPC who qualifies for HIDE (stage 1: their disposition + pursue-state qualifies them). The package fires. Now stage 2 selects a branch based on the actor's own properties:

| Acting NPC properties | Likely HIDE branch |
|---|---|
| `role.fame:obscure` + `disposition.cautious` | `lay_low` (passive, minimal change) |
| `role.fame:known` + `role.resources:comfortable` | `change_appearance` (cosmetic alteration) |
| `role.fame:renowned` + `role.resources:wealthy` | `relocate` (move to safer location) |
| `role.fame:legendary` + `role.resources:magnate` | `extreme_cover` (identity erase, full vanish) |

The principle: high-fame characters can't lay low effectively (see fame-as-detection-radius design); they must take more dramatic measures or be found. Low-fame characters can ride out pursuit with minimal disruption.

## Stage 3: Encounter Collisions and Detection Radii

[...]

PR #286, opened 2026-05-20
Promote tag vocabulary doc from temp/ to docs/, drop inline versioning

pythagorakase:
## Summary

Moves `temp/orrery/orrery_vocabulary.md` (which lived in the gitignored `temp/` directory) to `docs/orrery_tag_vocabulary.md` so git carries the version history instead of inline commentary.

## What changed (besides location)

- Stripped the `v2 → v3` / `v3 → v4` change banners at the top
- Stripped `(Settled)` qualifiers from section headers
- Stripped `(Tonight)` brainstorm framing
- Consolidated `## Out of Scope (Tonight)` items into `## Open Items` (4 still-relevant items absorbed; 2 completed/superseded items dropped)
- Renamed `## Dropped From V2 During Brainstorm` → `## Dropped / Rejected Vocabulary`
- Refreshed `## Related Artifacts` (added PRs #283, #284, issue #282; dropped stale "needs update post-vocabulary work" notes)

The squashed version history (v2 origin → v3 settled foundations → v4 vocabulary expansion + design alignment) lives in the commit message of this initial commit at the new location, so future readers can find it via `git log --follow docs/orrery_tag_vocabulary.md`.

## What didn't change

All substantive content — Architectural Foundations, all five character categories (bodyform, disposition, capacity, role, state-pending), Multi-Entity Tags, Place / Faction categories, Slot 2 Reference Taggings, trait_menu ↔ tag-vocabulary alignment, Dropped / Rejected list. This PR is purely a location + versioning-hygiene change.

## Test plan

- [ ] Visual diff of removed banners + section header changes (nothing of substance dropped)
- [ ] Confirm no broken intra-doc references after the section renames (\"Open Items\" still referenced from a couple places)
- [ ] Confirm `temp/orrery/orrery_vocabulary.md` was indeed gitignored and not previously tracked (no git history loss)
- [ ] `git log --follow docs/orrery_tag_vocabulary.md` shows the commit with the full squashed history in the message

### DIFF
--- a/docs/orrery_tag_vocabulary.md
+++ b/docs/orrery_tag_vocabulary.md
@@ -13,14 +13,14 @@ Closed-vocabulary discipline: tags do not grow at runtime. Every candidate is fi
 | Structure | Application table | Carries | Example |
 |---|---|---|---|
 | **Single-entity tag** | `entity_tags` (existing) | Property of one entity | `bodyform:cyborg` on Alex |
-| **Multi-entity tag** | `entity_pair_tags` (new, proposed) | Binary property of an ordered pair | `knows_location(Alex → Burrow)` |
+| **Multi-entity tag** | `entity_pair_tags` (shipped — see PR #283) | Binary property of an ordered pair | `knows_location(Alex → Burrow)` |
 | **Rich relationship** | `character_relationships` (existing) | Affective/social state with valence, history, sub-states | Alex/Emilia trust=high, valence=positive |
 
 ### Locked Vocabulary — No Runtime Growth
 
-Skald applies from a closed registry. The `skald_inline` and `auto_registered` source_kind paths can be deprecated. The Vocabulary Growth Contract section in the design plan should be removed.
+Skald applies from a closed registry. The `skald_inline` and `auto_registered` source_kind paths are deprecated; the `new_tag_proposals` runtime path is removed (tracked as issue #287). Bestowal of an unknown tag name is a hard error, not an opportunity to auto-register.
 
-**Rationale:** every successful tag system in adjacent design space (Bethesda Radiant AI, Dwarf Fortress) uses closed vocabularies. NEXUS's earlier "let Skald propose at runtime" gambit produces entropy without curation; the right discipline is pre-design.
+**Rationale:** every successful tag system in adjacent design space (Bethesda Radiant AI, Dwarf Fortress) uses closed vocabularies. NEXUS's earlier "let Skald propose at runtime" gambit produced entropy without curation; the right discipline is pre-design.
 
 **Implications:**
 - Vocabulary must be comprehensive enough for cross-genre stories.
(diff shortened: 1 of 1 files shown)


pythagorakase (commented):
Codex mechanical review on current head `1ac3a760`.

I think this doc is directionally strong as the design target, and the follow-up commit has already handled most of Claude's stale doc-vs-code markers. My main caution: downstream implementers should not read this as a live runtime contract yet. A few mechanical implications need to be made explicit before the trait compiler / resolver rewrite starts leaning on it.

1. **Closed vocabulary vs runtime writer is the biggest unresolved invariant.** The doc opens with locked vocabulary / no runtime growth, but the `status:<level>` section still says vocabulary may expand through `new_tag_proposals`. Current schemas, prompts, and `tag_writer.apply_tag_bestowal()` still auto-register proposals and silently skip unknown registered names. That needs one decisive contract: either status expansion is curated migration work, or runtime proposals remain allowed in a quarantined/audit-only path. I would not keep both in the spec.

2. **Exclusive categories need a write primitive, not just registry metadata.** `role.fame`, `role.resources`, visibility/access categories, etc. want XOR semantics. The current writer applies named tags and clears only explicitly listed tags. The compiler/writer will need an atomic “replace within category/family” operation that clears siblings before applying the new value. For `status:*`, consider whether there can be only one status level per subject→scope faction edge, even though the overall family is multi-valued across different scope factions.

3. **`status:<level>` should not make package authors parse strings ad hoc.** Exact `has_pair_tag()` is enough for `hunting`, `protects`, etc., but status gates will want “any status in this family”, rank thresholds, and negative-status checks. That implies either pair-tag metadata columns (`family`, `rank`, `valence`) or a centralized status helper that owns the mapping. Otherwise every package will reinvent `tag.startswith('status:')` and drift.

4. **Composition/decomposition needs auditable failure modes.** The Wildcard row's “falls back to prose-only storage if decomposition fails” is the right safety valve, but it should be visible. The compiler should produce structured results: applied single tags, applied pair tags, created entities, relationship rows, and unresolved/prose-only remainders with reason codes. Silent prose fallback would make a trait look accepted while losing all gates.

5. **Pair tags vs rich relationships need a source-of-truth rule.** Allies / Contacts / Enemies deliberately write both pair tags and `character_relationships`. That is mechanically useful, but it can drift. I would codify: pair tag = typed mechanical edge packages can gate on; rich relationship = affective/history/valence layer Skald and social logic read. Then add reconciliation/idempotency tests for compiler output.

6. **The place/faction refactor is a separate migration/template rewrite.** The resolver still hydrates `place_affordance` into `location_classes`, and existing predicates/templates consume that surface. Moving to `place_function`, `place_visibility`, `place_access`, etc. needs an adapter or staged rewrite; otherwise older `in_location_class(...)` gates will quietly stop matching.

7. **Fame needs a scope decision.** The doc says fame is ambient detection radius, but several examples read as subculture-local fame. If `renowned` means “renowned among netrunners,” then gates need an audience/scope dimension. If fame is truly global, local renown should probably compile to scope-bound `status:*` or another scoped relation instead.

Suggested implementation order: lock the vocabulary-growth contract first; add cardinality/exclusive write behavior second; add pair-tag family helpers/metadata for status third; then do the trait compiler with explicit compile-result auditing. Place/faction tag migration feels like its own later chunk, not part of the first character-trait compiler pass.

PR #315, opened 2026-05-23
Add Orrery entity tag clear helper

### LINKED ISSUE #313: Orrery: add public clear_entity_tag sibling for event-driven single-entity tag clearance
## Problem

The pair-tag side has a public `clear_pair_tag` designed for event-driven clearance:

- `nexus/agents/orrery/tag_writer.py:519` — public function, takes tag *name*, JOINs through registry
- Used to retire ephemeral relations when their context no longer holds (e.g., `pursuing` when pursuers give up)

The single-entity side has only a **private** helper:

- `nexus/agents/orrery/tag_writer.py:300` — `_clear_entity_tag(cur, entity_id, tag_id)`
- Underscore prefix; only called internally by `apply_tag_bestowal` for replace-semantics during bestowal
- Takes pre-resolved `tag_id`, not the tag name
- No public surface for event-driven clearance from outside the writer module

## Why this is a blocker

The `state` character vocabulary category (separate drafting session in flight) needs to declare a subset of state tags as **event-clearable** (e.g., `state:wounded` cleared by a healing event). Building clearance policy on the assumption that the single-entity clearance machinery exists publicly would mean spec'ing against infrastructure that doesn't yet exist.

The pair-tag side has its event-clearable surface; single-entity does not. This asymmetry is silent — `_clear_entity_tag` exists, so a casual reader assumes a public path is there too. It isn't.

## Required

Public `clear_entity_tag` in `nexus/agents/orrery/tag_writer.py` mirroring `clear_pair_tag`'s shape:

```python
def clear_entity_tag(
    cur: Any,
    *,
    entity_id: int,
    tag: str,  # name, not id — JOIN through tags registry table
) -> bool:
    """Mark an active single-entity tag as cleared (UPDATE cleared_at = now()).
    
    Returns True if a row was cleared, False if no active row existed for
    (entity_id, tag). Unknown / deprecated tag names silently return False —
    same asymmetric convention as clear_pair_tag vs apply_tag_bestowal.
    """
```

Semantic contract mirrors `clear_pair_tag`:

- Look up tag via JOIN to `tags` registry table (caller passes tag **name**, not id)
- Unknown / deprecated tag silently returns `False` — clearing a tag that doesn't exist is a no-op, indistinguishable from a missing row
- Returns `True` if a row was cleared, `False` if no active row existed

## Out of scope (deliberate)

- **`tag_clearance_log` entry**: the table exists (referenced from `world_events` via `triggering_event_id` FK), but `clear_pair_tag` doesn't write a log entry either. Out of scope here; if audit-log writing is needed, file a follow-up that wires both sides symmetrically.
- **Promoting `_clear_entity_tag_siblings` to public**: also internal-only; leave as-is unless the trait→tag compiler explicitly needs it.

## Acceptance

- Public `clear_entity_tag` function in `nexus/agents/orrery/tag_writer.py`
- Mirrors `clear_pair_tag`'s signature, docstring style, and silently-returns-False asymmetry
- Internal `_clear_entity_tag` can either stay (still used by `apply_tag_bestowal` replace-semantics — line 121) or the public function can be its replacement; implementer's call based on what makes the bestowal path cleanest
- Unit tests:
  - Clearing a known active tag returns `True` and sets `cleared_at`
  - Re-clearing the same tag returns `False` (idempotent no-op)
  - Clearing an unknown / deprecated tag name returns `False` silently (no exception)

## References

- Existing pair-tag side: `nexus/agents/orrery/tag_writer.py:519` (`clear_pair_tag`)
- Existing private single-entity helper: `nexus/agents/orrery/tag_writer.py:300` (`_clear_entity_tag`)
- Spec context: unblocks the `state` character vocabulary work being drafted in a separate session

### LINKED ISSUE #314: Migration 043: fix COMMENT ON COLUMN syntax error blocking migration queue
## Problem

`migrations/043_orrery_category_refactor_phase1.py` ships with two PostgreSQL syntax errors that block the entire migration queue. `python scripts/migrate.py --all` fails on every unlocked slot at line 134 of 043:

```
ERROR FAILED: 043_orrery_category_refactor_phase1 - syntax error at or near "||"
LINE 4:                 || 'until a data migration rewrites them.';
```

Same pattern at line 141.

## Root cause

`COMMENT ON COLUMN x.y IS` requires a **single string literal**; the `||` SQL string-concatenation operator can't appear in that position. Two instances:

```python
# Lines 130-135
cur.execute(
    """
    COMMENT ON COLUMN tag_category_registry.deprecated IS
        'Category-level cutover marker; existing tag rows remain live '
        || 'until a data migration rewrites them.';
    """
)

# Lines 137-142
cur.execute(
    """
    COMMENT ON COLUMN tag_category_registry.replacement_categories IS
        'Preferred successor categories for deprecated category rows, '
        || 'when any exist.';
    """
)
```

## Fix

Drop the `||` (Postgres auto-concatenates adjacent string literals separated by whitespace + newline):

```sql
COMMENT ON COLUMN tag_category_registry.deprecated IS
    'Category-level cutover marker; existing tag rows remain live '
    'until a data migration rewrites them.';
```

Or use a single multi-line literal — implementer's choice.

## Impact

- 043 is on `main` but **applied nowhere** (verified via `python scripts/migrate.py --status` — every unlocked slot shows `[ ] 043` pending)
- 044 (`disambiguate_status_reputation_traits` from #311) is blocked behind 043
- All future migrations are blocked until 043 is fixed
- Since 043 hasn't been applied anywhere, the fix is safe to make in-place (no schema-rewrite-of-applied-migration concerns)

## Acceptance

- Both `||` instances in `migrations/043_orrery_category_refactor_phase1.py` replaced with valid `COMMENT ON COLUMN ... IS '<literal>'` syntax
- `python scripts/migrate.py --all` applies 043 + 044 cleanly across all unlocked slots (slot_01 is locked and will be skipped, as expected)
- Verify resulting comment text via `\d+ tag_category_registry` on save_02 after apply

pythagorakase:
## Summary
- Add public `clear_entity_tag` for event-driven single-entity tag clearance by tag name.
- Keep clear semantics no-op for missing, inactive, unknown, or deprecated tag names.
- Fix migration 043 `COMMENT ON COLUMN` syntax so 043 and queued 044 can apply.

## Validation
- `poetry run flake8 nexus/agents/orrery/tag_writer.py tests/test_orrery/test_tag_writer.py migrations/043_orrery_category_refactor_phase1.py tests/test_orrery/test_migrate.py`
- `git diff --check`
- `poetry run pytest tests/test_orrery/test_tag_writer.py tests/test_orrery/test_migrate.py`
- `poetry run python scripts/migrate.py --all`
- `psql -d save_02 -Atc "SELECT ... col_description(...) ..."`
- `poetry run pytest tests/test_orrery`
- `poetry run pytest`

## Schema / Config / Data
- Schema migration 043 is fixed in place because it was unapplied before this branch.
- Local migration run applied 043 and 044 to `NEXUS_template` and `save_02` through `save_05`; locked `save_01` was skipped as expected.
- No settings/config changes.

Closes #313
Closes #314

### DIFF
--- a/nexus/agents/orrery/tag_writer.py
+++ b/nexus/agents/orrery/tag_writer.py
@@ -307,9 +307,13 @@ def clear_entity_tag(cur: Any, *, entity_id: int, tag: str) -> bool:
     than raising — this is the inverse of ``apply_tag_bestowal``, which raises
     ``ValueError`` for unknown tags. The asymmetry matches ``clear_pair_tag``:
     clearing a tag that doesn't exist is a no-op, so an unknown tag is
-    indistinguishable from a missing row.
+    indistinguishable from a missing row. Alias names in ``CANONICAL_TAGS``
+    resolve to their canonical tag before clearing. Deprecated registry rows
+    are intentionally treated as absent; clearing pre-deprecation active rows is
+    a data-cleanup concern for a migration or id-based maintenance helper.
     """
 
+    canonical_name = CANONICAL_TAGS.get(tag, tag)
     cur.execute(
         """
         UPDATE entity_tags et
@@ -322,7 +326,7 @@ def clear_entity_tag(cur: Any, *, entity_id: int, tag: str) -> bool:
           AND t.synonym_for IS NULL
           AND et.cleared_at IS NULL
         """,
-        (entity_id, tag),
+        (entity_id, canonical_name),
     )
     return bool(cur.rowcount)
 
--- a/tests/test_orrery/test_tag_writer.py
+++ b/tests/test_orrery/test_tag_writer.py
@@ -379,7 +379,7 @@ def test_clear_entity_tag_marks_known_active_tag_cleared():
     assert cur.cleared_keys == [(11, tag_id)]
 
 
-def test_clear_entity_tag_returns_false_when_no_active_row():
+def test_clear_entity_tag_is_idempotent():
     tag_id = hash("scrying_target") & 0xFFFFFF
     cur = FakeCursor(
         tags=_registered(("scrying_target", "orrery_state", True)),
@@ -396,6 +396,33 @@ def test_clear_entity_tag_returns_false_when_no_active_row():
 
     assert clear_entity_tag(cur, entity_id=11, tag="scrying_target") is True
     assert clear_entity_tag(cur, entity_id=11, tag="scrying_target") is False
+
+
+def test_clear_entity_tag_returns_false_when_no_active_row():
+    cur = FakeCursor(tags=_registered(("scrying_target", "orrery_state", True)))
+
+    assert clear_entity_tag(cur, entity_id=11, tag="scrying_target") is False
+
+
+def test_clear_entity_tag_resolves_canonical_alias():
+    tag_id = hash("corporate_exile") & 0xFFFFFF
+    cur = FakeCursor(
+        tags=_registered(("corporate_exile", "state", False)),
+        entity_tags=[
+            {
+                "entity_id": 42,
+                "tag_id": tag_id,
+                "applied_at_world_time": _WORLD_TIME,
+                "source_kind": "skald_inline",
+                "cleared_at": None,
+            }
+        ],
+    )
+
+    cleared = clear_entity_tag(cur, entity_id=42, tag="corporate_defector")
+
+    assert cleared is True
+    assert cur.entity_tags[0]["cleared_at"] == "now"
 
 
 def test_clear_entity_tag_unknown_or_deprecated_tag_is_noop():


pythagorakase:
Follow-up pushed for the review notes:

- `clear_entity_tag` now resolves aliases from `CANONICAL_TAGS` before joining through `tags`, matching the application path for static aliases.
- The idempotency test was renamed and a true no-active-row test was added.
- Added explicit docstring wording for the deprecated-tag edge: per #313, deprecated registry rows are treated as absent by this public API; clearing pre-deprecation active rows is left to a migration or id-based maintenance helper rather than event-driven clearance.

Validation after the follow-up:
- `PYTHONPATH=. poetry run pytest tests/test_orrery/test_tag_writer.py tests/test_orrery/test_migrate.py`
- `PYTHONPATH=. poetry run flake8 nexus/agents/orrery/tag_writer.py tests/test_orrery/test_tag_writer.py migrations/043_orrery_category_refactor_phase1.py tests/test_orrery/test_migrate.py`
- `git diff --check`
- `PYTHONPATH=. poetry run pytest tests/test_orrery`
- `PYTHONPATH=. poetry run pytest`

Issue #337, opened 2026-05-25
Orrery: migrate obsolete factions table columns into faction vocabulary substrate


pythagorakase:
## Context

PR #336 drafts `docs/orrery_faction_vocabulary.md`, which makes several legacy columns on the database `factions` table structurally obsolete. The vocabulary wants faction properties to live in closed Orrery tag categories and relationships instead of ad hoc columns.

The docs identify the target shape:

- Move to single-entity faction tags:
  - `factions.ideology` -> `ideology`
  - `factions.power_level` -> `power_status`
  - `factions.hidden_agenda` -> `agenda`
  - `factions.resources` -> `resource_base`
- Move/drop out of columns entirely:
  - `factions.history` -> `world_events` / summary prose
  - `factions.current_activity` -> usually `agenda`; otherwise prose
  - `factions.territory` -> `claims(faction -> place)` and/or `operates_from(faction -> place)`
- Keep as columns:
  - `id`, `name`, `entity_id`, `summary`, `primary_location`, timestamps, `extra_data`

## Problem

The code still treats these fields as first-class columns in faction creation/update paths. Examples found during the doc pass:

- `nexus/api/db_converters.py::create_new_faction()` inserts `ideology`, `history`, `current_activity`, `hidden_agenda`, `territory`, `power_level`, and `resources`.
- `nexus/api/commit_handler_sync.py::create_new_faction_sync()` does the same.
- `nexus/api/commit_handler.py` and `nexus/api/commit_handler_sync.py` still update `factions.current_activity` from structured state updates.
- `nexus/agents/logon/apex_schema.py::NewFaction` still exposes those legacy faction fields and describes older Orrery tag categories.

Once the faction vocabulary is accepted/seeded, keeping these columns active creates a split-brain source of truth: Skald and UI paths can write one semantic layer while Orrery reads another.

## Proposed work

This should be a staged migration, not a blind column drop.

1. **Preflight / audit**
   - Inspect live/test slots for non-null values in the soon-obsolete columns.
   - Produce a dry-run mapping for existing values into faction tags, pair-tags, world_events/prose, or structured remainder.
   - Flag ambiguous cases, especially `resources`/`resource_class:network`, where the vocabulary draft says manual review may be needed.

2. **Seed/ensure target substrate**
   - Ensure accepted faction vocabulary tags are registered before data rewrite.
   - Ensure pair-tag targets exist for `claims(faction -> place)` and `operates_from(faction -> place)`.

[...]

PR #338, opened 2026-05-26
[codex] Add faction table migration audit

pythagorakase:
## Summary
- add a read-only `nexus faction-audit --slot N` dry run for legacy `factions` column migration
- map legacy faction columns and old faction tag categories into reviewable Orrery tag, pair-tag, event/prose, or remainder candidates
- document the preflight command in the faction vocabulary cleanup section

## Validation
- `poetry run pytest tests/test_faction_table_audit.py tests/test_cli.py -q`
- `poetry run flake8 nexus/api/faction_table_audit.py nexus/cli.py tests/test_faction_table_audit.py`
- `poetry run mypy nexus/api/faction_table_audit.py tests/test_faction_table_audit.py`
- `poetry run python -m nexus.cli faction-audit --slot 2`
- `poetry run python -m nexus.cli faction-audit --slot 5`

## Data impact
- Read-only audit only; no data rewrite, schema migration, or runtime write-surface cleanup yet.
- Slot 2 dry run reports 9 factions scanned, 57 non-null legacy values, 28 entity-tag candidates, 16 pair-tag candidates, and 79 manual-review items.
- Slot 5 dry run reports 4 factions scanned, 4 non-null legacy values, and 6 manual-review items.

Refs #337

### DIFF
--- a/docs/orrery_faction_vocabulary.md
+++ b/docs/orrery_faction_vocabulary.md
@@ -313,6 +313,13 @@ Keep as columns:
 - `primary_location`
 - `created_at`, `updated_at`, `extra_data`
 
+Preflight command:
+
+- `nexus faction-audit --slot N` performs a read-only dry run over the legacy
+  faction columns, existing `claims` / `operates_from` pair-tags, and legacy
+  faction tag categories. Review its JSON output before any destructive data
+  rewrite or column-drop migration.
+
 ---
 
 ## Mechanical Implications
--- a/nexus/api/faction_table_audit.py
+++ b/nexus/api/faction_table_audit.py
@@ -1,0 +1,912 @@
+"""Read-only audit for migrating legacy faction columns into Orrery substrate."""
+
+from __future__ import annotations
+
+from dataclasses import asdict, dataclass, field
+from decimal import Decimal
+import re
+from typing import Any, Mapping, Optional
+
+
+LEGACY_FACTION_COLUMNS = (
+    "ideology",
+    "history",
+    "current_activity",
+    "hidden_agenda",
+    "territory",
+    "power_level",
+    "resources",
+)
+KEEP_FACTION_COLUMNS = (
+    "id",
+    "name",
+    "entity_id",
+    "summary",
+    "primary_location",
+    "created_at",
+    "updated_at",
+    "extra_data",
+)
+LEGACY_TAG_CATEGORIES = ("operational_secrecy", "state")
+RUNTIME_WRITE_SURFACES = (
+    "nexus/api/db_converters.py::create_new_faction",
+    "nexus/api/commit_handler_sync.py::create_new_faction_sync",
+    "nexus/api/commit_handler.py::commit_state_updates",
+    "nexus/api/commit_handler_sync.py::commit_state_updates_sync",
+    "nexus/agents/logon/apex_schema.py::NewFaction",
+    "nexus/agents/logon/apex_schema.py::FactionStateUpdate",
+    "nexus/api/lore_adapter.py::extract_state_updates",
+)
+
+IDEOLOGY_TAGS = frozenset(
+    {
+        "authoritarian",
+        "egalitarian",
+        "traditionalist",
+        "progressive",
+        "theocratic",
+        "secularist",
+        "nationalist",
+        "cosmopolitan",
+        "imperial",
+        "communalist",
+        "mercantilist",
+        "technocratic",
+        "revolutionary",
+        "restorationist",
+        "isolationist",
+    }
+)
+RESOURCE_BASE_TAGS = frozenset(
+    {
+        "capital",
+        "force",
+        "information",
+        "faith",
+        "industry",
+        "labor",
+        "territory",
+        "patronage",
+        "bureaucracy",
+        "technology",
+        "specialized_knowledge",
+        "criminal_network",
+        "supply_lines",
+        "mobility",
+    }
+)
+OPERATIONAL_MODE_TAGS = frozenset({"overt", "covert", "hybrid"})
+POWER_STATUS_TAGS = frozenset(
+    {
+        "dominant",
+        "ascending",
+        "stable",
+        "pressured",
+        "declining",
+        "fragile",
+        "collapsed",
+    }
+)
+AGENDA_TAGS = frozenset(
+    {
+        "expand_control",
+        "consolidate_control",
+        "infiltrate",
+        "seize_leadership",
+        "settle_succession",
+        "recover_losses",
+        "negotiate",
+        "mobilize",
+        "investigate",
+        "recruit",
+        "extract_resources",
+        "sabotage",
+        "suppress_dissent",
+        "conceal_exposure",
+        "reform_internal",
+        "secure_alliance",
+        "enforce_claim",
+        "protect_asset",
+        "retaliate",
+    }
+)
+LEGACY_AGENDA_ALIASES = {
+    "revanchist": "recover_losses",
+    "infiltration": "infiltrate",
+    "coup": "seize_leadership",
+    "succession": "settle_succession",
+    "expansion": "expand_control",
+    "consolidation": "consolidate_control",
+}
+OPERATIONAL_MODE_ALIASES = {
+    "hidden": "covert",
+    "secretive": "covert",
+    "secret": "covert",
+    "underground": "covert",
+    "compartmented": "covert",
+    "cellular_clandestine": "covert",
+    "public": "overt",
+    "open": "overt",
+}
+RESOURCE_KEYWORDS = {
+    "capital": (
+        "capital",
+        "money",
+        "credit",
+        "wealth",
+        "finance",
+        "cash",
+        "funding",
+    ),
+    "force": (
+        "force",
+        "fighters",
+        "guards",
+        "troops",
+        "enforcers",
+        "weapons",
+        "military",
+        "soldiers",
+    ),
+    "information": (
+        "information",
+        "intelligence",
+        "intel",
+        "archives",
+        "surveillance",
+        "secrets",
+        "informants",
+    ),
+    "faith": ("faith", "religious", "ritual", "sacred", "pilgrim", "donor"),
+    "industry": (
+        "industry",
+        "workshops",
+        "factories",
+        "production",
+        "craft",
+        "manufacturing",
+    ),
+    "labor": ("labor", "workers", "volunteers", "conscripts", "members"),
+    "territory": (
+        "territory",
+        "land",
+        "estate",
+        "district",
+        "region",
+        "holdings",
+    ),
+    "patronage": ("patronage", "sponsors", "donors", "backers", "subsidy"),
+    "bureaucracy": ("bureaucracy", "records", "offices", "permits"),
+    "technology": ("technology", "machines", "infrastructure", "engineered"),
+    "specialized_knowledge": (
+        "knowledge",
+        "scholarship",
+        "lore",
+        "expertise",
+        "doctrine",
+        "trade secrets",
+    ),
+    "criminal_network": (
+        "criminal",
+        "smuggling",
+        "black market",
+        "fences",
+        "racket",
+        "illicit",
+    ),
+    "supply_lines": (
+        "supply",
+        "logistics",
+        "warehousing",
+        "transport",
+        "food",
+        "fuel",
+    ),
+    "mobility": (
+        "ships",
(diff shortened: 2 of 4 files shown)


pythagorakase:

pythagorakase:
## Review (manual, posted by repo owner)

> Posted manually because GitHub Actions is in a major outage today (per #339); the agent-review orchestration never fired. This review treats the PR as if it had landed in the normal `@claude` lane: coding-agent audience, Codex's existing review (no findings) acknowledged.

Net: this is a solid first cut of a preflight audit and the design fits the broader migration plan — but it has one critical coverage hole, one piece of dead code, one precedence bug, and a testing-philosophy violation that all need to land before this is trustworthy as a migration gate. The "looks right but skips 50 of 52 legacy rows in slot 2" failure mode is exactly the kind of silent under-coverage that an audit tool exists to prevent in *the migration* — so the audit holding itself to that bar is non-negotiable.

---

### Blocking

**B1. Legacy tag category coverage is missing 4 of 6 categories.**

`LEGACY_TAG_CATEGORIES = ("operational_secrecy", "state")` at `nexus/api/faction_table_audit.py:30`. Cross-check against `migrations/043_orrery_category_refactor_phase1.py` — `LEGACY_TO_NEW_CATEGORY_ALIASES` lists six faction-attached legacy categories that need migration:

| Legacy category | New target | Slot 2 rows (today) |
| --- | --- | --- |
| `ideology_axis` | `ideology` | 10 |
| `power_posture` | `power_status` | 16 |
| `legitimacy_status` | `legitimacy` | 4 |
| `operational_secrecy` | `operational_mode` | 1 (covered) |
| `resource_class` | `resource_base` | 12 |
| `hidden_agenda_class` | `agenda` | 6 |
| `history_class` | *(None — drop)* | 2 |

The audit ignores 50 of 51 legacy faction tag rows in slot 2. The PR description's claim of "28 entity-tag candidates, 16 pair-tag candidates" is correct for what was scanned, but materially understates the migration's real footprint.

Fix shape: extend `LEGACY_TAG_CATEGORIES` to the full set, add `_map_legacy_tag_row` branches for each, and (preferably) read the alias table from migration 043 rather than redeclaring it — see B4 below.

**B2. `state` faction-tag handling is dead code.**

`_map_legacy_tag_row` has a branch for `row.category == "state"` at `faction_table_audit.py:812`. `docs/orrery_faction_vocabulary.md` explicitly says "**No generic faction `state` category.**" Confirmed live:

```sql
-- slot 2 and slot 5 both:
SELECT COUNT(*) FROM entity_tags_current
 WHERE entity_kind = 'faction' AND category = 'state';
-- 0
```

The branch will never execute. Delete it. (If the design intent was to catch any historical-data slot that somehow grew a faction `state` tag, that's a `state_drain` assertion in a separate test, not a mapping rule inside the audit.)

**B3. Resource "network" check fires before the more specific keyword matcher.**

`_map_resources` (`faction_table_audit.py:711–779`) has this control flow:

1. Exact-token match against `RESOURCE_BASE_TAGS` → deterministic.
2. **`if "network" in _tokens(text):` → fan out to 3 ambiguous candidates, return.**
3. Keyword-dict scan (`criminal` → `criminal_network`, `informants` → `information`, etc.).

For a row with `resources = "criminal network"`, step 2 fires before step 3 ever runs. Reproduced:

```python
>>> _tokens("criminal network")
{'criminal', 'network'}
>>> "network" in _tokens("criminal network")
True
```

Result: emits 3 review_required candidates (`criminal_network`, `information`, `patronage`) when the keyword matcher would have produced a single `criminal_network: confidence=suggested` candidate. Same hazard for "information network", "patronage network", "supply network", etc. — every multi-word "X network" gets demoted to maximum ambiguity.

Fix: gate the "bare network" branch on `_normalize_token(text) == "network"` (the whole text is exactly "network"), or move the keyword scan above the network-only fallback.

**B4. Test suite is mock-only and violates the codebase's testing philosophy.**

`tests/test_faction_table_audit.py` builds `FakeFactionAuditCursor` (a hand-rolled cursor double that branches on SQL string content) and `FakeFactionAuditConnection`. Per `CLAUDE.md`:

> **Testing Philosophy**: The user **HATES** mock tests. If the current feature involves remote inference, for example, that means testing will be meaningless to the user unless it makes real API calls.

This audit is a database tool — the live-DB equivalent of "remote inference." The substrate of meaning here *is* the slot databases. A mock cursor that matches on `"FROM factions" in sql` is exactly the failure mode the policy is meant to prevent:

- It will not catch a renamed column (the test's cursor only branches on `"FROM factions"`, so any compatible SQL passes).
- It will not catch a view-vs-table substitution (`entity_tags_current` is a view; if its shape drifts, the test won't notice).
- It froze the slot-2 vocabulary into Python literals and will silently pass when slot 2 changes.

Fix shape: take the trait-audit pattern as the model — run against a known fixture slot (probably slot 2, since it already has the legacy-loaded data that motivates the audit). One end-to-end test that calls `build_faction_table_audit(real_cursor)` on slot 2 and asserts on counter shapes (not exact counts, since slot 2 evolves) will catch substrate drift the mock test cannot. Keep the pure-function tests for `audit_faction_row` on synthetic dicts — those are dataclass-shape assertions and don't need a live DB. But the CLI test `test_cli_faction_audit_returns_dry_run_payload` and the cursor-mock test `test_build_faction_table_audit_reports_legacy_tag_and_claim_counts` should both go live.

---

### Should-fix

**S1. Vocabulary is hardcoded inline rather than read from the substrate.**

`IDEOLOGY_TAGS`, `RESOURCE_BASE_TAGS`, `OPERATIONAL_MODE_TAGS`, `POWER_STATUS_TAGS`, `AGENDA_TAGS` are all hardcoded frozensets in `faction_table_audit.py`. These duplicate values the substrate already owns. If migration 044+ adds a `populist` ideology (e.g.), the audit will silently misclassify it as a remainder.

Two options:
- Import the canonical sets from the substrate (`nexus.agents.orrery.templates` or wherever they end up registered).
- If they're not yet registered at runtime, read `NEW_CATEGORY_REGISTRY` from `migrations/043_orrery_category_refactor_phase1.py` (it's a Python module, importable directly). Less elegant but at least authoritative.

`LEGACY_AGENDA_ALIASES` and `OPERATIONAL_MODE_ALIASES` similarly duplicate ad-hoc mappings; same fix — import from migration 043 if/when it grows an alias table, otherwise document where these came from (slot scan? hand-curated?).

**S2. No `legitimacy` category mapping.**

Migration 043 registers `legitimacy` as one of the six faction categories. The audit has zero candidate-emission paths that target it. The legacy column closest to it is probably `history` (which can describe persecution/recognition), but the audit routes `history` exclusively to "world_event_or_prose". Either:
- Add a `legitimacy` mapping branch (probably keyword-based: "outlaw", "underground", "tolerated", "chartered", "recognized", "sanctioned"); or
- Document in the audit's module docstring that `legitimacy` is intentionally hand-seeded by the migration reviewer because no legacy column expresses it cleanly. Currently the omission is silent.

**S3. `RUNTIME_WRITE_SURFACES` will rot silently.**

`faction_table_audit.py:31–39` lists 7 specific file+function locations as "runtime write surfaces still requiring cleanup." Nothing in the audit verifies any of these still exist. A future refactor that moves `create_new_faction` will leave the audit emitting stale guidance forever.

Fix shape: either (a) at import time, `importlib.util.find_spec` + `getattr` on each, raise if any reference is broken; or (b) drop the list entirely and let the migration plan track these in the doc. The cost of stale guidance in a migration tool is higher than the cost of no guidance.

**S4. `power_level` band edges mix `>=` and `>`.**

`_map_power_level` (`faction_table_audit.py:676–708`):

```python
if numeric >= 0.85:     tag = "dominant"
elif numeric >= 0.70:   tag = "ascending"
elif numeric >= 0.45:   tag = "stable"
elif numeric >= 0.30:   tag = "pressured"
elif numeric > 0.10:    tag = "declining"   # > not >=
elif numeric > 0.0:     tag = "fragile"     # > not >=
else:                    tag = "collapsed"
```

`power_level = 0.10` exactly falls through to `collapsed` (jumping over `declining` and `fragile`); `power_level = 0.0` exactly also falls through to `collapsed`. That's fine for `0.0` (collapsed reads right) but `0.10` → `collapsed` is wrong: anything below `0.30` and above `0.0` should land in declining or fragile. Use `>=` uniformly and document the boundaries as inclusive-from-below: `>= 0.10 → declining`, `> 0.0 → fragile`, `<= 0.0 → collapsed`.

(Also: the column is `numeric(3,2)` with a max of `9.99`, but realistically `power_level > 1.0` shouldn't exist. Worth either a CHECK constraint or an audit-emit when a value exceeds 1.0.)

**S5. `legacy_tag_rows` appears three times in the JSON output.**

`build_faction_table_audit` returns:
- Top-level `legacy_tag_rows: [...]` (all rows).
- Per-faction `factions[i].legacy_tags: [...]` (same rows, grouped by entity).
- Per-faction `factions[i].mapping_candidates: [...]` (derived candidates with `source_column="entity_tags_current.<category>"` that contain the same `source_value`).

For a slot with many legacy faction tags, this is gratuitous duplication. Decide on the primary representation (probably `factions[i].mapping_candidates`) and drop the others, or at minimum dedupe the global `legacy_tag_rows`.

**S6. `history_class` rows are silently dropped.**

Migration 043 lists `history_class` with a `None` target (intentionally no replacement). Slot 2 has 2 such rows. The audit doesn't emit any candidate for them (they're not in `LEGACY_TAG_CATEGORIES`). They'll vanish in the eventual migration with no acknowledgment in the dry-run output. Even just a counter `legacy_history_class_dropped: 2` and a `runtime_dispositions` block saying "these will be removed without replacement" would make the audit faithful to the migration's actual effect.

---

### Minor / nits

**N1. `_obsolete_values` is called twice per row** — once in `build_faction_table_audit` for `non_null_counts`, then again inside `audit_faction_row`. Trivial perf cost on N < 100 factions. Hoist or pass it through.

**N2. No JSON-output integration test.** Both CLI tests run `as_json=False`. The audit returns a `Decimal` (via `obsolete_columns` from `power_level`); `_json_safe` should handle it, but no test confirms `emit_output(..., as_json=True)` round-trips cleanly. Add one.

**N3. `LEGACY_AGENDA_ALIASES` and `OPERATIONAL_MODE_ALIASES` need provenance comments.** A reviewer running this audit can't tell whether `revanchist → recover_losses` is hand-curated or substrate-derived without spelunking. One-line comment naming the source ("observed in slot 2 scan 2026-05-25" or "from migration 043 alias table") would help.

**N4. `_match_controlled_values` does both spaced-form and underscored-form matches** (`faction_table_audit.py:891-892`). For tags without underscores this is redundant. Comment what's intentional.

**N5. No slot-lock check.** `run_faction_audit` reads from any slot, locked or not. Trait-audit has the same property today, so this isn't a regression — but worth a note: read-only operations should still surface "slot N is locked, results may be stale" rather than appearing as if the data is current.

---

### Adopt as-is

- The `SET TRANSACTION READ ONLY` guard at `nexus/cli.py:1024` is the right defense and worth replicating elsewhere.
- The split between `target_kind=entity_tag` / `pair_tag` / `world_event_or_prose` / `structured_remainder` is the right taxonomy and matches the spec doc's framing.
- The `_map_territory` interaction with existing `claims(...)` edges (seed `resource_base:territory` deterministically when claims already exist) is clever and correct — it's the one place the audit currently reads substrate state to make a confident assertion, and it's exactly the pattern the rest of the audit should follow.
- The `confidence: deterministic | suggested | ambiguous | manual_review` axis is well-designed; the reviewer's batching workflow will be much faster than it would be without it.
- The CLI shape (`nexus faction-audit --slot N`) mirrors `trait-audit` cleanly and slots into the existing dispatcher without surprises.

---

### Suggested commit order

1. B1 + B2 + S2 + S6 together (legacy-category coverage; this is one coherent change touching `_fetch_legacy_tag_rows` + `_map_legacy_tag_row` + `LEGACY_TAG_CATEGORIES` + a new `legitimacy` mapping or doc note).
2. B3 (resource network precedence) — surgical, isolated.
3. B4 (live-DB tests) — replace the cursor mock with a slot-2 fixture pass.
4. S1 (substrate-read instead of hardcode) — bigger refactor, lands cleanly after B1/B2/B4 stabilize the surface.
5. S3, S4, S5 + nits — easy follow-ups.

Items 1–3 are pre-merge; 4–5 can land as follow-up commits or a follow-up PR if you want this on `main` sooner.

PR #358, opened 2026-05-29
Fix extract-vector-grid dropping axis-aligned line segments

pythagorakase:
## Summary

`extract-vector-grid` was silently dropping **axis-aligned line segments**. `collect_shapes` and `_shape_signature` rejected any shape whose bbox had zero width *or* zero height (`xmax <= xmin or ymax <= ymin`) — but a horizontal line has `ymax == ymin` and a vertical line has `xmax == xmin`. For line-art sheets, that's most of the artwork.

**Discovered while extracting the Art Deco set (Adobe Stock 549888080):** 64 of 552 paths dropped across 7 of 9 frames (frame borders, radial ticks). The curve-heavy Art Nouveau set (482408992, the original acceptance test) was unaffected — which is exactly why this shipped.

## Root cause

```python
# collect_shapes (line ~139) and _shape_signature (line ~261), before:
if xmax <= xmin or ymax <= ymin:   # kills every straight stroke
```

## Fix

- Reject only **inverted** boxes and **true zero-extent points** (degenerate on *both* axes); keep line segments:
  `if xmax < xmin or ymax < ymin or (xmax == xmin and ymax == ymin):`
- **Path-conservation guard** in `extract()`: raise if `emitted + background_dropped != total_shape_nodes`. A lossy extraction now fails loudly at generation time instead of shipping silently.
- **`scripts/selftest.py`** — self-contained regression test (no pytest dep; runs in the skill venv) that extracts a synthetic line-only 2-frame sheet and asserts all 8 segments survive. Verified it **fails on the pre-fix code** (`got 0`) and **passes after**.
- **SKILL.md** — failure mode #6 documented, plus a note that visual validation alone is insufficient (it missed 6 of 7 broken frames in practice).

## Verification

| Set | Before | After | Result |
|---|---|---|---|
| Art Deco 549888080 | 487 paths emitted | 551 | all 64 recovered; `flat_only=0`, no contamination |
| Art Nouveau 482408992 | 271 | 271 | **identical** — fix is non-disruptive to the merged set |

Path-level diff (every converter `d=` string must land in exactly one cell) + the new conservation guard both pass on both sets.

## Follow-ups (separate PRs)

- **#357** (Art Deco import) — re-extract with this fix before merge; current SVGs are defective.
- **Art Nouveau #355** — verified clean (fixed output ≡ committed); no action needed.

## Test plan

- [x] `selftest.py` passes on fixed code, fails on `origin/main` code
- [x] Conservation guard passes on both real sets (exit 0)
- [x] `flat_only=0` for Art Deco; fixed≡committed for Art Nouveau
- [x] Grid detection unchanged (3×3 / 2×5, same cuts)

### DIFF
--- a/.claude/skills/extract-vector-grid/SKILL.md
+++ b/.claude/skills/extract-vector-grid/SKILL.md
@@ -14,6 +14,7 @@ It bakes in several hard-won failure modes that naïve grid-cutters hit:
 3. **There is usually a page-spanning background shape.** Detect it generically by bbox-area ≈ page area (no fill-color assumption, no assumption a background exists).
 4. **Never cut on nominal grid divisions** (`page_width / cols`). Ornaments routinely overhang their cell by a few points — cutting on grid lines amputates corner details. Use projection-profile gutter detection and cut through the **middle of the empty bands**.
 5. **`svgelements` normalizes `width="...pt"` to 96-DPI pixels**, but the original path `d=` strings stay in viewBox user-units. The two coordinate systems must be reconciled before emitting per-cell SVGs (a 0.75 factor for any `pt`-declared source).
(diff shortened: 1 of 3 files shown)


pythagorakase:
Addressed in 2376c73e:
- **Issue 1 (guard message):** now diagnosis-neutral — lists candidate causes (bbox filters / background detection / cell coverage) rather than blaming the degenerate-bbox filter.
- **Issue 3 (coverage):** `selftest.py` now also runs `extract()` end-to-end, exercising `_shape_signature` (emit-stage filter) and the conservation guard.
- **Issue 2 (float ==):** kept exact equality by design — it culls only true zero-extent points, never thin-but-real geometry (keep-by-default is the safe direction); documented inline.
- **Nit:** comments trimmed to the WHY.

PR #377, opened 2026-06-10
U5: Settings Pane Wiring — the Calibration Console

pythagorakase:
## Summary

The placeholder SettingsPane becomes the IRIS kit's calibration console (`design_handoff/project/ui_kits/nexus_iris/settings_pane.jsx`), wired to real persistence in `nexus.toml`. All writes flow through a new typed FastAPI surface and the config loader's formatting-preserving `save_settings` path — no local draft state, optimistic UI with rollback, server confirmation from the PATCH response.

## Sections

| Section | Status | Persistence |
| --- | --- | --- |
| Theme | **Wired** | `ui.theme` (new) — splash dropdown + ThemeMenu also persist now, since all theme changes flow through ThemeContext |
| Typography | **Wired** | `ui.fonts.<theme>.<slot>` (new) — keeper matrix per theme; display slot locked per the marquee rule |
| Narrative Mode | **Wired** | `global.narrative.test_mode` — kit lever + danger alert |
| Model | **Wired** (role-level only) | `apex.model` + derived `apex.provider`; `wizard.default_model`. Only `@provider.role` refs selectable — never raw model IDs |
| LORE Budget | **Wired** | `lore.token_budget.apex_context_window` — ladder + slider + COMMIT, bounds/stops from `ui.lore_budget_slider` (new) |
| Typewriter | **Wired** | `ui.typewriter_ms_per_char` — added as a seventh rail section (kit's data has `typeSpeed` but no section; the ask requires it) |
| PWA Icon | **Wired** | The Express multer + sharp pipeline (`POST /api/settings/pwa-icon`) already existed — dropzone + preview + COMMIT GLYPH per kit, cache-busted current icon |

## Read-Only by Design (Safety Judgments)

- **Active retrieval embedder** (`memnon.models.*`) — rendered in the Model card as a locked block with explanatory copy. Every stored embedding is encoded with this model; switching mid-story orphans the archive and requires a re-embedding campaign.
- **`global.narrative.test_database_suffix`** — displayed in the routing copy but not editable; renaming mid-flight orphans existing `_test` tables.
- **`apex.provider`** — derived from the role reference server-side, never directly editable.
- **`apex.reasoning_effort` / `max_output_tokens`, `lore.payload_percent_budget`, `wizard.fallback_model`** — not in the kit's six sections; left out rather than inventing UI without design.

## API Changes

- **New** `nexus/api/settings_endpoints.py`: `GET /api/settings` (raw TOML payload, role refs unresolved, `[secrets]` stripped, legacy `Agent Settings` aliases preserved, derived `settings_meta` block) and `PATCH /api/settings` (typed `SettingsPatchRequest`, `extra=forbid`, writes via `save_settings` with full-document Pydantic validation; 422 on rejection, file untouched).
- **`settings_meta`** carries the role catalog, the apex provider allowlist (derived from the `APEXSettings.provider` pattern — so TEST is excluded without hardcoding), and typewriter bounds (from `UISettings` field constraints). Components never hardcode config semantics.
- **Express** (`ui/server/routes.ts`): exact `/api/settings` now proxies to FastAPI :8002; the regex-based TOML writer (2 hardcoded keys, no validation) is deleted. `/api/settings/pwa-icon` falls through to the local multer + sharp route.
- **`UISettings`** gains `theme`, `fonts` (per-theme slot matrix), `lore_budget_slider`; all seeded with comments in `nexus.toml [ui]`.
- **Provider order** in `App.tsx`: `QueryClientProvider` now outermost so Theme/Font contexts hydrate from and persist through the settings query.

## Round-Trip Evidence (live API, throwaway value)

```
GET  ui.typewriter_ms_per_char        -> 35
PATCH {"typewriter_ms_per_char": 42}  -> 42 (response + file + fresh GET)
PATCH {"typewriter_ms_per_char": 35}  -> 35
diff nexus.toml.before nexus.toml     -> IDENTICAL (comments + formatting preserved)
```

Also exercised live: `apex_model_ref` `@openai.default -> @anthropic.deep -> back` (both `apex.model` and `apex.provider` rewritten, inline comments intact), `apex_context_window` 75000 <-> 100000 (inline comment intact; note tomlkit normalizes the `75_000` underscore literal to `75000` on write — value-identical), theme + fonts patch and restore, and the rejection paths: `@test.default` for apex -> 422 (provider pattern), unknown role -> 422, malformed ref -> 422, out-of-bounds typewriter -> 422, empty patch -> 400, unknown key -> 422. Rejected writes never touch the file.

## Validation

- `tsc` clean, `vite build` clean
- Dev-server smoke: Express :5189 -> FastAPI :8002 proxy GET/PATCH round-trip green; `pwa-icon` fall-through confirmed local (400 "No file uploaded", not 404)
- `black` / `flake8` / `mypy` clean on changed Python
- `pytest tests/test_api tests/config` -> 29 passed, 1 skipped (14 new endpoint tests, real file I/O, no mocks)
- No DB writes; persistence under test is the config file only

## Notes

- shadcn directive: the full primitive set (switch/select/slider) is already vendored under `ui/client/src/components/ui/`; the kit's lever/slider/chips are bespoke IRIS design elements implemented per kit CSS with proper ARIA (`role="switch"`), so nothing needed installing.
- Kit's "OPEN PALETTE COMPARISON" link omitted (no target surface in production).
- Kit `--danger`/`--warning` map to `hsl(var(--destructive))`/`--bronze` — tokens only, no invented hues.

### DIFF
--- a/nexus.toml
+++ b/nexus.toml
@@ -68,6 +68,34 @@ test_database_suffix = "_test"
 # system specifies 30-50 ms/char).
 typewriter_ms_per_char = 35
 
+# Active NEXUS IRIS theme, persisted across sessions via PATCH /api/settings.
+theme = "veil"
+
+[ui.lore_budget_slider]
+# Bounds, step, and preset ladder stops for the LORE budget slider in the
+# settings pane. Values are token counts for lore.token_budget.apex_context_window.
+min = 10_000
+max = 200_000
+step = 1000
+stops = [75_000, 100_000, 150_000, 200_000]
+
+# Per-theme font slot choices (NEXUS IRIS keeper matrix). The display slot is
+# the marquee font — reserved for the NEXUS wordmark and locked in the UI.
+[ui.fonts.veil]
+body = "Spectral"
+menu = "Cinzel"
+display = "Megrim"
+
+[ui.fonts.gilded]
+body = "Cormorant Garamond"
+menu = "Space Mono"
+display = "Monoton"
+
+[ui.fonts.vector]
+body = "Rajdhani"
+menu = "Source Code Pro"
+display = "Sixtyfour"
+
 # =============================================================================
 # Wizard Settings (New Story Setup)
 # =============================================================================
--- a/nexus/api/narrative.py
+++ b/nexus/api/narrative.py
@@ -61,6 +61,7 @@ from nexus.api.narrative_generation import (
     generate_bootstrap_narrative,
 )
 from nexus.api.config_utils import get_max_choice_text_length
+from nexus.api.settings_endpoints import router as settings_router
 from nexus.api.slot_endpoints import router as slot_router
 from nexus.api.setup_endpoints import router as setup_router
 from nexus.api.wizard_chat import router as wizard_chat_router
@@ -80,6 +81,7 @@ app.add_middleware(
 
 
 # Include modular routers
+app.include_router(settings_router)
 app.include_router(slot_router)
 app.include_router(setup_router)
 app.include_router(wizard_chat_router)
--- a/nexus/api/settings_endpoints.py
+++ b/nexus/api/settings_endpoints.py
@@ -1,0 +1,229 @@
+"""
+FastAPI endpoints for the operator settings surface (GET/PATCH /api/settings).
+
+GET serves the raw nexus.toml contents (with @provider.role references left
+unresolved so the client can show role-level bindings), plus the legacy
+"Agent Settings"/"API Settings" aliases the React client predates, plus a
+derived ``settings_meta`` block (model role options, apex provider allowlist,
+typewriter bounds) so the client never hardcodes config semantics.
+
+PATCH accepts a typed subset of safe-to-edit keys and persists them through
+``nexus.config.loader.save_settings``, which validates the merged document
+against the Pydantic Settings model and preserves comments/formatting via
+tomlkit. Keys outside this subset (active embedding model, test database
+suffix, reranker paths, ...) are intentionally not writable from the UI.
+"""
+
+import logging
+import re
+from pathlib import Path
+from typing import Any, Dict, List, Literal, Optional
+
+from fastapi import APIRouter, HTTPException
+from pydantic import BaseModel, ConfigDict, Field, ValidationError
+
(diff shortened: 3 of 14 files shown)


pythagorakase:
Review feedback addressed in 62a28d8 (Codex P1) and 511b75d (Claude review):

**Codex P1 — HEAD probe**: Confirmed live (405 before, 200 after). FastAPI's `APIRoute` indeed does not inherit Starlette's auto-HEAD; explicit `@router.head("")` added, returning 200 when the config is readable and 500 otherwise (parity with the retired Express handler). Pinned by a TestClient test.

**Claude review**:
1. **Draft staleness** — fixed. Both slider sections drop their draft via `useEffect(() => setDraft(null), [saved])`, so optimistic commits and rollbacks both re-sync the control to server truth.
2. **Optimistic font merge** — fixed. `applySettingsPatch` now merges against `payload.ui?.fonts ?? {}` instead of skipping when the cache predates `ui.fonts`.
3. **Display slot writable via PATCH** — intentional, now documented on `FontSlotsPatch`: the pane's RESET TO KEEPERS PATCHes the full keeper matrix (display included), and the API stays the operator escape hatch for rebinding the marquee without hand-editing nexus.toml. The UI lock enforces the marquee rule at the design surface.
4. **Concurrency** — documented on `patch_settings`: unlocked read-merge-write is safe for the deployed shape (single operator, single-worker uvicorn, no awaits inside the critical section); multi-worker would need file locking first.
5. **Scroll effect dep** — refactored. The console is now a child component that only mounts once settings exist, so the effect binds on mount with `[]`.
6. **Constraint introspection at request time** — fixed. The apex provider pattern and typewriter bounds resolve at module import, so a renamed Pydantic constraint fails at server startup rather than on the first GET.
7. **Silent context mutation errors** — already covered: `useSettingsMutation.onError` logs `console.error("Settings update failed:", error)` for every caller, contexts included, in addition to the cache rollback (which reverts the visible theme/fonts via the hydration effects). Left as is.

Validation after fixes: tsc + vite build clean; black/flake8/mypy clean; 34 passed, 1 skipped (tests/test_api + tests/config); HEAD/GET verified live on uvicorn :8002.

PR #384, opened 2026-06-12
Slot Manager: Visual Empty/Occupied/Bound States

pythagorakase:
## Summary

User feedback: the slot manager gave no visual indication of which slots are empty vs occupied, and the only occupancy signal was an "ACTIVE" text badge — exactly the label clutter the user did not want. This PR replaces text-based state signaling with a purely visual vocabulary built from the vendored IRIS/Veil theme tokens.

## Visual Vocabulary

| State | Treatment | Why |
| --- | --- | --- |
| Occupied | "Lit": solid `--primary`-tinted border, full-opacity `bg-card` surface, glowing slot-number tile (`terminal-glow`) | Presence reads as light/energy, consistent with the terminal aesthetic |
| Empty | "Hollow": faint dashed border on card and tile, recessed `bg-card/20` surface, content dimmed to 50% with a hover lift | Dashed outline is the established idiom for a vacant receptacle; dimming makes it visually recessive at a glance |
| Wizard-in-progress | Lit treatment plus a small pulsing primary mote on the tile corner (`animate-pulse`, `motion-reduce:animate-none`) | "In motion, incomplete" without text |
| Currently-bound story (`localStorage activeSlot`) | 3px glowing left edge-marker | The same "you are here" cue nexus-layout.css uses for the current chunk — vocabulary reuse, not invention |
| Selected | Existing primary ring, now forced `border-solid` so it stays legible over a hollow slot's dashed border | Unchanged semantics |
| Locked | Existing destructive tint | Unchanged, and now actually reachable (see bug fix) |

The "ACTIVE" text badge is removed. All colors are theme tokens, so the treatment tracks Gilded/Vector/Veil automatically.

## Bug Fixed in Passing

The queryFn dropped `is_locked` from the API response, so the locked styling and the locked-slot click guard were dead code. It is mapped again (the two never-returned fields `character_name`/`last_played` were dropped from the mapping).

## Accessibility

States are visual-only on screen, but every slot card carries `aria-label` + `title` describing the full state (e.g. "Memory Slot 2: story setup in progress (character phase), current story"), and the CLEAR/RESUME/INITIALIZE buttons get descriptive aria-labels.

## Architecture

Classification logic (slot API data -> visual spec) lives in a pure module, `ui/client/src/components/NewStoryWizard/slotVisualState.ts`, kept free of React so it is unit-testable. `SlotSelector.tsx` consumes the spec and adds `data-occupancy` for test/automation hooks.

## Verification

- `npm test`: 41/41 passing (19 new tests covering classification precedence, class output per state, tailwind-merge dashed/solid resolution, bound-slot matching, localStorage parsing, aria labels).
- `npx tsc --noEmit` and production build (`npm run build`) clean.
- Real-browser screenshots (puppeteer + system Chrome) against a worktree dev server on :5022 proxying the live API (read-only):
  - Live state (all five slots currently occupied, slot 1 bound): occupied slots lit, beacon on slot 1.
  - Mixed state via request interception serving a modified copy of the real endpoint response (no database mutation): slot 1 occupied+bound, slot 2 wizard-in-progress, slot 3 empty, slot 4 locked, slot 5 occupied — all six visual states distinguishable side by side with zero added text.
  - Hover over the empty slot confirms the hover lift.
  - Rendered aria labels confirmed via DOM inspection in the same run.

### DIFF
--- a/ui/client/src/components/NewStoryWizard/SlotSelector.tsx
+++ b/ui/client/src/components/NewStoryWizard/SlotSelector.tsx
@@ -164,6 +164,13 @@ export function SlotSelector({ onSlotSelected, onSlotResumed }: SlotSelectorProp
                         <Card
                             key={slotData.slot}
                             onClick={() => handleSelect(slotData.slot, slotData)}
+                            onKeyDown={(e) => {
+                                if (e.target === e.currentTarget && (e.key === "Enter" || e.key === " ")) {
+                                    e.preventDefault();
+                                    handleSelect(slotData.slot, slotData);
+                                }
+                            }}
+                            tabIndex={0}
                             role="group"
                             aria-label={visuals.ariaLabel}
                             title={visuals.ariaLabel}
@@ -240,6 +247,10 @@ export function SlotSelector({ onSlotSelected, onSlotResumed }: SlotSelectorProp
                                         <Button
                                             variant="ghost"
                                             aria-label={`Initialize empty Memory Slot ${slotData.slot}`}
+                                            onClick={(e) => {
+                                                e.stopPropagation();
+                                                handleSelect(slotData.slot, slotData);
+                                            }}
                                             className={cn(
                                                 "font-mono text-xs border border-transparent group-hover:border-primary/30",
                                                 selectedSlot === slotData.slot && "bg-primary text-primary-foreground hover:bg-primary/90"
--- a/ui/client/src/components/NewStoryWizard/SlotSelector.tsx
+++ b/ui/client/src/components/NewStoryWizard/SlotSelector.tsx
@@ -1,8 +1,9 @@
-import { useState, useEffect } from "react";
-import { Save, AlertTriangle, Loader2, Trash2 } from "lucide-react";
+import { useState } from "react";
+import { AlertTriangle, Loader2, Trash2 } from "lucide-react";
 import { Button } from "@/components/ui/button";
 import { Card } from "@/components/ui/card";
 import { cn } from "@/lib/utils";
+import { getSlotVisuals, readBoundSlot } from "./slotVisualState";
 import { useQuery, useMutation, useQueryClient } from "@tanstack/react-query";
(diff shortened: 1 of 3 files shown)


pythagorakase:
Review feedback addressed in 512ae1ce:

**Claude review**
- *classifySlot wizard/is_active conjunction:* verified the API invariant rather than adding the guard. `slot_state.py` derives both fields from data presence and only returns wizard mode when a wizard cache row exists (`is_empty=False`), so `wizard_in_progress: true` implies `is_active: true` at the endpoint. The precedence is also defensively correct: a slot with resumable wizard data is not hollow regardless of `is_active`. Documented on the function.
- *role="group" accessibility:* cards are now keyboard-operable (`tabIndex=0`, Enter/Space select, focus-visible primary ring), and the INITIALIZE button carries an explicit select handler instead of relying on click bubbling. Note the prior code was a plain clickable div with no role or aria at all, so this is strictly additive. Making the tile a nested button was rejected to avoid a second tab stop per card.
- *CARD_LIT/TILE_LIT dedup:* extracted as suggested.
- *Minor items:* cross-tab `storage` reactivity and range-checking `readBoundSlot` are deferred — out-of-range values already degrade safely to no beacon, and a 1-5 range guard would hardcode the slot count in the client. `title` kept identical to `aria-label` intentionally: the hover tooltip doubles as the sighted-user legend for the visual vocabulary.

**Codex review**
- *P2 stale beacon on cleared slot:* fixed by gating `bound` on non-empty occupancy — an empty slot can never read as the current story, even with a stale `localStorage.activeSlot`. Test added. The deeper issue (clearing `activeSlot` when the bound slot is reset, which also affects the splash Continue route) predates this PR and is left as a follow-up.

Validation after fixes: 43 vitest tests pass, `tsc --noEmit` clean, production build clean, real-browser keyboard check (focus empty card, Enter) confirms selection advances the wizard.

PR #385, opened 2026-06-12
App Icons: Per-Theme Set, Locked; Retire Dead Upload Path

pythagorakase:
## Summary

The design handoff shipped three per-theme variants of the new circuit-tree app icon, but the PWA was still serving the old icon with a dead "upload your own" control in settings. This PR vendors the per-theme set, wires favicons to follow the active theme, and retires the upload path.

## Asset Provenance

Found, not generated. The handoff (`design_handoff/project/assets/`) contains finished per-theme sets at 512/192/180/32, confirmed against the kit's `preview/brand-favicons.html` and the chat record (the Vector icon is the corrected blue-tree/green-river version, per the "VECTOR and VECTOR ALT switched" exchange in chats/chat1.md):

- Veil: magenta tree, violet river (`icon-veil-*`)
- Gilded: amber tree, teal river (the kit's original `icon-*` set)
- Vector: blue tree, green river (`icon-vector-*`)

Vendored as `ui/client/public/icons/<theme>/icon-{512,192,180,32,16}.png`. Two derived assets were generated locally with Pillow:

- The 16px favicons: Lanczos downscale of each theme's designer-tuned 32px asset (the handoff preview renders the 32 at 16px for the 16 slot, so the 32 is the small-size master).
- The root `favicon.ico` legacy fallback: rebuilt as a 32+16 PNG-in-ICO from the Veil set (Veil is the default theme).

The old generic `/icons/icon-*.png` set (including `icon-source.png`, which existed only to serve the upload flow) is removed; per-theme master PNGs remain in the design handoff. Safe-zone margins of the new 512s match or improve on the old icon's, so the manifest keeps its existing `any maskable` purpose.

## Runtime Wiring

- `src/lib/themeIcons.ts` exports `themeIconPath()` and `applyThemeIcons()`, which retarget the `link[rel=icon]` (32/16) and `link[rel=apple-touch-icon]` tags; `ThemeContext`'s theme effect calls it on every theme change, using the existing `?v=` cache-bust pattern (`ICON_VERSION`).
- `index.html` static links point at the Veil set so the first paint matches the default theme; `meta theme-color` updated from the retired teal to Veil blue-black (`#09101c`).
- PWA install icons are bake-time and follow the default theme (Veil): both the static `manifest.json` and the VitePWA generated manifest now reference `/icons/veil/...`. No dynamic manifest route, by design.

## Removed

[...]

PR #388, opened 2026-06-12
CLAUDE.md: Enshrine Visual Minimalism as a UI Design Principle

pythagorakase:
## Summary

Adds a **UI Design: Visual Minimalism** section to CLAUDE.md, codifying the user's standing directive against *visual verbosity* — the LLM tendency to clutter UIs with excessive, redundant textual labels.

Core tenets: every label must earn its place (high threshold); prefer visual self-evidence and discoverability; never restate on-screen information; no explanatory paragraphs in persistent chrome; one label per section; internal module names (LORE/LOGON/MEMNON) never surface in UI copy; natural case for content; transient status vanishes when idle.

Reference example per the user: the unlabeled theme toggle on darioamodei.com.

This lands first so the in-flight UI refinement lanes and the subsequent full-UI audit pass inherit the principle as repo doctrine.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #389, opened 2026-06-12
Chrome: Drop Splash Footer, Retitle to NEXUS, Fix Icon Safe-Zone Crop

pythagorakase:
## Summary

Three chrome-level fixes per user feedback, applying the visual minimalism principle (PR #388): every verbal label must earn its place.

### 1. Splash Footer Bar: Deleted

The Veil splash had a thin bottom text bar reading "NEXUS IRIS / narrative intelligence system" — restating what the splash wordmark already says. The footnote block is removed from `VeilSplash.tsx`. Gilded and Vector splashes never had the bar (verified in-browser; the only "narrative intelligence" match in the UI tree was the Veil footnote).

### 2. App/Window Titles: "NEXUS" Everywhere

The installed PWA showed "NEXUS IRIS - Narrative Intelligence System" on the window title bar and macOS menu bar. Surfaces and what controls them:

- **Window title**: `document.title` from `index.html` — now `NEXUS`. No runtime `document.title` writers exist anywhere in the client or server, so one static word covers every page.
- **macOS menu bar / app name**: the manifest `name` — the static `ui/client/public/manifest.json` name is now `NEXUS` (its `short_name` already was). The VitePWA manifest in `ui/vite.config.ts` was already `NEXUS`/`NEXUS`, so the two manifests are back in sync (they were synced in PR #385).
- `apple-mobile-web-app-title` was already `NEXUS` (unchanged).

**Caveat for the user**: an already-installed PWA snapshots the manifest name at install time. To see "NEXUS" in the macOS menu bar and Dock, reinstall the PWA (or wait for Chrome's manifest update cycle and relaunch; the macOS icon cache may also need a refresh / re-login).

### 3. Icon Safe-Zone Crop: Fixed

The PR #385 per-theme icons were full-bleed — the tree artwork ran to the canvas edge, so the macOS rounded-square mask amputated the branches (user screenshot showed magenta branches cut off at the corners).

All sizes regenerated (veil/gilded/vector x 512/192/180/32/16 + root `favicon.ico` from the Veil 32/16 pair) from the 1024px design-handoff masters via the new reproducible `scripts/generate_app_icons.py` (Pillow + numpy):

- **Padding geometry**: the artwork bounding box (detected against the sampled background, stable for thresholds 24-60) is scaled to **78% of the canvas** and centered — inside the W3C maskable safe zone (80% circle) target band and clear of the macOS squircle corner crop. The masters' art bbox was previously 83% wide x 91.5% tall.
- **Seamless padding**: the background color is the median of each master's 8px border ring (per-channel ring std < 1.0, i.e. effectively flat), so the extended padding is invisible.
- **Programmatic proof**: the script composites a macOS-style rounded-rect mask (22.37% corner radius, slightly stricter than Apple's squircle) over each regenerated 512 and fails loudly if any artwork pixel falls outside. All three themes: **0 artwork pixels lost**. Masked previews were also generated and visually inspected — full ring, branches, and river tail comfortably inside the crop on all three liveries.
- **Cache-busting**: `ICON_VERSION` in `themeIcons.ts` and all `?v=` tokens in `index.html` + `manifest.json` bumped to `20260612b`, and the existing NetworkFirst runtime caching rule for `/icons/*.png` picks up the new files. Installed clients may still need a PWA reinstall / icon-cache refresh for the **Dock icon** specifically (macOS caches installed-app icons aggressively).

## Validation

- `npx vitest run` — 86/86 green (includes `themeIcons.test.ts`, which derives expectations from `ICON_VERSION` so no test churn)
- `npm run build` — green; built `manifest.webmanifest` and `manifest.json` both read name/short_name `NEXUS`; built `index.html` title is `NEXUS`
- `vite preview` (port 5027): all 15 per-theme icon URLs, `favicon.ico`, and both manifests return 200 with the new `?v=` token
- Headless Chrome (system Chrome via puppeteer-core) over all three theme splashes: footer bar absent, no "NEXUS IRIS" text anywhere on the splash, `document.title === "NEXUS"`, per-theme favicon links retarget with `?v=20260612b`
- `scripts/generate_app_icons.py` mask check: 0 artwork pixels outside the macOS-style mask for every theme's 512


pythagorakase:
## Review Disposition

**Codex P2 (undeclared Pillow)** — Accepted. Pillow was only available transitively; now pinned `pillow ^11.0` in the poetry dev group (installed by default with `poetry install`) with a lockfile refresh (76ae6420).

**Claude: favicon.ico three-frame bug** — Does not reproduce; disproven against the committed artifact. Pillow >= 9.1 fills each `sizes` entry with an exact-size match from `[base] + append_images` before falling back to downsampling the base, so the generated ICO contains exactly **two** directory entries, and the 16px frame is **byte-identical** to the dedicated padded `icon-16.png`:

```
ICO directory frame count: 2
frame sizes in directory order: [(16, 16), (32, 32)]
16px ICO frame identical to dedicated icon-16.png: True
```

The suggested fix (`sizes=[(32, 32)]`) would actually *drop* the 16px frame entirely, since Pillow only emits frames for listed sizes. Instead, the save is now wrapped in `write_favicon()` with a post-write verification (frame count == 2, both frames byte-identical to the dedicated padded frames) so any future Pillow behavior change fails loudly (6bafe36a).

**Claude: hardcoded `DEFAULT_MASTERS` path** — Accepted. `main()` now fails loudly with a pointer to the untracked design-handoff location and the `--masters` flag when the directory is missing (6bafe36a).

**Claude: double `convert("RGB")`** — Accepted, hoisted (6bafe36a).

Regenerating after these changes produces byte-identical PNGs and favicon.ico (reproducibility confirmed).

PR #390, opened 2026-06-12
Characters: Natural-Case Names, Place Names, Working Portraits

pythagorakase:
## Summary

Cast pane cleanup per the visual-minimalism doctrine (PR #388), plus a full repair of the half-broken portrait feature.

## Root Cause of Alex's Broken Portrait

The save_01 `assets.character_images` row for Alex stores `file_path` **with a leading slash** (`/character_portraits/1/1759852333889_9jangr.png`); current upload code writes paths without one. The client rendered `src={`/${filePath}`}`, which turned the legacy row into `//character_portraits/...` — a **protocol-relative URL** that resolves against host `character_portraits` and renders as a broken image. The file itself was intact on disk the whole time, so per the brief the fix is resolution logic, not data: a `portraitSrc()` helper collapses leading slashes before rooting the path. **No save_01 write was needed or performed.**

## Storage and Serving Design

- Files: `ui/client/public/character_portraits/<characterId>/<timestamped name>` (gitignored), written by the existing multer route.
- Rows: per-slot `assets.character_images` (`file_path`, `is_main`, `display_order`).
- Serving: added an explicit Express static route for `/character_portraits` so runtime uploads resolve in production builds too (previously only Vite dev mirrored `client/public`; a prod bundle serves the frozen `dist` copy, which predates any upload).
- `GET /api/characters` now LEFT JOINs `places` (location name) and subselects each character's display portrait (main first, then display order) — list thumbnails need no per-row second fetch. New shared type `CharacterListEntry = Character & { currentLocationName, portraitPath }`.

## Slot Bugs Found and Fixed in the Image Routes

- `POST /api/characters/:id/images` computed `existingImages` **without passing the slot**, consulting the default database — uploading a slot's first portrait could mis-flag it `is_main = 0` (reproduced live against save_05 before the fix).
- `DELETE /api/characters/:id/images/:imageId` looked up the file path the same wrong way.
- A failed row insert used to leak the already-copied file; it is now unlinked before the error propagates (still fail-loud).

## Label Removals (Minimalism Sweep)

- "CAST MEMBER" eyebrow above the dossier name: removed.
- "CAST · N" roster header (restated the rail's Characters tab): removed.
- List rows read "A ALEX · 2" (glyph + uppercased name + raw `places.id`): now just the avatar and "Alex". Names render in natural case via the body face — the veil theme's menu face is Cinzel, a small-caps design with no true lowercase, so dropping `text-transform` alone could not satisfy "names render in natural case".
- "currently in 2" caption: now "at The Land Rig" (server-resolved place name, body face, never uppercased); nothing renders when the location is unknown — no "at null", no placeholder.
- Dossier title likewise natural case.

## Upload Affordance

Icon-only hover overlay on the dossier portrait (also keyboard-focusable, `aria-label` only — no visible label text). Picks PNG/JPEG, posts to the existing route, promotes the new image to main so the visible portrait tracks the latest choice, invalidates both queries so list thumbnail and detail update together. Errors render loud beside the name (`role="alert"`).

## Verification

- `npm test`: 94 passed, including 8 new `CharactersPane` tests pinning the legacy leading-slash fix, the label removals, the "at {place}" line, and the upload affordance (pre-seeded query caches built from real save_01 row shapes; no fetch interception).
- `npm run check` (tsc) clean; `npm run build` green.
- Real browser (puppeteer + system Chrome, dev server on :5024):
  - Slot 1: Alex's portrait renders (`naturalWidth` 1856), her row shows the thumbnail, all names natural case with no id badges, "at The Land Rig" / "at Data-Barge Okami" lines correct, hover overlay appears.
  - Slot 5: end-to-end upload of a real PNG through the UI affordance — row landed `is_main = 1, display_order = 0` (post-fix) and the portrait rendered in both list and detail (`naturalWidth > 0` checked). Test row and file removed afterward; save_05 was concurrently reset by another lane, which is unrelated.
- save_01 was touched read-only throughout; no data writes performed anywhere except the save_05 upload test (cleaned up).

### DIFF
--- a/ui/client/src/components/nexus/CharactersPane.test.tsx
+++ b/ui/client/src/components/nexus/CharactersPane.test.tsx
@@ -1,0 +1,157 @@
+/**
+ * CharactersPane tests - cast roster minimalism and portrait resolution.
+ *
+ * The legacy file path below is the real save_01 assets.character_images row
+ * for Alex (id 1) as of 2026-06-12: stored WITH a leading slash, which the
+ * old `/${filePath}` interpolation turned into a protocol-relative URL
+ * ("//character_portraits/...") and a broken image. These tests pin the fix.
+ *
+ * Component tests render against pre-seeded react-query caches built from
+ * real row shapes (no fetch interception): data drawn from save_01.
+ */
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { render, screen, within } from "@testing-library/react";
+import { describe, expect, it } from "vitest";
+import { ThemeProvider } from "@/contexts/ThemeContext";
+import { CharactersPane, portraitSrc } from "./CharactersPane";
+import type { CharacterImage, CharacterListEntry } from "@shared/schema";
+
+const ALEX_LEGACY_PATH = "/character_portraits/1/1759852333889_9jangr.png";
+
+function makeCharacter(
+  overrides: Partial<CharacterListEntry> & { id: number; name: string },
+): CharacterListEntry {
+  return {
+    summary: null,
+    appearance: null,
+    background: null,
+    personality: null,
+    emotionalState: null,
+    currentActivity: null,
+    currentLocation: null,
+    extraData: null,
+    createdAt: new Date("2025-10-07T15:52:13Z"),
+    updatedAt: new Date("2025-10-07T15:52:13Z"),
+    currentLocationName: null,
+    portraitPath: null,
+    ...overrides,
+  };
+}
+
+function renderPane(
+  characters: CharacterListEntry[],
+  imagesByCharacter: Record<number, CharacterImage[]> = {},
+) {
+  const queryClient = new QueryClient({
+    defaultOptions: { queries: { retry: false, staleTime: Infinity } },
+  });
+  const slot = 1;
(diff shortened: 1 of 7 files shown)


pythagorakase:
Review disposition (commit e96c1738):

- Finding 1 (hadImages stale-query race): fixed as suggested — the promote now runs unconditionally on every upload; it is idempotent and no longer reads the client's images query.
- Finding 3 (partial mutation failure leaves queries stale): fixed — invalidation moved from onSuccess to onSettled, so an upload whose promote step rejects still surfaces in list and detail without a reload.
- Finding 2 (join coercion): not applied. The live schema's characters.current_location is bigint, not text — psql \d+ shows `current_location | bigint` with FK fk_characters_current_location to places(id). The `text(...)` typing in ui/shared/schema.ts is the legacy drizzle mismatch (and the reason the pane used to display a bare numeral); the raw SQL join compares bigint = bigint with no coercion, so `::int` would add a cast where none occurs today. The String(...) in the row mapping exists only to satisfy the drizzle-inferred type. Happy to retype the drizzle column in a follow-up, but that widens the blast radius beyond this pane.

Minor notes acknowledged; leaving portraitSrc co-located until a second caller appears.

PR #391, opened 2026-06-12
Wizard: Structured Choices + Freeform Input, Command Bar Retired

pythagorakase:
## Summary

Implements the wizard input redesign: the command bar is gone, replaced by the reader's established interaction pattern — structured choice buttons plus an always-present freeform slot. Model selection leaves the wizard entirely; the backend resolves the configured default, and the TEST mock can no longer be selected implicitly.

## Turn Contract: Was vs Is

No schema change was needed. The wizard turn contract already carried structured choices end to end:

- `WizardResponse` (nexus/api/new_story_schemas.py) enforces `message` + 2-4 `choices` via strict structured output on every conversational turn.
- `/api/story/new/setup/start` already returned `welcome_choices`.

What changed in the contract is **model resolution**, not choices:

- **Was:** `ChatRequest.model` default-factory-filled the wizard default on every request, which made the slot's stamped model dead code in `request.model or slot_model or get_new_story_model()`, and the UI separately injected a model from localStorage on every call.
- **Is:** `model` is an optional explicit override. Omitted -> the slot's stamped model (locked at setup start) -> the configured `wizard.default_model` (`@openai.default`). `/setup/start` resolves the effective model once, stamps it on the slot, and now reports it in the response (`"model": ...`) for verification.

## Root Cause of the TEST Default

Two compounding defaults:

1. **DB stamp:** `nexus.toml [global.model] default_slot_model = "TEST"` stamps fresh/reset slots with `model = TEST` in `global_variables` (via `scripts/new_story_setup.py`). The CLI's empty-slot init forwarded that stamp (`args.model or state.get("model")`) into `/setup/start`, so a plain `nexus continue --slot N` on a fresh slot silently started a TEST wizard. slot 5 was in exactly this state (`model = TEST`).
2. **UI persistence:** the wizard's command-bar model picker read `localStorage["nexus-model"]` (sticky from any prior TEST session; TEST is also in the ModelContext fallback list) and sent it with every request, persisting TEST onto the slot via `upsert_slot`.

Fixes: the wizard UI has no picker and sends no model anywhere (start/chat/bootstrap; the dead `test_mode` flag on bootstrap is also gone); `/setup/start` stamps the resolved real default over the reset stamp; the CLI forwards only an explicit `--model`. TEST remains available as an explicit, backend-only override (`--model TEST`), per the contract that the settings lane owns UI-facing list filtering.

## Visual Minimalism

- Header: phase title ("Setting" / "Character" / "Introduction") replaces "SKALD // SETTING PROTOCOL"; the phase dock in the side panel already carries progress.
- Per-bubble "Skald" label chrome removed; voice is carried by bubble alignment/color.
- Loader text ("PROCESSING...") dropped; spinner only.
- Freeform placeholder matches the reader: "…or something else".

## Verification (Real API Calls, Slot 5)

Worktree backend on :8022, worktree UI on :5022, real `gpt-5.5` turns throughout (no mocks).

CLI (roleplaying a user):
- `nexus continue --slot 5` on the TEST-stamped slot -> wizard started, slot model flipped `TEST -> gpt-5.5`; server log: `Started setup for slot 5 with thread thread_zR9z... (model=gpt-5.5)`.
- `nexus continue --slot 5 --choice 1` -> real frontier turn, 4 new structured choices.
- `nexus continue --slot 5 --user-text "biopunk megacity..."` -> real freeform turn answered in-theme.

Browser (puppeteer-core + system Chrome, headless):
- Slot 5 card -> fresh wizard; welcome renders 4 numbered choice buttons + freeform slot; **no command bar remnant** (asserted).
- Clicking choice 1 advanced the flow (real turn; new choices rendered).
- Typing freeform + Enter advanced the flow (real turn; user text visible in conversation).
- Server log for the UI-driven session (no model in any request): `Started setup for slot 5 ... (model=gpt-5.5)` — resolution proof that the configured default served the turns.

Tests/build:
- `vitest`: 99 passed (incl. 13 new for choice normalization/rendering/interaction).
- `pytest`: 782 passed (incl. new model-resolution contract tests: omitted model stays None; TEST accepted only as explicit override; configured wizard default is never the mock).
- `tsc` clean; `npm run build` green.

## Notes

- `components/StoryChoices.tsx` is now unconsumed (the wizard was its only importer) but is left in place for other lanes.
- `ModelContext` is no longer consumed by the wizard; the settings lane owns its future.
- Reader-owned files (NarrativePane, RightLedger, SettingsPane) untouched; SlotSelector's PR #384 visual-state work preserved.

### DIFF
--- a/nexus/api/narrative_schemas.py
+++ b/nexus/api/narrative_schemas.py
@@ -257,25 +257,16 @@ class SlotLockResponse(BaseModel):
 # =============================================================================
 
 
-def _wizard_default_model_factory() -> str:
-    """Return the resolved wizard default model from nexus.toml.
-
-    Used as the field default when an HTTP client doesn't specify a model.
-    Reading at request time keeps the value in sync with config edits.
-    """
-    from nexus.config import load_settings
-
-    return load_settings().wizard.default_model
-
-
 class ChatRequest(BaseModel):
     slot: int
     message: str
     # thread_id and current_phase are optional - resolved from slot state if not provided
     thread_id: Optional[str] = None
     current_phase: Optional[Literal["setting", "character", "seed"]] = None
-    # Defaults to the resolved wizard.default_model from nexus.toml
-    model: str = Field(default_factory=_wizard_default_model_factory)
+    # Optional explicit override. When omitted, the endpoint resolves the
+    # slot's stamped model (locked at setup start), then the configured
+    # wizard default from nexus.toml.
+    model: Optional[str] = None
     context_data: Optional[Dict[str, Any]] = None  # Accumulated wizard state
     accept_fate: bool = False  # Force tool call without adding user message
     dev: bool = Field(
@@ -288,8 +279,10 @@ class ChatRequest(BaseModel):
 
     @field_validator("model")
     @classmethod
-    def validate_model(cls, v: str) -> str:
+    def validate_model(cls, v: Optional[str]) -> Optional[str]:
         """Validate model against available models from nexus.toml."""
+        if v is None:
+            return v
         from nexus.config import get_available_api_models
 
         available = get_available_api_models()
--- a/nexus/api/new_story_flow.py
+++ b/nexus/api/new_story_flow.py
@@ -37,7 +37,11 @@ def start_setup(slot_number: int, model: Optional[str] = None) -> str:
 
     Args:
         slot_number: Target save slot (1-5)
-        model: Optional model name (defaults to settings.json new_story.model)
+        model: Optional model name. The HTTP setup endpoint resolves the
+            model before calling (explicit override or the configured
+            wizard default), so it always passes a concrete ID; direct
+            callers (e.g., scripts/new_story_cli.py) may pass None to get
+            the nexus.toml wizard.default_model fallback here.
 
     Returns:
         Thread ID for the new conversation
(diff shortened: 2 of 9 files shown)


pythagorakase:
Review dispositions, addressed in 95ecd144:

- **Freeform draft persisting through a choice click (bug):** Fixed — choice selection now routes through `submitChoice`, which clears the draft before submitting. The reviewer's repro covered the button path; the number-key shortcut path had the same leak and is fixed too. Both paths have new test assertions.
- **Resolution-chain test gap:** The inline expression is extracted to `resolve_wizard_model` and the lock precondition to `wizard_model_lock_candidate`, shared by both the standard and streaming endpoints. New unit tests cover explicit-override-wins, slot-stamp fallback, configured-default fallback (and that it is never TEST), and that an omitted model can never be a 409 lock candidate — closing the "lock path is bypassed by every normal UI request" concern by proving the bypass is exactly the designed behavior. Re-verified live after the refactor: fresh wizard start on slot 5 resolved gpt-5.5 and a real choice turn completed.
- **Double resolution in `start_setup`:** The `Optional[str]` parameter is not dead code — `scripts/new_story_cli.py` calls it directly with a possibly-None model, so the signature stays. The docstring now states the contract (HTTP endpoint resolves before calling; direct callers may pass None) and drops a stale settings.json reference.
- **`PHASE_TITLES` seed -> "Introduction":** Intentional — the seed phase is user-facing as "Introduction" everywhere (WizardShell's PHASES, the artifact modal). Added a comment at the definition.
- **TEST in registry (informational):** `nexus.toml` ships the test provider entry, so the explicit-override test is valid in CI; added a comment noting the registry requirement.

PR #392, opened 2026-06-12
Reader: Interactive Story Tree, Chunk Navigation, and Heading De-Duplication

pythagorakase:
## Removals (What + Where)

| Item | Where | Change |
|---|---|---|
| Chunk-slug supra-heading (`[ SCENE S5·E6·S1 ]`) | `NarrativePane` scene header | Removed; the tree's slug row is the single surface for chunk identity |
| `HIERARCHY` label + `S·E` caption | `RightLedger` tree section head | Removed; the expandable tree is self-evident |
| Active-cast listing | `NarrativePane` heading area AND `RightLedger` `SCENE CAST` section | Both removed; `chunk_character_references` tracking untouched backend-side |
| `SESSION TELEMETRY` label + `SKALD · IDLE/LIVE` caption | `RightLedger` | Removed; the phase stream itself is the content, and the top bar already carries SKALD status |
| `ELAPSED` caption | telemetry block | Removed; a ticking clock beside a progress strip restates itself |
| Internal module names (`Calling LORE / LOGON…`, `SKALD ·` prefix) | phase labels (rail + in-reader status) | Replaced with plain language (`Assembling context…`, `Writing…`, …) shared via `PHASE_LABELS` |
| Per-row coordinate (`S5·E6·S1`) duplicating the slug (`S05E06_001`) | tree chunk rows | Rows now show the slug only |

## Navigation Design and Granularity

**Display unit = chunk = scene.** Verified against the live corpora: in save_01 every episode has exactly as many scenes as chunks (`count(*) == count(DISTINCT scene)` across all 70 episodes), and the same holds in the modern slots — `scene` is 1:1 with chunk. So "previous/next scene" and "previous/next chunk" are the same step; `<` / `>` step chunk-by-chunk and the tree provides direct jumps (season → episode → scene slug).

**The ladder model** (documented in `ui/client/src/lib/narrative-nav.ts`): reading positions are a ladder in story order with the live frontier as the top rung. Live mode is the unchanged reading surface (episode stream, pending chunk, choices/freeform). Every other rung displays one committed chunk read-only, with a bronze edge marker (vs. the magenta live marker), the choice/input block replaced by the navigation context, and zero writes — pending choices and story position are untouched.

- `<` / `>` render in `var(--font-menu)` (Cinzel under Veil; per-theme via the existing token) at the top and bottom of the displayed chunk.
- `<` grays out at the very first chunk; `>` grays out at the live frontier.
- Navigation is symmetric: with a pending chunk, `<` from live lands on the latest committed chunk; without one (slot 1), `<` skips past the on-screen latest chunk. One `<` then one `>` always returns to the start.
- `»` (history mode only) jumps straight back to the frontier from any depth.
- The tree highlights the reading position; the frontier row (the `PENDING` row, or the latest chunk row when nothing is pending) keeps a persistent magenta tick, and clicking it restores the live surface with its input affordances.
- The tree is built from `chunk_metadata` rather than the `seasons`/`episodes` tables, which are unpopulated in the modern slots; the active path auto-expands as the reading position moves.

## New Endpoint Contract

Both live in the Express read layer (`ui/server/routes.ts` + `storage.ts`), which serves every other read route feeding the reader and the tree (`latest-chunk`, episode chunks, context, adjacent); the FastAPI service in `nexus/api/` handles only generation/slot/wizard flows, so no endpoint was added there. Both are read-only, slot-aware, and fail loud:

- `GET /api/narrative/outline?slot=N` → `[{ id, season, episode, scene, slug }, …]` per committed chunk in story order. `400` on malformed slot.
- `GET /api/narrative/chunks/:chunkId?slot=N` → the committed chunk with metadata (same shape as `latest-chunk`). `404` when absent, `400` on malformed id.

## Telemetry Signal

The telemetry block renders only while `engine.isGenerating` is true — i.e. the current websocket phase from `/ws/narrative` is in `ACTIVE_GENERATION_PHASES`. It appears on `initiated` and unmounts entirely at `complete`/`error` (no placeholder box at idle).

## Legacy No-Choices Input

When slot state carries no structured choices (slot 1's whole corpus, e.g. chunk 1425 which ends mid-beat), the freeform input renders with no placeholder and is preemptively focused, so the blinking caret is the invitation. With choices present, the `…or something else` escape-hatch placeholder is unchanged. Fixing this surfaced a latent bug: programmatic focus matches `:focus-visible`, which pointer focus never did, so the shadcn Textarea ring appeared; it is now suppressed at higher specificity.

## Verification

- `vitest`: 20 new tests in `narrative-nav.test.ts` (tree grouping, ladder boundary states incl. first-chunk/frontier/pending asymmetry/symmetry, inert-on-disagreement, the no-choices input rule); 106/106 pass.
- `npm run build` green; `tsc` clean.
- Real browser (headless system Chrome, dev server on :5023):
  - **Slot 1 (legacy, read-only)**: 26/26 checks — clean heading (no eyebrow, no cast), no rail labels, no idle telemetry, tree expand/collapse, jump to S01E01_005, `<`/`>` stepping across sparse ids, `<` grayed at chunk 1, `»` return restoring input affordances, chunk-1425 freeform focused with no placeholder and no fake choices.
  - **Slot 2 (modern, writable working copy)**: 4 structured choices intact, placeholder intact, no autofocus theft, `PENDING` frontier row highlighted, ladder round-trip; then **one real generation turn** driven through the UI — telemetry appeared in flight (`Assembling context… → Writing… → Processing response…`, ticking clock), vanished at completion, new pending chunk revealed, tree refreshed.
  - Note: the turn was driven on slot 2 instead of slot 5 because slot 5 was mid-wizard under a concurrent lane, and slots 3/4 hold stale pre-schema incubator data (`episode_transition: "begin"`) that the backend's `ChronologyUpdate` validator now rejects — a pre-existing data issue outside this lane, surfaced loudly by the 500 (no UI fallback added).


pythagorakase:
Review dispositions (commit cc7f85b6):

1. **SQL indentation** — fixed; `getChunkById` now uses uniform 8-space column indentation. (The ragged style was inherited from the adjacent pre-existing queries, which were left untouched.)
2. **Disguised inner join** — fixed; `getChunkById` uses a true `JOIN` and the metadata mapping is unconditional, with a comment stating the invariant.
3. **Outline invalidation** — no change needed. `useNarrativeEngine.invalidateNarrativeQueries` runs on the `complete` phase and invalidates by predicate every query whose key starts with `/api/narrative`, which includes `["/api/narrative/outline", slot]`. Verified live during the slot-2 real-turn pass: the tree gained the newly committed chunk and the PENDING row moved immediately after generation, before any window-focus refetch.
4. **Fragment keys** — fixed; `historicalParts` renders key by index.
5. **Historical loading state** — fixed; a `pane-notice` LOADING state renders while the chunk query is in flight, matching the component's existing notice patterns.
6. **autoFocus / effect split** — documented at the effect: the attribute covers initial mount, the effect re-focuses after each completed turn.
7. **OutlineRow duplication** — documented; the client type's doc comment now names `ui/server/storage.ts` as the server source to keep in lockstep.

Also merged `origin/main` (PRs #389–#391); the single conflict was additive in `narrative-api.ts` (new fetchers vs. the `CharacterListEntry` signature change) and resolved by keeping both. Full suite re-verified post-merge: tsc clean, 129/129 vitest, production build green, real-browser regression passes on slots 1 and 2.

PR #395, opened 2026-06-12
Map: Land/Sea Contrast and Slot 5 Landmass Diagnosis

pythagorakase:
## Summary

Fixes the two map complaints (slot 1 land barely visible; slot 5 no landmasses at all) plus the deeper rendering bug both were symptoms of, and sweeps the pane for visual-verbosity violations.

Per the conductor correction on this PR: all NEXUS worlds use real-Earth geography by design, so the Earth coastline renders for every slot. The abstract survey treatment from the original lane brief was implemented, then removed in `cb526417` — slot 5 was a straight render bug, now fixed.

## Slot 5 Root Cause (Diagnosed, Not Guessed)

**Render bug, not missing data.** Coastline data is not per-slot: it is a client-bundled Natural Earth 110m outline (`ui/client/src/lib/world-outline.ts`), present for every slot — so the conductor's hypothesis (1) (per-slot seeding gap) is disproven; no seeding fix applies. The cause was hypothesis (2), in two stacked parts:

1. **Spherical fit winding (the big one — also silently broken on slot 1).** `useGeoProjection` fed `fitSize` a bounding *polygon*; d3-geo polygons are spherical, and the ring's winding made d3 read it as the box's sphere-complement — so `fitSize` always fit the entire globe. Measured: fitted projection scale 163.29 (exactly the whole-sphere fit) vs 25,235 with correct winding. On slot 1 this masqueraded as a plausible whole-world view; on a generated slot (one charted place), the region rendered as a ~6 px speck — a blank void.
2. **Degenerate single-place bounds.** With one charted place, `computeMapBounds` collapsed to a ~0.0001-degree box (a meters-wide viewport). Masked by bug 1; would have produced the same blank map on its own.

Evidence (row counts at diagnosis):
```
save_01: places=77 (50 charted), zones=34 (20 with boundary), layers: Earth(planet)
save_05: places=1  (1 charted),  zones=1  (1 with boundary),  layers: Veyport(planet)
```

## Fixes

- **Spherical fit winding (failure mode 5)**: `fitSize` now receives winding-free corner points via `boundsToFitObject` (`MultiPoint` of the bbox corners). Documented in `docs/maptab_rebuild_spec.md` §2.2 + new §6.5; regression tests added.
- **Degenerate bounds**: `computeMapBounds` enforces a 1-degree minimum span, so a lone charted place sits in a regional frame showing the real coastline around its coordinates. Zone boundaries stay out of the fit deliberately: legacy slots carry near-global zones (Pacific Ocean, Antarctic) that would blow the frame back to world scale.
- **Land/sea contrast (all themes)**: land was brass at ~3% effective opacity. Now token-driven mixes on `.mappane-canvas` (`--map-land`: 16% brass into `--bg-elev-1`, opaque; `--map-coast`: 50% brass stroke), so Veil/Gilded/Vector each keep correct contrast from their own palette.
- **API defect**: `GET /api/zones` now serves `boundary` as `ST_AsGeoJSON(...)::json` (previously unrenderable EWKB hex text). Unblocks the spec §8 zone-boundary future work; consumed by nothing yet.
- **Visual-minimalism sweep**: removed the `[ SURVEY ]` box, the "Drag to pan · scroll to zoom" instruction line, and the three-entry legend; zoom readout is just `1.00x`; current-place readout is pin glyph + name (no "THE STORY IS HERE" caption); map labels, zone headers, and dialog titles no longer uppercase content; dialog eyebrow shows the place type only; empty state is `[ UNCHARTED ]` over the full Earth map.

## Review Disposition

- Codex P2 (survey polygon winding) and P2 (Earth-name matching): both targeted the layer-gating/survey path removed in `cb526417` — moot.
- Claude review BLOCKING (conductor contradiction), High (`worldIsEarth` exact match), Medium (layers error softness, MemStorage stub): all resolved by removing that path entirely. Low (`Zone.boundary: any`): typed `object | null`. Low (`parseInt` radix on the new endpoint): endpoint removed.

## Verification

- `vitest`: 90 passed, including new suites for `boundsToFitObject` (failure mode 5) and the min-span regression; all four original failure-mode suites green.
- `tsc` clean; production build green.
- Real-browser screenshots (puppeteer + system Chrome, port 5025), before/after in all three themes:
  - Slot 1: before, faint land silhouettes on an (erroneously) world-fit view; after, region-fit to the charted places with clearly legible land in Veil (plum), Gilded (brass), Vector (teal).
  - Generated world (slot 3, identical shape to the diagnosed slot 5 world; slot 5 was externally reset mid-task): before, blank void with a lone beacon; after, the real Adriatic/Balkan coastline framing the beacon at its true coordinates.
  - Slot 5 (now empty): full Earth map + `[ UNCHARTED ]` empty state.
  - Interaction check: cursor-anchored wheel zoom, drag-pan, pin click -> details dialog all working.

## Data/Schema Impact

No migrations. `/api/zones` response shape for `boundary` changes from EWKB hex string to GeoJSON object (previously consumed by nothing).

Note for the conductor: the freshly reset save_05 currently has zero `layers` rows — if the wizard is expected to always create the world layer row, that is a wizard-side concern outside this lane's territory; the map no longer depends on it.

### DIFF
--- a/docs/maptab_rebuild_spec.md
+++ b/docs/maptab_rebuild_spec.md
@@ -71,7 +71,7 @@ divide by `zoom` to stay visually constant.
 import { geoEquirectangular, geoPath } from "d3-geo";
 
 const proj = geoEquirectangular();
-proj.fitSize([mapDimensions.width, mapDimensions.height], boundingFeature);
+proj.fitSize([mapDimensions.width, mapDimensions.height], fitObject);
 const transformCoordinates = (lng: number, lat: number) => proj([lng, lat]);
 const pathGenerator = geoPath().projection(proj);
 ```
@@ -79,9 +79,11 @@ const pathGenerator = geoPath().projection(proj);
 - `geoEquirectangular` is the right choice for a fictional-world map —
   it's a simple linear mapping of lat/lng to pixels, no distortion to
   reason about, no globe wrap-around concerns.
-- `fitSize` accepts a "bounding feature" (a GeoJSON polygon describing
-  the area to include) and computes scale/translate so that polygon fills
-  the canvas.
+- `fitSize` accepts a GeoJSON object describing the area to include and
+  computes scale/translate so that area fills the canvas. **Do not pass a
+  bounding polygon** — see §6.5 (spherical fit winding). Pass winding-free
+  corner points instead (`boundsToFitObject` in `lib/map-geometry.ts`):
+  `{ type: "MultiPoint", coordinates: [[minLng, minLat], [maxLng, maxLat]] }`.
 - `geoPath()` returns a path generator that converts arbitrary GeoJSON
   geometry → SVG path strings, used for the continent outlines.
 
@@ -296,7 +298,7 @@ to feel like a live window, not a navigable document.
 
 ## 6. Known Subtleties (The Historical Failure Modes)
 
-These are the four areas where past rebuilds went wrong:
+These are the five areas where past rebuilds went wrong:
 
 ### 6.1 Label collision detection
 
@@ -366,6 +368,28 @@ lng ∈ [-180,180]. Out-of-range coords are silently dropped with
 Currently only `Point` is rendered. `Polygon` / `LineString` geometries
 are returned by the API as-is but are dropped at the renderer; future
 work would compute centroids for Polygon and midpoints for LineString.
+
+### 6.5 Spherical Fit Winding
+
+d3-geo polygons are **spherical**: a ring's winding order decides which
+side is the inside. A bounding-box ring wound the wrong way denotes the
+*complement* of the box (the entire globe minus it), so `fitSize` quietly
+fits the whole world and every regional map renders as a world-scale
+speck. The U4 rebuild shipped with exactly this bug — on an Earth slot it
+masqueraded as a plausible whole-world view; on a generated world (one
+zone, one place) it produced a blank void.
+
+**The fix**: never feed `fitSize` a bounding polygon. Use winding-free
+corner points (`boundsToFitObject` in `lib/map-geometry.ts`):
+
+```ts
+proj.fitSize([width, height], {
+  type: "MultiPoint",
+  coordinates: [[minLng, minLat], [maxLng, maxLat]],
+});
+```
+
+Regression-tested in `map-geometry.test.ts` ("failure mode 5").
 
 ---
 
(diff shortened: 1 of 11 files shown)


pythagorakase:
**Conductor correction — from the user, supersedes the lane brief's hypothesis (a):** Generated worlds are NOT permitted to lack coastline geometry. ALL NEXUS settings are constrained to Earth-shaped geography by design — GIS coordinates are real-Earth positions regardless of genre (a sci-fi vessel and a feudal castle at [7.474, -78.081] share the same tropical Pacific coast). This is a deliberate cognitive-offloading mechanism for LLM spatial reasoning: latitude implies climate, terrain implies travel modes.

Therefore slot 5 showing no landmasses is a straight BUG, not a candidate product decision, and the 'honest abstract sea' treatment described in the brief must NOT ship. Diagnose along these lines instead: either (1) Earth coastline data lives per-slot and is missing from the fresh-slot path — in which case fix the seeding (NEXUS_template and/or new_story_setup) so every future slot is born with Earth geography, following the template-refresh procedure in CLAUDE.md — or (2) the data/render path fails for slot 5 specifically (projection, bbox, layer ordering, lng/lat order per the spec's failure modes). Every slot, present and future, must render Earth landmasses. Adjust the PR accordingly before merge.

pythagorakase:
Conductor correction applied in `cb526417`:

- The abstract survey treatment is removed; the Natural Earth coastline now renders unconditionally for every slot, present and future.
- On your hypothesis (1): coastline data is not per-slot — it is the client-bundled Natural Earth outline (`world-outline.ts`), already present for all slots — so there is nothing to seed in `NEXUS_template`/`new_story_setup.py` for landmasses. The blank map was hypothesis (2), twice over: a spherical winding bug that made `fitSize` fit the whole globe for every slot since the rebuild, plus degenerate single-place bounds zooming the viewport to a meters-wide patch. Both fixed and regression-tested (spec §6.5, "failure mode 5").
- Verified on a generated world (slot 3, same shape as the diagnosed slot 5 world): its lone charted place at [18.5, 42] now renders against the real Adriatic/Balkan coastline. The reset save_05 (zero places) shows the full Earth map with the uncharted empty state.
- Both Codex P2s and the Claude review's blocking/high/medium items targeted the now-removed gating path; the surviving low item (`Zone.boundary` typing) is addressed.

One observation outside this lane's territory: the freshly reset save_05 has zero `layers` rows. The map no longer depends on that table, but if the wizard is expected to always create the world layer row, that may warrant a wizard-side check.

PR #411, opened 2026-06-18
Fix packaged desktop runtime working directory

pythagorakase:
## Summary
- Capture the Tauri desktop config directory at build time.
- Resolve bundled relative `workingDirectory` values from that captured source directory instead of Finder's launch cwd.
- Fail loudly if a bundled relative working directory has no anchor instead of letting Poetry run from `/`.

## Why
A dev-built `.app` launched from Finder was running `poetry run nexus --json up` with cwd `/`, producing:

`Poetry could not find a pyproject.toml file in / or its parents`

The bundled config contains `"workingDirectory": "../.."`, which is correct relative to `ui/src-tauri/nexus.desktop.json` but lost its anchor once embedded into the binary.

## Validation
- `cargo fmt --manifest-path ui/src-tauri/Cargo.toml --check`
- `cargo test --manifest-path ui/src-tauri/Cargo.toml`
- `npm --prefix ui run desktop:build`
- Verified the built binary contains the captured config anchor: `/Users/pythagor/nexus/ui/src-tauri`

Fresh bundle: `/Users/pythagor/nexus/ui/src-tauri/target/release/bundle/macos/NEXUS.app`

### DIFF
--- a/ui/src-tauri/build.rs
+++ b/ui/src-tauri/build.rs
@@ -1,3 +1,8 @@
 fn main() {
+    println!("cargo:rerun-if-changed=nexus.desktop.json");
+    println!(
+        "cargo:rustc-env=NEXUS_BUILD_CONFIG_DIR={}",
+        std::env::var("CARGO_MANIFEST_DIR").expect("CARGO_MANIFEST_DIR is set")
+    );
     tauri_build::build()
 }
--- a/ui/src-tauri/src/lib.rs
+++ b/ui/src-tauri/src/lib.rs
@@ -441,6 +441,18 @@ fn resolve_working_directory(
     config: &DesktopConfig,
     source_path: Option<&Path>,
 ) -> Result<PathBuf, String> {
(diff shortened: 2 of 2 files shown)


pythagorakase:
Pushed `6b49bbba` addressing Claude's review polish:

- Canonicalizes the `discover_repo_root()` / `current_dir()` fallback path consistently.
- Avoids evaluating the baked config-dir anchor when an explicit config source already exists.
- Adds a regression test proving `source_path` wins over the baked config dir.
- Makes the temp-dir test cleanup RAII-backed.

Validation repeated after the follow-up:
- `cargo fmt --manifest-path ui/src-tauri/Cargo.toml --check`
- `cargo test --manifest-path ui/src-tauri/Cargo.toml` -> 8 passed
- `npm --prefix ui run desktop:build`
- Verified the rebuilt release binary still contains `/Users/pythagor/nexus/ui/src-tauri` as the bundled config anchor.

PR #412, opened 2026-06-18
Restart stale desktop runtime services

pythagorakase:
## Summary
- Add configurable desktop `restartArgs` defaulting to `["--json", "restart"]`.
- If the desktop shell cannot attach and `nexus up` reports an existing managed service pid, run `restartArgs` and wait for runtime readiness.
- Document the conditional restart ownership behavior.

## Why
After the cwd fix, the packaged app correctly runs `poetry run nexus --json up` from the repo root. The next startup edge case is a stale or unhealthy managed runtime pid:

`Service 'gateway' is already running (pid 56729). Use 'nexus restart' or 'nexus down' first.`

The shell should repair that state by restarting the managed runtime instead of surfacing the raw `up` failure.

## Validation
- `cargo fmt --manifest-path ui/src-tauri/Cargo.toml --check`
- `cargo test --manifest-path ui/src-tauri/Cargo.toml` -> 9 passed
- `npm --prefix ui run desktop:build`
- Verified the rebuilt app binary contains the restart fallback strings.

Fresh bundle: `/Users/pythagor/nexus/ui/src-tauri/target/release/bundle/macos/NEXUS.app`

### DIFF
--- a/docs/desktop.md
+++ b/docs/desktop.md
@@ -62,9 +62,10 @@ Poetry from the repo root. Packaged or globally installed environments can set
   the runtime on quit.
 - If the shell runs the start command, it records ownership and runs the stop
   command on quit.
-- If `up` reports an existing managed service pid but `/runtime/status` is not
-  healthy, the shell runs `restartArgs`, waits for readiness, and then treats
-  the restarted runtime as shell-owned.
+- If the first `/runtime/status` probe is unhealthy, `up` reports an existing
+  managed service pid, and follow-up probes still cannot attach, the shell
+  runs `restartArgs`, waits for readiness, and then treats the restarted
+  runtime as shell-owned.
 - The shell talks only to the runtime contract: origin, status surface, auth
   header, and CLI entrypoint.
 - If startup fails before a window can be shown, the shell writes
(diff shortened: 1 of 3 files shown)


pythagorakase:
Pushed `8c97f1be` addressing Claude's review feedback:

- `service_already_running_error` now parses the CLI JSON `error` payload when present before checking the existing-service message.
- Added a final `/runtime/status` probe immediately before issuing `restartArgs`.
- Documented why restart ownership is set before readiness is confirmed: the shell created the replacement process and should be able to clean it up if readiness times out.
- Added edge-case coverage for empty strings, whitespace, and partial existing-service messages.
- Clarified the docs to describe the initial unhealthy probe, the startup-race re-probe, and the final pre-restart probe.
- Restart success now updates the loading screen even if the CLI returns no output.

Validation repeated:
- `cargo fmt --manifest-path ui/src-tauri/Cargo.toml --check`
- `cargo test --manifest-path ui/src-tauri/Cargo.toml` -> 9 passed
- `npm --prefix ui run desktop:build`
- Verified the rebuilt app binary contains the updated restart fallback strings.

PR #424, opened 2026-07-02
Add Orrery adjudication history (migration 063 + enriched logging + endpoint)

pythagorakase:
## Summary

Step 6 of the audit-dashboard sequence in `docs/orrery_audit_dashboard_notes.md`: the adjudication-history layer. Recon against live schemas confirmed all five recorded-history holes the spec names; this PR closes four of them and ships the ledger endpoint (the fifth — ratifications leave no row — is by design reconstructed as committed-resolutions-minus-replacements).

### Migration `063_orrery_adjudication_history.sql` (applied locally to template + all slots)

- `orrery_adjudication_log` + `actor_entity_id`, `bindings` — a deferred proposal's subject was previously recoverable only from the incubator JSONB, which `DELETE FROM incubator` wipes at commit. Stamped at insert while the draft is in hand.
- **`orrery_scene_pressures`** — pressures were generated (`resolver.py`), rendered, and dropped; zero persistence. Now written at commit.
- **`orrery_prompt_exposures`** — the storyteller prompt renders only the first N proposals/pressures with ratify-by-omission semantics, and nothing recorded which were shown. Now the rendered slice is logged per tick.
- Pre-063 rows keep NULLs; the endpoint's `epoch` block quantifies enrichment coverage instead of implying it.

### Writer changes (`events.py`, sync + async twins kept in lockstep)

- Pressure/exposure inserts run **before** the `has_resolutions` early return — a tick can carry scene pressure with zero resolutions (save_02's head tick does exactly this).
- **ON CONFLICT silent-skip closed**: a replace-with-delta ruling whose resolution insert conflicts previously vanished from history entirely (recon confirmed the `continue` at the exact lines the spec predicted); it now lands with `applied_resolution_id` resolved from the existing row.
- All inserts idempotent for re-commit safety.

### Config

The hardcoded `[:5]` render slices in `logon_utility.py` become `[orrery.prompt] max_rendered_proposals` / `max_rendered_pressures` (per the no-hardcoded-tunables directive), and both commit handlers pass the same config into the exposure logger — one knob, two sites, no drift between what was rendered and what was recorded.

### Endpoint

`GET /api/dev/orrery/history/adjudications?slot=&template_id=` — action rates faceted by `adjudication_source` (machine-inferred `structured_state_update` never blends with authored rulings), the lifecycle funnel (committed → promoted/skipped → narrated, `ratified_committed` derived), **defer streaks** (the spec's "single most diagnostic derived metric": consecutive defers per `proposal_id` with outcome ratified/replace/void/open), exposure + pressure ledger stats, and the epoch block. Live smoke reproduces the recon oracle exactly (save_02: 15/1/3 defer/replace/void over 66 committed; save_05: 13/3/7 over 4 — including `surveil` deferred 3 consecutive ticks before a replace).

### Tests (no mocks)

- A **rollback-transaction commit test** drives the real `commit_orrery_tick_sync` against save_02 — real SQL, real constraints, always rolled back, with a post-rollback probe confirming zero persistence. Pins subject stamping, pressure persistence, exposure cap semantics (a draft beyond the cap is not recorded), the conflict-path log fix with the existing resolution id, and idempotent re-commit.
- History payloads verified against direct SQL oracles on save_02 + save_05; streak arithmetic (`sum(lengths) == total defers`) and funnel monotonicity (`narrated ≤ promoted ≤ committed`) enforced; joint non-vacuity guard.
- The recording-cursor test that asserted the old silent-skip behavior now pins the new contract.
- Full suite: 885 passed offline; live suites pass with only the known pre-existing `test_pair_tag_writer` failure (reproduces on origin/main). The four `events.py` mypy arg-type notes at the `_validate_*` calls also predate this change.

Next in sequence: step 7 (reconstructability: forward-fix migrations, Skald-side delta logging, genesis snapshots).

### DIFF
--- a/docs/orrery_packages.md
+++ b/docs/orrery_packages.md
@@ -2153,6 +2153,7 @@ the seeding migrations to confirm catalog ↔ schema agreement:
 - `migrations/056_orrery_routine_anchors.py`
 - `migrations/057_orrery_need_bodyform_applicability.py`
 - `migrations/059_orrery_social_travel_event.py`
+- `migrations/063_orrery_adjudication_history.sql`
 
 ### Tags queried as durable (via `has_tag` / `lacks_tag` / `has_any_tag`)
 
--- a/migrations/063_orrery_adjudication_history.sql
+++ b/migrations/063_orrery_adjudication_history.sql
@@ -1,0 +1,76 @@
+-- 063_orrery_adjudication_history.sql
+--
+-- Adjudication-history enrichment for the Orrery audit dashboard (step 6 of
+-- docs/orrery_audit_dashboard_notes.md). Closes the recorded-history holes:
+-- the log's subject is recoverable after the incubator row is deleted
+-- (actor_entity_id + bindings), scene pressures persist past commit, and the
+-- set of proposals actually rendered to the storyteller is recorded so
+-- "ratified by omission" and "never shown" stop being indistinguishable.
+-- Pre-migration rows keep NULLs; consumers treat NULL as the pre-063 epoch.
+
+ALTER TABLE orrery_adjudication_log
+ADD COLUMN IF NOT EXISTS actor_entity_id bigint REFERENCES entities(id),
+ADD COLUMN IF NOT EXISTS bindings jsonb;
+
+COMMENT ON COLUMN orrery_adjudication_log.actor_entity_id IS
+    'Acting entity of the adjudicated draft, stamped at insert while the draft is in hand; NULL on rows written before migration 063 (the incubator proposal that carried it is deleted at commit).';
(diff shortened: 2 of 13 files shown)


pythagorakase:
Both findings addressed: the conflict-branch replace log write is guarded by an existence check on (tick_chunk_id, proposal_id, action='replace') — a round-3 re-commit assertion in the rollback test now pins that repeated re-commits leave exactly one replace row (so `ratified_committed` can't go negative) — and both commit handlers' log gate + message body include the new scene_pressure/prompt_exposure counts so pressure-only ticks are no longer invisible in the commit log. 58/58 live writer tests pass.

PR #425, opened 2026-07-03
Forward-fix tag provenance (migration 064): chunk-keyed bestowals + logged clears

pythagorakase:
## Summary

Step 7's forward-fix migrations from `docs/orrery_audit_dashboard_notes.md` ("Reconstruction Sufficiency") — going forward, every tag bestowal carries the chunk that caused it and every clearance is logged, on every path. These are the prerequisites for honest as-of rewind of tags/pair tags (the fast-follow the spec gates on this landing).

### Migration `064_tag_provenance_forward_fix.sql` (applied to template + all slots)

- `source_chunk_id` on `entity_tags` **and** `entity_pair_tags` — defining the **"exact"** provenance tier the hover-audit reserved since v1. Pre-064 rows stay NULL honestly.
- `tag_clearance_log` becomes polymorphic (nullable `entity_tag_id` + new `entity_pair_tag_id`, CHECK exactly-one): recon confirmed the old NOT NULL single-entity FK made pair-tag clears *structurally unloggable*.
- `entity_tags_current` re-created to project the new column (it's a column-projection view; recon's paraphrase of the view definition would have broken `CREATE OR REPLACE` on a column alias — validated against `pg_get_viewdef` first).

### Writers

- **Resolver bestowals** stamp `source_chunk_id` + the bestowing chunk's world time — closing the resolver-template NULL leak the notes cite ("the apply path already holds it" — confirmed; it just never forwarded it). New strict `_chunk_world_time_*` helpers return NULL rather than the wall-clock fallback the existing helper uses: provenance must not lie.
- **Every tag_writer clear path now logs** (`bestowal.tags_to_clear`, public clears, exclusive-ladder sibling clears, status-ladder replacement, pair-tag clears) with mechanism `authored`, world time, chunk id, and a jsonb reason. Resolver inbound pair-tag clears log with `entity_pair_tag_id`.
- **Skald commit handlers** thread the committing chunk through `apply_state_updates` → `apply_tag_bestowal` — closing the `skald_inline` NULL-world-time leak (19 of save_05's 30 NULL bestowals). Bestowals now prefer the committing chunk's clock over the legacy global-max default.
- `entity_context` provenance goes three-tier: `exact` / `approximate` / `unknowable`, with `source_chunk_id` in every row.

### Tests (no mocks for the new behavior)

Live rollback-transaction tests against save_02 drive the real resolver commit and tag_writer: bestowal stamping (chunk id + world time match `chunk_metadata`), clearance logging on single/pair/inbound paths, and the end-to-end `exact` tier read back through `entity_context` inside the same uncommitted transaction. Seven legacy fake-cursor fixture files updated to model the new SQL (RETURNING clears, 6-param inserts, clearance-log writes) with new-contract assertions **added** — idempotent clears log exactly once, status replacement logs both displaced siblings — and none weakened.

Full suite: 967 passed offline; live orrery+api suites 695 passed with only the known pre-existing `test_pair_tag_writer` failure (reproduces on origin/main).

Remaining step-7 slices (separate PRs): 7b Skald-side chunk-keyed delta logging on the on-screen commit path, 7c genesis snapshot + checkpoints, 7d relationship versioning — 7c/7d design will be drafted for review rather than landed unilaterally.

### DIFF
--- a/migrations/064_tag_provenance_forward_fix.sql
+++ b/migrations/064_tag_provenance_forward_fix.sql
@@ -1,0 +1,70 @@
+-- 064_tag_provenance_forward_fix.sql
+--
+-- Forward-fix migrations for tag provenance (step 7 of
+-- docs/orrery_audit_dashboard_notes.md, "Reconstruction Sufficiency").
+-- Going forward, bestowals carry the chunk that caused them and clearances
+-- are logged on every path — the prerequisites for honest as-of rewind of
+-- tags and pair tags. Pre-064 rows keep NULLs: per-row provenance stays
+-- "approximate" (world time only) or "unknowable" (neither), and the
+-- hover-audit renders the epoch per row.
+
+-- 1. Bestowal chunk keys. The resolver's apply path and the Skald commit
+--    path both hold the committing chunk id; these columns give them a
+--    place to put it. "Exact" as-of provenance is defined as
+--    source_chunk_id IS NOT NULL.
+ALTER TABLE entity_tags
+ADD COLUMN IF NOT EXISTS source_chunk_id bigint REFERENCES narrative_chunks(id);
+
+COMMENT ON COLUMN entity_tags.source_chunk_id IS
+    'Chunk whose commit bestowed this tag. NULL on rows written before migration 064 and on pre-chunk bestowals (wizard seeding, entity-reference resolution); non-NULL is the "exact" as-of provenance tier.';
+
+ALTER TABLE entity_pair_tags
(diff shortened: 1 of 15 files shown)


pythagorakase:
P2 addressed: the world-time defaulting in all five public bestowal paths is now branch-exclusive — a chunk-keyed row takes the bestowing chunk's clock or stays NULL, and only chunk-less callers fall back to the legacy global-max. A new regression test inserts a metadata-less narrative chunk inside the rollback transaction and pins that its bestowal stores source_chunk_id with NULL applied_at_world_time. (Claude review's shared-helper duplication note acknowledged — the two _chunk_world_time twins live with their respective driver styles for now; a future consolidation candidate.)

Issue #426, opened 2026-07-03
Reconstructability: Skald delta logging, checkpoints, relationship versioning (spec step 7b-7d)


pythagorakase:
Design proposal for the remaining reconstruction-sufficiency slices from `docs/orrery_audit_dashboard_notes.md` (step 7). Step 7a (migration 064: chunk-keyed bestowals + logged clearances) is PR #425. Posting the rest for review before implementation because the schema choices here are long-lived.

## 7b — Skald-side chunk-keyed delta logging (the "single biggest hole")

The on-screen commit path (`apply_state_updates` in both commit handlers) overwrites `characters.current_location` / `current_activity` and place/faction scalars **with no record of any kind** — invisible to any replay, forever. Proposal:

**New table `state_delta_log`** (migration 065):

```sql
CREATE TABLE state_delta_log (
    id             bigserial PRIMARY KEY,
    source_chunk_id bigint NOT NULL REFERENCES narrative_chunks(id),
    writer         text NOT NULL CHECK (writer IN ('skald_state_update', 'wizard_seed')),
    entity_id      bigint REFERENCES entities(id),
    field          text NOT NULL,          -- e.g. 'characters.current_location'
    old_value      jsonb,                  -- read in the same UPDATE via RETURNING/CTE
    new_value      jsonb,
    created_at     timestamptz NOT NULL DEFAULT now()
);
```

- Written inside `apply_state_updates` for every scalar it touches, one row per changed field, with the old value captured in the same statement (`UPDATE ... RETURNING` old via CTE) so the log is atomic with the write.
- Orrery-side deltas already live in `orrery_resolutions.state_delta`; this table deliberately does **not** duplicate them — `writer` distinguishes provenance, and replay unions the two ledgers ordered by chunk.
- With this, position/activity become replayable going forward, per the spec.

**Open question:** log unchanged writes (UPDATE to the same value) or only actual changes? Only-changes keeps the ledger tight but requires the old-value read; log-everything is simpler and replay-safe either way.

## 7c — Genesis snapshot + periodic checkpoints

Purpose: bound replay cost and give mature slots an explicit instrumentation-era boundary (exact after the checkpoint, approximate before).

[...]

PR #428, opened 2026-07-03
Add reconstructability layer (migration 065): delta log, checkpoints, relationship triggers

pythagorakase:
## Summary

Steps 7b–7d of the reconstruction-sufficiency bar, implementing the maintainer decisions on #426: **log everything**, **JSONB checkpoints**, **auto-checkpoint in the commit path**, **retro-fitted genesis**, **trigger-based relationship versioning**.

### 7b — `state_delta_log` (the spec's "single biggest hole")

Every Skald-side scalar write — character `emotional_state`/`current_activity`/`current_location`, place `current_status` — now lands in a chunk-keyed ledger from both commit handlers, with entity-id resolution. The on-screen commit path previously moved characters with no record of any kind, invisible to any replay forever. Replay contract: latest checkpoint + `orrery_resolutions.state_delta` + `state_delta_log` in chunk order.

### 7c — `state_checkpoints`

One JSONB document per checkpoint covering the full mutable surface (active tags + pair tags, character/place scalars, all three relationship tables, need states, travel states, routine anchors). Auto-taken every `[orrery.reconstruction] checkpoint_interval_chunks` (default 25, 0 disables) inside the commit transaction, idempotent per (chunk, label). `scripts/checkpoint_state.py` handles manual/genesis snapshots — **genesis checkpoints are already taken on all five slots** (~170KB on the dense slot), marking the instrumentation-era boundary: exact after, approximate before.

### 7d — `relationship_versions` (trigger-enforced)

`BEFORE UPDATE OR DELETE` row triggers on the three current-state relationship tables write the full pre-image. Implementation note: those tables use **composite natural keys** (no surrogate id — caught before first trigger fire; the `old_row` jsonb is self-identifying). Attribution comes from the transaction-local `nexus.source_chunk_id` setting both commit handlers now set right after chunk creation; a writer that forgets attribution still gets versioned with NULL chunk — never a silent skip. The first future runtime relationship writer cannot destroy history on contact.

## Tests

Live save_02 in always-rolled-back transactions: checkpoint section coverage against `CHECKPOINT_SECTIONS` with SQL count parity + idempotency + label validation; the Skald ledger end-to-end through `apply_state_updates_sync` (three fields, entity ids resolved); trigger pre-image capture on UPDATE and DELETE with chunk attribution; NULL attribution for unattributed writes. 700 live / 967 offline pass (only the known pre-existing `test_pair_tag_writer` failure).

With this, the spec's step-7 sufficiency bar is met going forward: bestowals chunk-keyed (#425), clearances logged on every path (#425), Skald deltas ledgered, checkpoints bounding replay, relationships version-safe. Closes the implementation scope of #426.

### DIFF
--- a/migrations/065_reconstructability.sql
+++ b/migrations/065_reconstructability.sql
@@ -1,0 +1,133 @@
+-- 065_reconstructability.sql
+--
+-- The "sufficient-from-now-on" reconstruction bar from
+-- docs/orrery_audit_dashboard_notes.md (step 7b-7d), per the decisions on
+-- issue #426: log every Skald-side scalar write (7b), JSONB state
+-- checkpoints auto-taken in the commit path plus a retro-fitted genesis
+-- (7c), and trigger-based relationship versioning that cannot be forgotten
+-- by a future runtime writer (7d).
+
+-- 7b. Skald-side chunk-keyed delta log. Orrery's half is already
+--     event-sourced in orrery_resolutions.state_delta; this is the peer
+--     ledger for the on-screen commit path, which previously moved
+--     characters and set activity with no record of any kind.
+CREATE TABLE IF NOT EXISTS state_delta_log (
+    id              bigserial PRIMARY KEY,
+    source_chunk_id bigint NOT NULL REFERENCES narrative_chunks(id),
+    writer          text NOT NULL
+        CHECK (writer IN ('skald_state_update', 'wizard_seed')),
+    entity_id       bigint REFERENCES entities(id),
+    field           text NOT NULL,
+    old_value       jsonb,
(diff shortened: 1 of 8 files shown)


pythagorakase:
Both findings addressed: the sync Step-8.55 checkpoint now opens its own cursor (the P1 closed-cursor crash would have hit every 25th accepted chunk — good catch), and migration 065 performs the genesis backfill itself with an idempotent NOT EXISTS guard, plus a tripwire test pinning the migration's jsonb_build_object sections against reconstruction.CHECKPOINT_SECTIONS so the two copies of the section list cannot drift silently.

PR #429, opened 2026-07-03
Multi-entity dynamism: joint beats + live event chains (migration 066)

pythagorakase:
## Summary

Dynamism rounds 2 and 3 — the multi-entity half of the emergence mandate. Two stacked commits:

### Joint beats (reciprocal/conflict post-pass — spec idea 6)

`compose_actor_target_bindings` already emits both orderings of every relationship edge, so A→B and B→A drafts coexist in one proposal. A pure post-pass now composes them: **`reciprocal`** (same template both directions — mutual reach, mutual surveillance: one scene) or **`crossed`** (different templates — one cultivates while the other hunts: tension to spring). Beats are advisory: rendered to the storyteller as a new `ORRERY JOINT BEATS` prompt section (cap shares `[orrery.prompt]`), carried on the proposal (serialization tolerant of pre-beat incubator payloads), and mirrored in the audit payload with an endpoint↔production parity test. The commit path is untouched — the two underlying proposals remain the adjudicable units.

### Live event chains (branch signals — spec idea 7, migration 066)

The four dead gate arms existed because nothing could legally emit them. Branches gain an optional **`signal_event_type`** — a second, *additive* world-event emission: the deed keeps its own `event_type` and every cooldown that consumes it; the signal is what the act gives off.

- `extract_vengeance`'s visible branches signal `threat_issued` → the target's `warn_ally`/`protect_kin`/`consult_rival`/`surveil` gates open next ticks.
- `surveil`'s detectable branches signal `compliance_alert` → the hunted feel the net (`evade_pursuers`/`warn_ally`).
- `cultivate_informant`'s distanced overture signals `encoded_message` → `honor_debt` hears it.
- `faction_realignment` stays dormant (no natural emitter) and is now the **only** exogenous-only arm — the catalog lint, dead-arm assertions, and coverage payload updated accordingly.

Migration 066 seeds `compliance_alert` + `encoded_message` into `event_types` (gate-consumed but never registered; `world_events` FKs the vocabulary). Applied to template + all slots.

## Tests

Pure: composition semantics (reciprocal/crossed, one-sided pairs yield nothing, deterministic ordering, serialization round trips including pre-beat payloads), resolution/draft signal passthrough, builtin wiring pinned with deed ≠ signal everywhere. Live (rollback transactions on save_02): a signalling draft commits **both** world-event rows (deed + signal with a `signal_of` payload marker), and the consumer-gate semantics are proven — the target hears `threat_issued` as next tick's actor while the avenger does not. Endpoint joint-beat parity on both audit slots. 717 live / 985 offline pass (only the known pre-existing `test_pair_tag_writer` failure).

Together with #427 (stochastic branches): variation within a character's day, pairs acting on each other, and one character's choice changing another's next choice. Closing follow-up: the before/after variety audit via the coverage analyzer.

### DIFF
--- a/migrations/066_signal_event_vocab.sql
+++ b/migrations/066_signal_event_vocab.sql
@@ -1,0 +1,24 @@
+-- 066_signal_event_vocab.sql
+--
+-- Register the two gate-consumed event types that no writer could legally
+-- emit (world_events FKs event_types.type): compliance_alert and
+-- encoded_message. With these seeded, branch signal emissions (spec idea 7:
+-- lighting up the dormant trigger events) can turn the dead gate arms into
+-- live package-to-package chains. threat_issued and faction_realignment
+-- already exist in the vocabulary.
+
+INSERT INTO event_types (type, category, severity, description)
+VALUES
+    (
+        'compliance_alert',
+        'intelligence',
+        'moderate',
+        'The subject has reason to believe surveillance or enforcement attention is closing in on them. Consumed by evasion/warning gates; emitted as a signal by detectable surveillance branches.'
+    ),
+    (
+        'encoded_message',
+        'interpersonal',
+        'minor',
+        'A covert communication reached the target through indirect channels. Consumed by obligation gates; emitted as a signal by distanced informant overtures.'
+    )
+ON CONFLICT (type) DO NOTHING;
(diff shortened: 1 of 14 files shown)


pythagorakase:
Finding addressed: the async signal INSERT now carries the same `::world_layer_type` / `::text[]` casts as its adjacent main-event statement, and — since the review correctly noted `commit_orrery_tick_async` had no coverage anywhere — the fix is pinned by a new live rollback-transaction test driving the real asyncpg commit path end to end (both deed and signal rows asserted).

Issue #442, opened 2026-07-09
Retrograde R3/R4/R5: Build the Procedural Core as Specified (Graph Builder + Injected Entropy + Split Selection)


pythagorakase:
## The Gap
User design-intent review (2026-07-09) found the creative core of Retrograde was flattened during implementation while the plumbing stayed faithful:

- **Spec** (`docs/orrery_retrograde_spec.md`): *R3. Graph builder — expand the sparse core into a candidate relationship/event graph with **open attachment points (dangling edges where history could connect)**. R4. Seed generation (high-entropy, high-friction) — over-generate candidate seeds **directionally connected to the graph but NOT implied by it. This is where surprise is injected.** R5. Semi-constrained selection.*
- **Implementation**: `retrograde_packet.py::_candidate_scaffolds` — "Return deterministic **R2/R3 scaffold cards**" — R3 collapsed into flat cards; no graph, no dangling edges. R4 became a single Skald call with prose-level weirdness exhortation (the model generates its own seeds). R5 selection runs inside the same call as generation.
- **Consequence**: no stochastic injection anywhere in the pipeline; the weird dial's genre bands resolve to rubric adjectives; slot-5 evidence shows all cold-start facts orbiting the premise's gravity well — coherent, zero orthogonal elements. The intended entropy amplifier is an LLM self-serve.

## Design Intent (user, verbatim spirit)
Retrograde is a cognitive offload and entropy amplifier: procedural generation that preserves enough structure and choice that Skald selects elements and must exercise divergent creativity to merge them — yielding a world with more and varied detail than the model generates de novo.

## Build Plan
1. **R3 graph builder** (`retrograde_graph.py`): nodes from the intentional core; dangling edges sampled from registered vocabulary (relationship types, pair tags with subject/object kinds, event types) with unknown endpoints. Seeded deterministic RNG (replay-identical). The weird band becomes the sampler's reach: orthogonality = probability of drawing vocabulary categories absent from the premise core vs adjacent to it.
2. **R4 against the graph**: each candidate seed must claim ≥1 dangling edge and resolve its unknown endpoint; schema + validation enforce edge accounting. Surprise from the roll; creativity spent on the merge.
3. **R5 split**: selection in a separate Skald call (menu decoupled from generation), keeping the existing combined response shape downstream so R6 expansion is untouched.
4. Serve both triggers: wizard cold start and runtime maturation (edge anchored on the matured entity).

[...]

PR #445, opened 2026-07-09
Retrograde Death-State Operation: Prose Deaths Become Engine Facts

pythagorakase:
## Why

Issue #442's thesis case: Retrograde asserted Kessa Brine's death in an event summary, but `entities.is_active` stayed true — so the Orrery committed five resolutions animating a dead woman (strolling, upkeep, recreation). Facts that live only in prose are invisible to the engine. Her slot-5 row is already data-fixed; this PR builds the mechanism so it cannot recur.

## What

**Expansion contract (`retrograde_expansion.py`)**
- New `RetrogradeExpansionDeathPlan {entity_ref, entity_kind, cause_event_ref, rationale}` on the R6 plan response; rides the wire as a `mechanical_plan` row with `plan='death'` (subject_ref/subject_kind/source_event_ref), so the provider schema shape is unchanged.
- Validation: cause_event_ref is required (a prose-only death is invalid), must name a planned event, and that event must list the dying entity as a participant — the death exists as a queryable event *fact*, not a flag. Duplicate deaths (case-normalized) rejected. Death refs count toward the entity-stub budget and stub staging.
- R6 prompt now states the contract and the hard rule.

**Persistence (`retrograde_persistence.py`)**
- `_plan_death_row`: dry-run reports `would_deactivate` / `blocked` / `already_inactive`; unresolved refs flow into the existing `unresolved_or_ambiguous_entity_refs` execute blocker. Execute flips `entities.is_active = false` — the complete Orrery kill switch (actor discovery/hydration filter on it at every site in resolver.py). Idempotent by re-checking `is_active` before writing; links `cause_world_event_id` once events insert. Deaths of stubs created by the same plan work (the dead historical figure who needs a row so relationships can reference her — the Kessa shape).
- Manifest gains `death_rows` + `deaths_*` counters.

**Maturation (`retrograde_maturation.py`)**: job-prefix event-ref namespacing now rewrites `death_plan[].cause_event_ref` alongside `source_event_ref`.

**CLI**: expansion summary prints `deaths:`; persistence summary prints the four `deaths_*` counters.

**Spec** (`docs/orrery_retrograde_spec.md`): death-state contract recorded under Integration/Reuse.

## Also

Fixes `test_retrograde_seed_candidates_reads_packet_and_writes_response`, broken on main since #443: it mocked generation with an empty candidate menu, which the real R5 selection stage now loudly rejects. It mocks `run_seed_stage` (the boundary the CLI actually calls) instead.

## Tests

8 new: expansion accept/reject×4 + wire expansion + prompt contract; persistence dry-run, execute (asserts the UPDATE fired via cursor double), idempotency, unresolved-blocker, stub-death staging; maturation namespacing. Full suite: 1029 passed, 2 pre-existing local failures unrelated (dashboard flag config test; both known).

Closes the death-state item of #442.

### DIFF
--- a/nexus/agents/orrery/retrograde_maturation.py
+++ b/nexus/agents/orrery/retrograde_maturation.py
@@ -900,6 +900,10 @@ def namespace_expansion_event_refs(
             source_ref = plan_row.get("source_event_ref")
             if source_ref is not None:
                 plan_row["source_event_ref"] = mapping.get(str(source_ref), source_ref)
+    for death_row in payload.get("death_plan", []):
+        cause_ref = death_row.get("cause_event_ref")
+        if cause_ref is not None:
+            death_row["cause_event_ref"] = mapping.get(str(cause_ref), cause_ref)
     return payload
 
 
--- a/docs/orrery_retrograde_spec.md
+++ b/docs/orrery_retrograde_spec.md
@@ -160,6 +160,7 @@ This is the most iteration-prone component. Budget for it.
 - **MEMNON:** retrieves generated history identically to play-generated. No special-casing; the storyteller cannot and need not distinguish setup-authored from play-accumulated history.
 - **Dehydrated-entity pattern:** generate for first-class starting entities and their pairwise relationships only. Implied others (spouses, children, vanished parties) stay tags/stubs until narrative attention promotes them. Retrograde must NOT recursively generate full histories for implied entities — that reopens the "it never ends" hole. **[OPEN]** exact promotion trigger reuse from the live dehydration mechanism.
 - **Reuse over rebuild:** stages R4–R6 are conceptually a short backward-Orrery pass with variance turned up. Prefer adapting existing resolver/event-writer machinery to building a parallel system.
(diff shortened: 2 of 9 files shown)


nexus/agents/orrery/retrograde_maturation.py
```diff
@@ -900,6 +900,10 @@ def namespace_expansion_event_refs(
             source_ref = plan_row.get("source_event_ref")
             if source_ref is not None:
                 plan_row["source_event_ref"] = mapping.get(str(source_ref), source_ref)
+    for death_row in payload.get("death_plan", []):
+        cause_ref = death_row.get("cause_event_ref")
+        if cause_ref is not None:
+            death_row["cause_event_ref"] = mapping.get(str(cause_ref), cause_ref)
```
pythagorakase, line 906: Addressed in the follow-up commit: deaths are now gated to backstory stubs staged by the same plan, at both the R6 validation layer (rejects first-class packet entities with a retryable message) and the persistence layer (`death_targets_preexisting_entity` execute blocker for anything that resolves to a pre-existing entity). `already_inactive` idempotency still short-circuits first. Covered by `test_persistence_death_of_preexisting_entity_blocks`, `test_persistence_execute_deactivates_staged_stub`, and `test_expansion_plan_rejects_death_of_first_class_entity`.

PR #491, opened 2026-07-14
Restore Divergence Detection: the Baseline Filter Was Subtracting Every Entity

### LINKED ISSUE #485: Pass-2 Karaoke Divergence Test Failing on Main: Warm Analysis Loses Its Characters
## Symptom

`tests/test_lore/test_pass2_chunk1369.py::test_pass2_handles_karaoke_divergence` fails on main (confirmed at `4e7dcd75` and earlier):

```
assert analysis.get("characters"), "Warm analysis should capture characters for notes"
E: {'chunk_count': 5, 'source': 'programmatic_warm_slice', 'target_chunk_id': None,
    'warm_chunk_ids': [1365, 1366, 1367, 1368, 1369]}.get('characters') → None
```

The `warm_analysis` phase state carries the programmatic warm-slice bookkeeping but no `characters` key at all — the test expects entity capture that isn't happening.

## Not Roster/Slot Fallout

Surfaced during the 2026-07-13 main repair (`10ad52a0`) but is a different class from the failures fixed there: it reproduces identically with `NEXUS_SLOT` unset and `NEXUS_SLOT=2`, so it is not a reset-dev-slot data precondition. Something in the warm-analysis/divergence path stopped populating `characters` (or the phase-state key moved).

## Next Step

Bisect when `test_pass2_handles_karaoke_divergence` last passed and diff the warm-analysis population path (`nexus/memory/manager.py` / `nexus/memory/divergence.py` / `entity_detector.py`) against what the test asserts. Sibling test file `tests/test_lore/test_memory_manager.py` still passes, so the break is likely specific to the pass-2/structured-stub flow.

pythagorakase:
## Summary

Closes #485 (Codex-first: Codex diagnosed from the issue's two-layer investigation and implemented; Claude reviewed the semantics and verified). Divergence detection has been structurally dead since 2026-05-18: commit `2ff10059` added a filter that discarded detector matches present in Pass 1's `baseline_entities` — but the hierarchical entity structure deliberately includes the **universal baseline (every entity in the game)**, so the filter removed *all* matches, always. `references_seen` was permanently empty; Pass-2 deep-cut retrieval never fired; the `requires_postgres` gate kept the karaoke integration spec out of CI.

The fix removes the filter and its helpers. The semantic judgment, made explicit: "entity present in the baseline" never implied "entity's history adequately covered" — asking about a deep-past incident involving on-screen characters (the karaoke case, chunks 743–770) is the feature's primary use, and the high-specificity detector + the Pass-2 token budget remain the retrieval bounds the retrieval bake-off validated. The manager unit test had asserted the filtered behavior (encoding the regression); it now guards the opposite. Layer 1 of the issue (the stale warm-analysis assertion) is fixed per the prescribed programmatic warm-slice contract.

## Root Cause Narrative

Inventory hydration was healthy (245 character/alias entries; the raw detector matched Emilia and Pete). The ancestry between the last-working commit and the break contains exactly `2ff10059`, so no bisect was needed. `test_memory_manager.py` missed it because it injected synthetic inventories *and* asserted the filtering.

## Verification (Claude, in the worktree)

- Karaoke repro green; `tests/test_lore/`: 106 passed offline, 118 passed live; full default suite 1170 passed.
- flake8 findings on touched files all pre-exist on HEAD (identical count); zero added lines over 88 chars.

### DIFF
--- a/nexus/memory/manager.py
+++ b/nexus/memory/manager.py
@@ -12,7 +12,7 @@ from sqlalchemy import text
 
 from .context_state import ContextPackage, ContextStateManager, PassTransition
 from .divergence import DivergenceDetector, DivergenceResult
-from .entity_detector import EntityMatch, HighSpecificityEntityDetector
+from .entity_detector import HighSpecificityEntityDetector
 from .incremental import IncrementalRetriever
 from .query_memory import QueryMemory
 
@@ -393,12 +393,10 @@ class ContextMemoryManager:
     def _detect_divergence(
         self,
         user_input: str,
-        context: Optional[ContextPackage],
     ) -> DivergenceResult:
         """Detect divergence using high-specificity entity matching."""
 
         entity_match = self.entity_detector.detect_entities(user_input)
-        entity_match = self._filter_baseline_entity_matches(entity_match, context)
         summary = self.entity_detector.to_divergence_format(entity_match)
         return DivergenceResult(
             bool(summary.get("detected")),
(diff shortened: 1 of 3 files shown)


pythagorakase:
Declining this one deliberately — it re-proposes the mechanism this PR removes, and the removal is the fix, not collateral:

1. **The subtraction can't mean what it says.** `baseline_entities` carries the hierarchical structure whose *universal* tier deliberately includes every entity in the game (that's the off-screen-tracking design in `query_entity_states`). Subtracting it is not "ignore entities already represented" — it is "ignore every known entity," which is why `references_seen` has been empty on every turn since 2026-05-18. There is no configuration of the removed filter that does what its docstring claims without first inventing a different input for it.

2. **The integration spec requires detection on baseline-present entities.** `test_pass2_chunk1369` (the karaoke case) asserts that "walk me back through the karaoke ambush" — naming two characters who are *in* the baseline and the warm slice — fires divergence and retrieves the deep-cut chunks (743–770). Deep history about the on-screen cast is the feature's primary use, not a false positive; a filter that suppresses it fails the spec by construction.

3. **The noise concern has its designed bounds elsewhere**: the high-specificity detector's matching constraints, the Pass-2 token budget, and incremental retrieval's dedup against already-present context. That is the pre-May shipped behavior the retrieval bake-off validated.

If live play shows Pass-2 metrics genuinely polluted by trivially-covered mentions, the right instrument is a *coverage-aware* signal (is this entity's relevant history in the warm slice?) — a new design with a new input, worth its own issue — not a roster subtraction. Happy to file that if the telemetry ever motivates it.

pythagorakase:
**Correction to this PR's blast-radius claim** (owner question prompted a closer look): Pass-2 retrieval was NOT dead these two months. `handle_user_input` runs the raw-input vector search unconditionally ("STEP 2: Always run raw vector search") — divergence gates nothing in the current pipeline; its result is recorded into context state and metrics only, and the entity-keyed `retrieve_gap_context` path has zero callers. What the May regression actually killed was the divergence *signal* (telemetry, `pass2.divergence_detected`, the karaoke spec's assertions), not retrieval. The fix stands as-is — the signal is what the spec pins and what this PR restores — but the commit message's "Pass-2 deep-cut retrieval was structurally dead" is wrong and the squash message at merge will say "the divergence signal was structurally dead" instead.

Follow-on design question (whether divergence should gate or supplement anything now that raw-chunk queries drive retrieval) is with the owner.

PR #505, opened 2026-07-15
Preserve operator-selected models during setup

### LINKED ISSUE #498: Wizard Flow Clobbers an Operator-Set Slot Model Back to the Roster Default
## Symptom

Setting a slot's model explicitly (`nexus model --slot 5 --set gpt-5.6`) does not survive the wizard/new-story flow: on every subsequent turn the server logs `Set model for slot 5 to gpt-5.5` / `Persisted model gpt-5.5 to slot 5` — the roster default — and generations run under that default, not the operator's choice.

Observed 2026-07-14 while standing up the Orrery-native test slot: the first ~36 chunks of save_05 were silently generated by `gpt-5.5` although the operator had set `gpt-5.6` before starting. Every surface reported success; only the server log and `global_variables.model` revealed the substitution. Silent wrong-model generation is the kind of failure that quietly invalidates fixture provenance and model comparisons.

## Workaround

Pass `--model <id>` explicitly on every `nexus continue` call — the per-request override is honored and persists (subsequent `global_variables.model` reads return the override).

## Expected

An explicitly-set slot model is an operator decision; the wizard/new-story path should treat it as authoritative and never re-persist the roster default over it. (Stamping the default onto a *virgin* slot is fine; overwriting a non-default value is the bug.)

## Notes

- Repro surface: fresh slot → `nexus model --slot N --set <non-default>` → run wizard → watch server log re-persist the default each turn.
- Related context: the slot-5 native playthrough (Orrery-native test-slot campaign) and the model-roster disambiguation (`gpt-5.6-sol` / `gpt-5.6-terra`).

pythagorakase:
## Summary

- Resolve setup models using explicit override first, then a non-placeholder slot stamp, and the wizard default only for a fresh slot.
- Preserve omitted model fields across HTTP, CLI, standalone wizard, storyteller, and UI surfaces until the core resolves them.
- Cover fresh-slot, operator-selected, explicit-override, and legacy setup paths.

## Impact

Starting or resuming the wizard no longer silently replaces an operator-selected model with the roster default, preserving fixture provenance and model-comparison integrity.

## Validation

- `PYTHONPATH="$PWD" poetry run pytest -q tests/test_api/test_wizard_model_resolution.py tests/test_cli_model_selection.py tests/test_new_story_cli.py` — 27 passed.
- Black and `git diff --check` passed.
- The broader `tests/test_cli.py` module retains an unrelated main-branch collection failure from its import of absent `GENERATION_POLL_SECONDS`; this branch does not touch that symbol.

Closes #498

Codex — GPT-5.6-Sol

### DIFF
--- a/nexus/api/new_story_flow.py
+++ b/nexus/api/new_story_flow.py
@@ -11,9 +11,8 @@ from typing import Any, Dict, Optional, TYPE_CHECKING
 import psycopg2
 
 from nexus.api.conversations import ConversationsClient
-from nexus.api.config_utils import get_new_story_model
 from nexus.api.new_story_cache import read_cache, write_cache, clear_cache, init_cache
-from nexus.api.save_slots import upsert_slot, clear_active
+from nexus.api.save_slots import clear_active, get_slot_model, upsert_slot
 from nexus.api.slot_utils import slot_dbname, all_slots
 from scripts.new_story_setup import create_slot_schema_only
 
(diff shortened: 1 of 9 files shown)


pythagorakase:
Addressed both Claude findings in 4064af7f.

- A prior assets.new_story_creator row now acts as the authoritative setup-started signal. That preserves a deliberately locked TEST model on an omitted-model restart while a genuinely fresh default-valued slot still selects wizard.default_model.
- resolve_setup_model now handles only the omitted-model path; the dead requested_model parameter and its test-only branch are removed. Explicit overrides remain in start_setup's direct authoritative branch.
- Added an end-to-end regression that starts setup explicitly on TEST, restarts with the model omitted, and verifies both conversations remain on TEST.

Validation: 28 focused tests passed across model resolution, setup HTTP surfaces, and CLI model forwarding. Black, scoped flake8, scoped mypy, py_compile, and git diff --check also passed.

Codex — GPT-5.6-Sol

PR #518, opened 2026-07-17
Record the Generating Model on Every Chunk (Migration 082)

pythagorakase:
Owner-requested (2026-07-17): chunk metadata now records which model generated it — the forensic reconstruction this week's bake-off required becomes a `GROUP BY`.

## What Lands

- **Migration 082**: `generation_model TEXT NULL` on `incubator` and `chunk_metadata`, with column comments. NULL means pre-082 history or non-storyteller writers (Retrograde, MEMNON import — intentionally unchanged).
- **Ground-truth stamping**: LOGON records `provider.model` off the *successful* call — after roster `@provider.role` resolution and per-turn overrides, last winning attempt under retries. Never re-read from slot config (the #498 wizard-clobber bug demonstrated config can diverge from what actually ran). Threaded through continuation, bootstrap, and regenerate paths into the incubator; both commit handlers copy it to `chunk_metadata` at accept. Missing provenance on a generated turn raises; the field is excluded from LLM-facing schemas.
- **`docs/skald_bakeoff_2026_07.md`**: the compiled July multi-Skald bake-off from all three surviving evidence tiers — including the previously unreported Phase D verdict (Hermes 4.3-36B: 1 of 5 submissions accepted, median 16.4 min/turn, failure mode stateful continuity rather than schema) and terra's combined 81-turn default-Skald confirmation.

## Verification

- Default suite 1393 passed; postgres suite 1606 passed (independently re-run). Mypy baseline improved 37→35.
- Rollback-only slot-5 live tests: accept-path copy, regenerate replacement, historical NULL tolerance.
- Migration on `NEXUS_template` only; slot rollout and the evidence-based slot-5 backfill (65 statements, chunks 1–148 minus the two genuinely unattested ids) follow merge.

Implemented by Codex (gpt-5.6-sol) from a frozen work order.

https://claude.ai/code/session_013UUKcWGZUXg2jF5Nqrhtqv

### DIFF
--- a/docs/skald_bakeoff_2026_07.md
+++ b/docs/skald_bakeoff_2026_07.md
@@ -1,0 +1,34 @@
+# Skald Bake-Off — Slot-5 Native Playthrough, July 2026
+
+Compiled 2026-07-17 from three evidence tiers: the orchestrating session's transcript (chunks 1–75, Phases A–C), `.nexus/runtime/slot5_sol_progress.log` (chunks 76–148, Phases D–E, machine-readable per-turn records), and the session summaries carried across the Fable→Sol→Fable handoffs. Slot 5 (`save_05`) is the first Orrery-native slot: Orrery and Retrograde ticked from the wizard phase onward, so every turn below exercised the full live pipeline. Per-chunk provenance is backfilled into `chunk_metadata.generation_model` (migration 082) for every chunk with direct evidence; chunks 77–78 remain NULL (absent from the surviving logs).
+
+## Verdict Table
+
+| Model | Phase / Chunks | Turns | Clean Rate | Latency | Verdict |
+|---|---|---|---|---|---|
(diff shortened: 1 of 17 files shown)


pythagorakase:
Addressed in bcbe7c9f — both redundant guards dropped; validation now lives only at the two real boundaries (the LOGON stamp on the successful call, and `write_to_incubator` at the DB write, which also covers bootstrap). Plain reads remain where the dicts are built. Suites re-run green.

Claude — Fable 5

pythagorakase:
Correction to my previous comment: bcbe7c9f was NOT green — one test (`test_response_to_incubator_rejects_missing_generation_model`) still pinned the removed adapter guard, and my verification pipeline masked pytest's exit status behind a `tail`, so I pushed and reported green on a red suite. Fixed in ce84819c: the test now asserts the adapter's actual contract (verbatim threading; the DB boundary owns rejection). Both suites genuinely green this time: 1393 default / 1606 postgres, exit codes checked.

Claude — Fable 5

PR #529, opened 2026-07-21
Add the COURT_PATRON Aspiration Package (#496 Slice 5)

pythagorakase:
## Summary

Fifth aspiration-band package per the #496 Arachne ruling. `court_patron`: an off-screen character works to win a named power's favor.

- **Stage ladder**: `gaining_notice` → `proving_worth` → `securing_favor`
- **Patron is a character in v1** — faction patrons stay a recorded future slice: `compose_actor_faction_bindings` only enumerates factions with a pre-existing institutional edge, so courting a new faction patron is structurally blocked today
- **Entry**: recruit_ally's eligibility base AND a power marker on the target (`has_any_status_at_or_above("senior")` OR the `leader` role tag OR existing `authority_over` the actor — status alone would never fire on live saves)
- **Completion** ("Secure the patron's favor"): patron's reverse trust ≥ 2 + the family hostility exclusion; three writes in one delta — **inbound `sponsors` (patron→actor) via the new `entity_pair_tags.add_inbound` applier verb**, outbound `obligation` (patronage has a price), and a `patron` relationship upsert mirroring the trait compiler's convention
- **`add_inbound`** is a thin subject/object-swapped delegation to the outbound applier (inherits idempotency + world-time stamping), loud on missing target binding, ledger-recorded with explicit orientation, replay-covered
- **SURVEIL** yields to due patron continuations (per the #527 lesson); the completion hostility exclusion was in the spec from the start (same lesson)
- **Migration 086**: five-type constraint widening, six event types, `DO NOTHING` idioms; zero new vocabulary

## Also in This PR — Audit Parity Fix (Fallout From #526)

The full live-PG suite (`NEXUS_RUN_POSTGRES=1`, which CI never runs) exposed that #526's `_arbitrate_project_starts` had no mirror in `explain_dry_run`: production kept one faction start while the explanation showed both stacks, failing the faction-binding parity test. Separate commit: the audit now mirrors the arbitration over solo+pair winner shims in production order, feeds only survivors to joint-beat detection, and reports drops as `project_start_arbitration_trimmed` (the `fanout_trimmed` doctrine). The parity contract is now *explained stacks − reported trims = production drafts*, encoded in the updated test.

## Validation

- `poetry run pytest -q` → **1471 passed, 303 skipped, 0 failed**
[...]

Issue #528, opened 2026-07-21
Orrery: Clear the Pre-Existing Mypy Debt (26 Errors in Six Files)


pythagorakase:
Surfaced during the #496 package campaign (PRs #526/#527): `poetry run mypy nexus/agents/orrery/` reports **26 pre-existing errors in six files**, none introduced by recent work — the old "3 known `PackageSelection | None` errors in substrate.py" baseline is stale (those were fixed or superseded during the #477 campaign; the current debt is different).

Breakdown:
- `retrograde_seed_candidates.py` — 18 errors: dynamically built Pydantic models used as annotations (`Variable "X" is not valid as a type`) and `default_factory=list[_T]` incompatibilities
- `tag_library.py`, `retrograde_maturation.py`, `worker.py` — 6 errors: missing `types-psycopg2` stubs (`import-untyped`)
- `status_family.py:136` — `str | None` assigned to `str`
- `declaration_validation.py:73` — `str` passed where `Literal['subject', 'object']` expected

Suggested slices: (1) add `types-psycopg2` to dev deps — clears 6 errors mechanically; (2) the two one-line typing fixes; (3) the seed-candidates dynamic-model annotations, which may warrant `# type: ignore[valid-type]` with comments or a small refactor to statically declared models.

Until cleared, every Orrery PR's mypy proof has to carry a "26 pre-existing errors are known baseline" disclaimer, which is exactly the kind of normalized warning that eventually hides a real regression.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #496, opened 2026-07-14
Orrery: Expand the Aspiration Band Beyond the Relocation Pilot


pythagorakase:
> Follow-up to #475, which closed when PR #494 shipped the `PLAN_RELOCATION` pilot. That issue proposed a six-package family; the pilot delivered one. This issue tracks the remaining four after PR #512 shipped RECRUIT_ALLY, plus the generalization questions still deferred.

## What the Pilot Proved

PR #494 built the family's shared chassis, deliberately not relocation-specific:

- **`character_project_states`** (migration 074) with a `project_type` CHECK vocabulary, per-project world-clock cadence (`next_eligible_at_world_time`), stage ladder, stall counter, and a partial unique index enforcing the one-open-project budget.
- **`project_due` predicate modes** (`ready`, `neglected`, etc.) — type-agnostic gating, so any project package can express "advance when due" and "suffer setbacks when neglected."
- **Promotable vs. skipped resolutions** (`Branch.promotable`) — routine progress ticks stay out of the promotion queue; only milestones can reach the Storyteller via bleed.
- **Replay-safe appliers** — project transitions persist their applied outcomes in the ledger, so reconstruction (#426 contract) works without recomputing from live config.

Live sampling confirmed the band coexists with routine/crisis behavior without distorting the routine-share distribution (max shift 0.2% across a 35-anchor comparison).

## Shipped Slices

- `PLAN_RELOCATION` — PR #494
- `RECRUIT_ALLY` — PR #512

## Remaining Candidates (From #475)

- `BUILD_VENTURE` — stand up a business/workshop/crew
- `COURT_PATRON` — win the favor of a named power
- `PURSUE_ROMANCE` — staged courtship of a named character
- `SEEK_REDEMPTION` — staged amends toward a wronged party

Names remain subject to the self-evident-naming doctrine; ordering is open.

## What Each New Type Needs

1. A migration extending the `project_type` CHECK plus the per-type stage and target constraints established by PR #512.
2. Start/advance/milestone/setback templates, following the relocation package's gate pattern (due-ness, transit/constraint guards, embodied-floor NOTs).
3. Catalog regeneration (`docs/orrery_packages.md` hook handles this).

## Open Design Questions

- **Polymorphic targets.** Resolved by PR #512: character projects use `target_character_entity_id` alongside `target_place_id`, with explicit FKs and per-type CHECK constraints.
[...]

Issue #476, opened 2026-07-13
Orrery: Relationship Drift — Off-Screen Outcomes That Move Trust and Relationship Types


pythagorakase:
> **Proposal set — Orrery/Retrograde lifelikeness (design review 2026-07-13).** One of seven sibling proposals; nothing here is committed, and feedback (including "don't build this") is the goal. Siblings: #474 (package selection), #475 (aspiration band), #476 (relationship drift), #477 (orbit distance / contagion), #478 (history shape), #479 (time-release backstory), #480 (mood / weather). Grounded in a full code survey of `nexus/agents/orrery/` post-Bundles 1–4b and post-#442.

## Context

Relationship trust is **read** by package gates (`trust_at_least` and kin, hydrated from `entity_relationships_v`) but never **written** by package outcomes: `SUPPORTED_STATE_DELTA_KEYS` (`nexus/agents/orrery/events.py:40`) covers activity, entity tags, pair tags, need fulfillment, travel, and world events — no trust delta, no relationship-type transition. Off-screen behavior can add a `hunting` edge or a contact tag, but the valence and type of standing relationships are frozen between player visits.

The lifelike cost is specific: returning to a location after 50 chunks, the player finds every friendship, rivalry, and marriage exactly as they left it. Much of what "the world moved on" *means* is that relationships drifted.

## Proposal

Two coupled pieces:

1. **Engine**: add `relationship.trust_delta` (bounded, e.g. ±0.05/tick) and possibly `relationship.type_transition` to the supported state-delta vocabulary, applied by the same committed-tick writer path as existing deltas.
2. **Content**: a small family of relationship-arc packages gated on trust levels, shared-history events, and cooldowns — e.g. `FALL_OUT`, `RECONCILE`, `GROW_APART`, `DEEPEN_BOND`. Existing packages could also gain drift as a side effect (`socialize` success nudges trust up; an undetected `surveil` of a friend nudges nothing — detection ecology already gives asymmetry).

## Constraints and Interactions

- **#426 (relationship versioning / reconstructability) is open and should be designed jointly** — drift makes point-in-time relationship reconstruction mandatory rather than nice-to-have. Landing drift before versioning risks unreconstructable history.
- **Authority model (#275 pattern)**: large moves (a marriage ending off-screen) arguably deserve Storyteller adjudication via the existing imminent-proposal surface, while small drift commits silently. Where's the line?
[...]

PR #530, opened 2026-07-21
Add the SEEK_REDEMPTION Aspiration Package (#496 Slice 6)

pythagorakase:
## Summary

Sixth and final aspiration-band package per the #496 Arachne ruling — the campaign's last slice. `seek_redemption`: an off-screen character stages amends toward a party they wronged.

- **Stage ladder**: `owning_the_wrong` → `making_amends` → `earning_forgiveness`
- **Entry evidence** keys on what live saves actually carry: the wronged party's directional trust at wary-or-worse, `enemy`/`rival` relationship types, or directed `hostile_to`/`hunting` edges (the semantically-obvious `grudge_active`/`vendetta_holder` tags have zero live rows anywhere); actors still actively hostile toward their target are excluded — but TARGET-side hostility deliberately does NOT block entry (amends toward someone who still hates you is the drama)
- **Completion** ("Have the amends accepted"): `project_due` + the family hostility exclusion, deliberately **no trust-recovery precondition** — nothing in the engine moves trust until the #476 drift tracker ships, so a warmth gate would block every redemption permanently; the staged amends are the repair, and active worsening (the spurned arm: resentful −2 or hostility) is what prevents it
- **The family's first valence-UPDATING upsert**: type → `complex`, valence → `+1|favorable` even on existing negative rows, with BOTH original type and original valence preserved as first-write provenance across re-completions (recruit/romance/patron deliberately never touch valence on conflict; this divergence is redemption's whole point)
- `grudge_active` cleared on the wronged party when present (absent-tag removal verified a no-op); entry −1 / spurned −2 threshold split leaves a deliberate high-risk band
- **Migration 087**: six-type constraint widening, six event types, `DO NOTHING` idioms; zero new vocabulary; SURVEIL yields to due redemption continuations

## Validation

- `poetry run pytest -q` → **1489 passed, 310 skipped, 0 failed**
- **Full live-PG orrery suite → 1043 passed, 0 failed** (includes 20 redemption tests: three-write completion, valence repair with originals surviving rewrite, replay drift, migration contract)
- flake8/black/catalog clean on changed files; mypy baseline exact (#528)

Also carries a one-file black reformat of the patron valence test from #529 (shipped flake8-clean but un-black-checked).

## Schema/Data Impact

Migration 087 pending everywhere; additive only. Applied at merge closeout. This PR completes the four-package build ordered by Arachne ruling seq 5.

Implemented by Sol (GPT-5.6 Codex) from a frozen spec carrying every prior slice's review lessons; reviewed line-by-line by Claude.

### DIFF
--- a/docs/orrery_packages.md
+++ b/docs/orrery_packages.md
@@ -925,6 +925,7 @@ Drive bands are authoring metadata: they explain whether a package is crisis/con
   - **NOT:** actor `recruit_ally` project passes `ready` due-state
   - **NOT:** actor `pursue_romance` project passes `ready` due-state
   - **NOT:** actor `court_patron` project passes `ready` due-state
+  - **NOT:** actor `seek_redemption` project passes `ready` due-state
   - **NOT:** actor is constrained or immobilized
   - **NOT:** actor has inbound `hunting` pair tag
   - **NOT:** actor has `grudge_active` ephemeral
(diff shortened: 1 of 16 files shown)


pythagorakase:

### Independent Sol P2 — Async Provenance Clobbered on Re-Run

**Confirmed and fixed, family-wide.** asyncpg returns JSONB as a string; the metadata helpers' `Mapping`-only check silently missed prior provenance on the async path, so a second async reconciliation rewrote the original `enemy`/`-4|hostile` record. The latent bug lived in **all four** family helpers (recruit_ally/pursue_romance/court_patron share the pattern — redemption's valence provenance just made it observable); all four now coerce through the existing `_decode_json_value` first. Pinned by the extended async test: seeded enemy row → complete → re-run the upsert directly → originals survive.

### Independent Sol P2 — CONSULT_RIVAL Fires on Reconciled `complex` Rows

**Accepted as intended behavior, recorded rather than changed.** A freshly-reconciled relationship *is* honestly `complex`, and CONSULT_RIVAL additionally requires a recent threat event to fire — post-amends wariness after a fresh threat is defensible off-screen drama, not corruption. The structural fix is #476's continuous-valence work making type-only dispatch valence-aware; retuning a sibling package's semantics inside this slice would be scope creep. Recorded on #496 as a #476 design input.

### Codex Bot P2 — Hostile-Evidence Arm Unreachable as a Composition Source

**Partially accepted; arm retained with the contour documented in code.** The observation is correct that pairs whose ONLY connection is a `hostile_to`/`hunting` edge never compose (`compose_actor_target_bindings` sources relationship rows — any type, bidirectionally — and `contact:social` edges). But the arm is not dead: it fires whenever the pair composed via either route and the hostile edge is the sole *evidence* (e.g., a hunting edge atop a neutral-valence relationship row). The dominant real-world evidence paths (negative valence, `enemy`, `rival`) ride relationship rows and are fully reachable. Widening the shared composer for the hostility-only corner would touch every two-party package for a data shape ("a wrong worth redeeming with no relationship row") that NEXUS's own model treats as impoverished — deliberately declined, documented in an inline comment on the arm.

### Sol's Clean Verifications (Concur)

Evidence/exclusion slot directions; completion-before-spurned with both-direction exclusion and the strict −1/−2 band; grudge removal orientation + absent no-op + replay ownership; migration 087 preserving all five prior types (faction-bound recruitment seeded); SURVEIL yield + arbitration auto-discovery.

### Validation After Fixes

- Default suite → **1489 passed, 0 failed**
- **Full live-PG orrery suite → 1043 passed, 0 failed** (20 redemption tests incl. the new async double-write)
- flake8/black/catalog clean

PR #533, opened 2026-07-21
Make Float Valence Canonical Under the Derived Eleven-Rung Ladder (#476 Stage 1)

pythagorakase:
## Summary

Stage 1 of #476 (relationship drift): the representation substrate, with **zero behavior change** as the explicit acceptance bar. The drift engine (producers, world-time-rated drain, rung-crossing claim minting) is Stage 2, a separate PR.

- **Canonical column**: `character_relationships.valence_current numeric NOT NULL`, `CHECK (> -1 AND < 1)` — the open interval is the owner's recorded Arachne position; endpoints are asymptotes reserved for drift
- **Mapping**: rung k ↔ center k/5.5; float→rung via `round(×5.5)` clamped. The original spec's k/5 mapping was mathematically inconsistent with the open interval (±5 → ±1.0, and live saves hold `+5|devoted` rows that would have failed backfill) — caught by the implementer *before* writing code and re-ruled to k/5.5, which round-trips exactly and makes the extreme rungs approachable-but-unattainable bands
- **Single boundary trigger** (fires before the migration-065 versioning trigger, alphabetically guaranteed): literal writes derive the float and canonicalize the literal; float writes project the literal; float wins when both change. Every legacy writer — trait compiler, all four package upserts, retrograde, the Skald enum path — keeps working unchanged
- **Ladder derived where it always was**: `entity_relationships_v.valence_magnitude` becomes `round(valence_current × 5.5)::int`, so all eleven trust-predicate sites, both `LOADED_TRUST` constants, and resolver hydration are byte-for-byte untouched
- **Unified derivation**: `communication.py`'s independent numeric-prefix regex (the second derivation mechanism) dies in favor of the float's sign — closing a live divergence where `+2|deferential` parsed in the regex but hit `ELSE NULL` in the view CASE
- **Canonicalized defaults**: the three non-canonical trait-compiler literals → `+2|friendly` / `+3|trusting` / `-1|wary`
- **Replay/checkpoint contract**: backfill runs through the boundary trigger so every migrated row gets a versioned pre-image with `valence_current: null`; `verify_checkpoints_sync` already compares absent keys as NULL (migration-070 precedent) — no verification weakening
- Live tests on unmigrated slots use a temp-table shadow that degrades to an inert pass-through once 088 is applied

## Validation

- `poetry run pytest -q` → **1508 passed, 320 skipped, 0 failed**
- **Full live-PG orrery suite → 1072 passed, 0 failed**
- `python scripts/replay_state.py --slot 5 --verify` → all checkpoints ok (pre-application; will be re-run post-application at merge closeout)
- Round-trip, backfill (incl. non-canonical literal), trigger-precedence, bounds, version-preimage, trust-parity, and communication-tier-parity tests all live-PG
- flake8/black/catalog clean; mypy baseline unchanged (#528)

## Schema/Data Impact

Migration 088 pending everywhere; additive column + trigger + view replacement + one-time backfill (no destructive changes). Applied at merge closeout with a post-application `--verify` as the final oracle.

Implemented by Sol (GPT-5.6 Codex) from a frozen spec — including one mid-order ruling exchange where Sol correctly halted on the spec's mapping inconsistency; reviewed line-by-line by Claude. Stage 2 follows.


pythagorakase:

### Independent Sol P1 + Claude Workflow Finding 1 (Converged) — Retired Literals Acquire Trust Magnitudes

**Rebutted with evidence; disposition: deliberate, documented correction with an empty blast radius.** Both channels correctly observe that a hypothetical `+2|deferential`/`+3|devoted`/`-1|beholden` row would go from NULL magnitude (no trust edge) to a real one. Two facts govern the disposition:

1. **The live census is empty.** All five saves were enumerated during design scouting: save_01/02 (84 rows each), save_05 (18 rows) carry exclusively canonical literals; save_03/04 are empty. Zero rows anywhere change behavior.
2. **The old NULL was itself the bug.** Pre-088, the same row produced *different* trust depending on which derivation read it — the view's exact-string CASE said NULL while `communication.py`'s numeric-prefix regex said `+2|trusting-tier`. Preserving NULL magnitude for these literals would preserve a two-mechanisms-disagree defect this stage exists to eliminate.

The "zero behavior change" bar therefore holds for every row that exists, and the class-level change is a correction, disclosed in the PR body. Declined to special-case NULL for retired literals.

### Codex Bot P2 — Same-Literal Writes Don't Re-Center

**Behavior is correct; the contract wording was the defect — fixed.** Wiping drift-accumulated intra-rung position when a writer merely re-asserts the current rung would be wrong (a package re-completion must not reset drift progress). The trigger and function comments now state the contract explicitly — same-projection writes are deliberate no-ops; re-entry at rung center happens only when the stored literal *changes* — pinned by a live test holding an off-center `0.4` through a same-literal rewrite.

### Codex Bot P2 — Responses Report the Input, Not the Stored Row

**Confirmed and fixed.** `_upsert_character_relationship` now `RETURNING`s the stored literal and float; both SQL-upsert call paths propagate storage truth into every response/reporting object, and the dry-run path mirrors the 088 projection so `+2|deferential` predicts `+2|friendly` exactly as execution produces. Live test added for typed non-canonical input.

### Validation After Fixes

- `poetry run pytest -q` → **1509 passed, 0 failed**
- **Full live-PG orrery suite → 1074 passed, 0 failed**
- flake8/black clean on all changed files

Sol's clean verifications (rounding semantics incl. SQL half-away-from-zero, versioning-trigger interplay with NULL chunk attribution, shadow inertness post-application, view compatibility, trigger edge cases, migration reentry-by-transaction) — concur.

PR #534, opened 2026-07-21
Add the Relationship Drift Engine (#476 Stage 2)

pythagorakase:
## Summary

Stage 2 of #476 — the engine that moves the Stage-1 float, completing the relationship-drift tracker's ruled design (producers, 3B world-time-rated accrual, 4C per-edge independence, 5C claim-minted visibility).

- **Drain** (`nexus/agents/orrery/drift.py`): Decimal-exact pure planner + sync/async appliers at accepted-chunk commit; hard-gated on `world_layer='primary'` + non-null `world_time` (two-clocks). Deterministic producer order: project milestones → hostile events → cooperative events → co-presence.
- **Producers**: two-party project milestones warm both directed edges (`project_milestone_delta`); hostile/cooperative events apply explicit `[orrery.drift.*_events]` config maps (sign-validated loudly by the Pydantic model); co-presence accrues `sign(v) × rate × capped_hours` — sign-following amplification that never creates affect from a zero edge, sampled at drain instants (limitation documented).
- **The asymptotic clamp**: `delta × (1 − |v|)` makes the owner-ruled open interval's endpoints unreachable by construction, with loud assertions.
- **Discipline**: drift never inserts rows — existing directed edges only, each independent (4C) — so every write is ledgered by migration-065 versioning with proper chunk attribution, and the Stage-1 boundary trigger re-projects the literal automatically.
- **Milestones (5C)**: a rung crossing emits one `relationship_drift_milestone` world event (migration 089 seeds the type; payload carries old/new rung+valence and the per-producer delta breakdown) and mints a claim through the existing epistemics rails via one `claim_event_types` addition — both endpoints aware.
- **Determinism**: Decimal end-to-end with `ROUND_HALF_UP` matching PostgreSQL's `round(numeric)`; identical inputs replay to identical floats, pinned by an exactness test.
- **Placement deviation, reasoned** (flagged loudly by the implementer): resolution-bearing ticks drain *after* the resolution loop so same-tick applied project milestones are visible to producer 1; resolution-free ticks drain after propagation. Deterministic per tick shape.

## Validation

- `poetry run pytest -q` → **1522 passed, 325 skipped, 0 failed**
- **Full live-PG orrery suite → 1090 passed, 0 failed** (engine unit + live commit-path coverage: producers, ordering, asymptote, sign-following, per-edge no-insert, elapsed cap, milestone event + claim + awareness, world-layer gate, disabled config)
- `replay_state.py --slot 5 --verify` green; `validate_config_commit.py` green; flake8/black/catalog clean
- En-route catch: a stale test fixture wrote bare `neutral` — migration 088's loud trigger rejected it exactly as designed; fixture canonicalized

## Schema/Data Impact

Migration 089 (event-type seed only, no schema changes) pending everywhere; applied at merge closeout — required before a real rung crossing can emit its event.

Implemented by Sol (GPT-5.6 Codex) from a frozen spec; reviewed by Claude. Completes the #476 build alongside Stage 1 (#533).

### DIFF
--- a/migrations/089_relationship_drift_milestone.sql
+++ b/migrations/089_relationship_drift_milestone.sql
@@ -1,18 +1,25 @@
 -- 089_relationship_drift_milestone.sql
 --
 -- Register the visibility event emitted when continuous relationship drift
--- crosses a derived Stage-1 valence rung. The runtime drain owns all movement;
--- this migration intentionally makes no schema changes.
+-- crosses a derived Stage-1 valence rung and the per-tick drain ledger marker.
+-- The runtime drain owns all movement; this migration makes no schema changes.
 
 INSERT INTO event_types (type, category, severity, description)
-VALUES (
-    'relationship_drift_milestone',
-    'emotional',
-    'minor',
-    'A continuous directed relationship valence crossed a derived rung.'
-)
+VALUES
+    (
+        'relationship_drift_milestone',
+        'emotional',
+        'minor',
+        'A continuous directed relationship valence crossed a derived rung.'
+    ),
+    (
+        'relationship_drift_drained',
+        'emotional',
+        'minor',
+        'The continuous relationship drift drain completed for a tick.'
+    )
 ON CONFLICT (type) DO NOTHING;
 
 COMMENT ON COLUMN event_types.type IS
-    'Stable event identifier, including system-emitted relationship drift '
-    'milestones registered by migration 089.';
+    'Stable event identifier; system-emitted event types are registered by '
+    'their introducing migrations alongside narrative vocabulary.';
--- a/migrations/089_relationship_drift_milestone.sql
+++ b/migrations/089_relationship_drift_milestone.sql
@@ -1,0 +1,18 @@
+-- 089_relationship_drift_milestone.sql
+--
+-- Register the visibility event emitted when continuous relationship drift
+-- crosses a derived Stage-1 valence rung. The runtime drain owns all movement;
+-- this migration intentionally makes no schema changes.
+
+INSERT INTO event_types (type, category, severity, description)
+VALUES (
+    'relationship_drift_milestone',
+    'emotional',
+    'minor',
+    'A continuous directed relationship valence crossed a derived rung.'
+)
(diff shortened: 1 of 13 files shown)


pythagorakase:

### Converged P1 (Codex Bot + Independent Sol) — Re-Commit Compounds Drift

**Confirmed and fixed.** Accumulation has no natural conflict key, so the drain now files a `relationship_drift_drained` marker world event per tick (even when zero edges move) and no-ops when the marker exists — drift's ledger analogue of the propagation drain's `ON CONFLICT` idempotency. Both paths, live-tested. (Version-row attribution was rejected as the signal: same-commit package upserts contaminate it.)

### Independent Sol P2s

- **28-digit Decimal context vs SQL numeric: fixed** — writes quantize to twelve decimals under a 40-digit planner context, reads defensively quantized, cross-language rung parity pinned by a live test seeding a 40-digit-scale value.
- **Epistemics not threaded on proposal-null paths: fixed** — both handlers now thread canonical epistemics settings; a resolution-free rung crossing mints its claim (live test).
- **Late FK abort on unmigrated DBs: fixed** — propagation-style loud gate at drain start names migration 089 immediately.
- **Silently-enabled drift on old configs: fixed** — model defaults `enabled=False`; the shipped block opts in explicitly.
- **Config escapes the soft clamp: fixed** — `< 1` magnitude validation for every discrete delta and the co-presence rate×cap product, loud at load.

### Claude Workflow Findings

- **Hostile/cooperative overlap double-applies: fixed** — `model_validator` requires disjoint maps (epistemics cross-check precedent), with a rejection test.
- **Self-referencing migration comment: fixed** — the `event_types.type` comment now describes the column generically.

### Validation After All Fixes

- `poetry run pytest -q` → **1531 passed, 0 failed**
- **Full live-PG orrery suite → 1102 passed, 0 failed**
- `replay_state.py --slot 5 --verify` green; `validate_config_commit.py` green; flake8/black clean

Sol's clean verifications (co-presence pair enumeration/directionality, elapsed-hours edge cases incl. first-chunk and negative-delta clamping, attribution ordering in both handlers) — concur.

Issue #479, opened 2026-07-13
Retrograde: Time-Release Backstory — Deferred-Seed Reveal Preconditions and Contradictory Accounts


pythagorakase:
> **Proposal set — Orrery/Retrograde lifelikeness (design review 2026-07-13).** One of seven sibling proposals; nothing here is committed, and feedback (including "don't build this") is the goal. Siblings: #474 (package selection), #475 (aspiration band), #476 (relationship drift), #477 (orbit distance / contagion), #478 (history shape), #479 (time-release backstory), #480 (mood / weather). Grounded in a full code survey of `nexus/agents/orrery/` post-Bundles 1–4b and post-#442.

## Context

Two [OPEN] decisions in `docs/orrery_retrograde_spec.md` concern how backstory *surfaces during play*, and both can be closed with the same move: make history time-release rather than fully-loaded at chunk one.

- Spec line ~142: deferred seeds — "**[OPEN]** ... whether deferred seeds carry explicit reveal-preconditions (tags/events that gate surfacing) vs. remain free-floating for Skald to surface opportunistically."
- The `hidden_truth` rubric function exists only as prose exhortation — nothing structural guarantees a backstory contains a discoverable secret with a fixed answer.

### 1. Deferred Seeds with Reveal Preconditions

Give deferred seeds explicit gate predicates drawn from the **existing package-gate vocabulary** (`recent_event`, `co_located`, tag presence, `since_last_event_at_least`) — no new predicate language. The forward tick evaluates deferred-seed gates exactly as it evaluates package gates; when one fires, the seed materializes as a `world_events` row (source `retrograde`, current tick) and enters the bleed menu.

Backstory becomes time-release: the debt surfaces when the creditor's faction gains status; the buried account surfaces when someone runs `uncover_past` near the right place. This is the cleanest available coupling between the two systems — **a deferred seed is a package whose actor is the past.** It also serves the seed-budget economics: deferred seeds cost nothing at bootstrap and pay out at the dramatically correct moment.

### 2. Contradictory Accounts as First-Class Seeds

[...]

PR #536, opened 2026-07-21
Add Durable Secrets and Template-Authored Reveal Gates (#479 Stage B)

pythagorakase:
## Summary

Stage B of #479: Regime B's durable secrets. A `backstory_secrets` row binds a **private** claim to a named reveal gate and a holder; the commit-time drain evaluates latent secrets and a firing gate composes three shipped subsystems — status flip → **incident-wide scope promotion** (the existing contagion drain takes over spreading) → co-located `told` grants → one `backstory_revealed` event.

- **Gates are template-source closures** (`reveal_gates.py`): `holder_death`, `participants_reunited`, `trust_earned` — reading the same hydrated state as every package predicate; an unregistered gate RAISEs at drain time; registry validated at import (the `validate_no_fame_in_entry_gates` precedent)
- **Idempotency verified, not assumed**: the `latent`-guarded `UPDATE ... RETURNING` row count gates every downstream write; live re-drain test proves 1-then-0 with exactly one event — no marker needed (the drift-marker pattern was the mandated fallback)
- **Drain state**: the resolver's WorldState doesn't survive to commit, so the drain hydrates its own minimal post-drift transactional snapshot (entity activity — now an explicit resolver-hydrated WorldState field — locations, trust with the resolver's tie-break, travel); an explicit `reveal_state` parameter accepts an exact snapshot when a caller has one
- **Authoring primitive** enforces private-while-latent and registered-gate-only, emitting a `backstory_secret_authored` ledger event
- **Migration 091**: the lifecycle table + two `revelation` event types; `[orrery.reveal]` opt-in (model default off, shipped config on — the drift precedent)
- Deliberate boundaries: no Retrograde generation-side authoring yet (recorded), no distortion (Stage C), no storyteller conduit (Stage D), no automatic retirement

## Validation

- `poetry run pytest -q` → **1538 passed, 0 failed**
- **Full live-PG orrery suite → 1125 passed, 0 failed** (gates fire/non-fire units; live authoring validation; end-to-end reveal with promotion, grants, single event; re-drain idempotency; world-layer gate; disabled config)
- `replay_state.py --slot 5 --verify` green; flake8/black/catalog/config-hook clean

## Schema/Data Impact

Migration 091 pending everywhere; new table + event-type seeds only. Applied at merge closeout.

Implemented by Sol (GPT-5.6 Codex) from a frozen spec; reviewed by Claude. Stage C (hop distortion) and Stage D (storyteller conduit) follow.

### DIFF
--- a/migrations/091_backstory_secrets.sql
+++ b/migrations/091_backstory_secrets.sql
@@ -1,0 +1,73 @@
+-- 091_backstory_secrets.sql
+--
+-- Durable template-authored reveal gates for time-release backstory claims.
+
+CREATE TABLE backstory_secrets (
+    id                      bigserial PRIMARY KEY,
+    claim_id                bigint NOT NULL UNIQUE REFERENCES claims(id),
+    gate_template_id        text NOT NULL,
+    status                  text NOT NULL DEFAULT 'latent'
+                                CHECK (
+                                    status IN ('latent', 'revealed', 'retired')
+                                ),
+    holder_entity_id        bigint NOT NULL REFERENCES entities(id),
+    source_chunk_id         bigint REFERENCES narrative_chunks(id),
+    revealed_at_world_time  timestamptz,
+    revealed_by_chunk_id    bigint REFERENCES narrative_chunks(id),
+    created_at              timestamptz NOT NULL DEFAULT now()
+);
+
+COMMENT ON TABLE backstory_secrets IS
+    'Durable time-release backstory claims. Lifecycle: latent -> gate fires at commit -> revealed; retired means withdrawn without reveal as a manual authoring correction and is never automatic.';
+COMMENT ON COLUMN backstory_secrets.id IS
+    'Monotonic identifier for one durable backstory secret.';
+COMMENT ON COLUMN backstory_secrets.claim_id IS
+    'Unique claim containing the secret account; authoring requires private scope while the secret is latent.';
+COMMENT ON COLUMN backstory_secrets.gate_template_id IS
+    'Name of a registered code-authored reveal gate, never a serialized predicate.';
+COMMENT ON COLUMN backstory_secrets.status IS
+    'Lifecycle state: latent, revealed, or manually retired without reveal.';
+COMMENT ON COLUMN backstory_secrets.holder_entity_id IS
+    'Entity whose secret this is and whose circumstances reveal gates normally read.';
+COMMENT ON COLUMN backstory_secrets.source_chunk_id IS
+    'Narrative chunk provenance for the authoring operation, when available.';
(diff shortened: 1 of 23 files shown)


pythagorakase:
## Review Responses (Fix Commit `2e9519eb`)

### Converged P1 (Codex Bot + Independent Sol) — Possession-Less Holder Poisons Commits

**Fixed at the root.** Authoring now grants the holder idempotent `claim_awareness` (a keeper knows their own secret — `granted` tier, authoring-chunk world time), in both sync and async paths. The reveal completes even for a holder authored with zero prior possession; live-tested. The Claude workflow's composing finding — the holder appearing in their own co-located revelation targets (a self-tell that previously masked this crash behind test fixtures) — is also fixed: `_co_located_participants` now excludes the holder, matching both gates' existing filter.

### Codex Bot P1 — Reveal Lifecycle Invisible to Replay

**Fixed.** `backstory_secrets` is now a checkpoint section, and replay reconstructs each secret's authoring and reveal state (status, `revealed_at_world_time`, `revealed_by_chunk_id`) from the `backstory_secret_authored`/`backstory_revealed` ledger events — events are truth, the projection is verified, per the claim-awareness pattern. Pre-091 checkpoints ride the absent-key precedent. Live replay coverage through an authored+revealed checkpoint pair.

### Independent Sol P1 — Incident-Wide Promotion Leaks Latent Sibling Secrets

**Confirmed and fixed with a named exemption.** `promote_claim_scope` now excludes claims bound to a latent secret (they stay private), and Stage A's sibling-scope agreement check permits exactly one divergence shape — private-with-latent-secret beside bounded siblings — while staying loud for everything else. Both sites cross-reference each other in comments. Live test: reveal A, sibling secret B stays private and unspread with clean hydration; reveal B, scopes converge.

### Converged P2 — Missing Activity Read as Death

**Fixed.** `holder_death` returns False for a holder absent from the activity map — irreversible reveals require positive evidence of death, per the `WorldState.is_active` contract's own missing-vs-inactive distinction. Unit-tested both ways.

### Validation After All Fixes

- `poetry run pytest -q` → **1538 passed, 0 failed**
- **Full live-PG orrery suite → 1130 passed, 0 failed**
- `replay_state.py --slot 5 --verify` green (now including the secrets checkpoint section); flake8/black clean
- En route: pre-091 hydration on unmigrated live DBs handled via the literal non-latent result (no graceful fallback — the table's absence is a schema-era fact, not an error)

Sol's clean verifications (single-transaction reveal atomicity, gate reachability under the 088 ladder, drain ordering visibility of same-tick drift, migration exactness, config threading) — concur.

Issue #532, opened 2026-07-21
Orrery: Faction Patrons and the Candidate-Composition Boundary


pythagorakase:
Deferred from #496 (COURT_PATRON shipped character-patrons-only in PR #529). Two related design questions about what the two-party composition layer can see:

1. **Faction patrons.** `compose_actor_faction_bindings` enumerates a faction only when the actor already has a `status:*`/`obligation`/`handles`/`authority_over` edge to it — so courting a NEW faction patron is structurally impossible (the chicken-and-egg recorded on #496). A faction-patron variant needs either an eligibility stance ("courting a power you're already on the periphery of") or a composition extension.

2. **Hostility as a composition source.** SEEK_REDEMPTION's review (PR #530) established that `compose_actor_target_bindings` sources candidate pairs from relationship rows (any type) and `contact:social` edges only — a pair connected solely by a `hostile_to`/`hunting` edge never composes. Declined as a package-slice change (blast radius across every two-party package for a data-shape-poor corner), but if composition widening is ever done for faction patrons, the hostile-edge question should ride the same decision.

Also recorded for the #476 drift design: CONSULT_RIVAL (and any type-dispatching gate) treats every `complex` relationship as distrust; SEEK_REDEMPTION now writes `complex` with favorable valence on reconciliation. Type-only dispatch should become valence-aware when #476's continuous valence lands.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

pythagorakase:
## Ruling Relay — Arachne Seq 6 (Filed 2026-07-22T18:54Z)

The owner has ruled on the composition boundary via the decision loom. As filed:

- **Boundary**: Discuss — *"I don't see why we can't implement both B and C."*
- **Option B sources**: hostile edges ✓, faction rosters ✓ — same-zone co-presence deliberately unchecked
- **Hostile-edge direction**: both orderings
- **Faction-patron shape**: widen `court_patron` (polymorphic targets within one package)

[...]

PR #543, opened 2026-07-22
Widen the Composition Boundary: Hostile Edges and Faction Rosters (#532 Stage 1)

pythagorakase:
Closes nothing yet — Stage 1 of #532 (Stage 2 lands the patron widening, status-bestowal verb, and introduction mechanisms).

## What This Ships (Arachne Ruling Seq 6)

Composition gains **both ruled shared sources**, each behind a `Template` opt-in flag and an `[orrery.composition]` config gate (model defaults off; shipped config on):

- **Hostile edges** — active character-only `hostile_to`/`hunting` pair tags compose ACTOR/TARGET pairs in **both orderings** (the ruling's directionality choice). Hostile-only pairs admit only templates setting `composes_from_hostility`; `START_SEEK_REDEMPTION` opts in, closing the #530 review's reachability contour (a wrong with no relationship row can now source its own amends project — live end-to-end test included). Faction-endpoint hostility deliberately composes nothing until a desire names the faction.
- **Faction rosters** — a faction with any live `status:*` roster becomes enumerable for actors within `roster_reach` relationship-graph hops of a member. The recursive CTE is membership-equivalent to `relative_orbit_distance <= reach` over the identical edge set (`entity_relationships_v`, character scope, active endpoints, undirected). Only `courts_factions` templates admit roster pairs — **nothing sets that flag until Stage 2's patron widening**, so this source is dormant scaffolding with live coverage.
- **CONSULT_RIVAL** drops its `has_relationship_of_type("complex")` arm: SEEK_REDEMPTION now writes favorable `complex` for reconciled pairs, so complex-alone is no longer a fraught tie; `trust_below(0)` keeps genuine distrust covered (#530 review provenance in the comment; catalog regenerated).

## Settings

`[orrery.composition]`: `hostile_source_enabled`, `roster_source_enabled` (both model-default **false**), `roster_reach = 2` (validated 1..4, loud ValueError at the resolver too). Threaded through resolver, audit (`explain_dry_run`), coverage, turn cycle, and dev endpoints in lockstep — disabled sources are byte-identical to legacy composition (tested).

## Evidence

- `NEXUS_RUN_POSTGRES=1 pytest tests/test_orrery/ tests/test_api/test_orrery_dev_endpoints.py -q` → **1246 passed, 5 skipped, 0 failed** (rerun independently after the build)
- Full suite: 1596 passed, 405 skipped
- `replay_state.py --slot 5 --verify` → all checkpoint intervals ok
- flake8/black/validate-config clean; pre-commit hooks passed on commit

No migrations. No schema changes.

### DIFF
--- a/docs/orrery_packages.md
+++ b/docs/orrery_packages.md
@@ -1831,7 +1831,6 @@ Drive bands are authoring metadata: they explain whether a package is crisis/con
   - **OR:**
     - actor has `rival` relationship to target
     - actor and target share `enemy` relationship (either direction)
-    - actor has `complex` relationship to target
     - trust actor→target < 0
   - **OR:**
     - recent `compliance_alert` event in last 10 ticks
@@ -3793,7 +3792,6 @@ the seeding migrations to confirm catalog ↔ schema agreement:
 - `captor`
 - `chosen_kin`
 - `companion`
-- `complex`
 - `comrade`
 - `enemy`
 - `family`
--- a/nexus.toml
+++ b/nexus.toml
@@ -253,6 +253,12 @@ enabled = true
 [orrery.binding]
 window_chunks = 30
 
+[orrery.composition]
+# Shared binding sources remain model-default-off for legacy configurations.
+hostile_source_enabled = true
+roster_source_enabled = true
+roster_reach = 2
+
 [orrery.contagion]
 enabled = true
 
--- a/nexus/agents/lore/utils/turn_cycle.py
+++ b/nexus/agents/lore/utils/turn_cycle.py
@@ -683,6 +683,7 @@ class TurnCycleManager:
                 contagion_settings=orrery_settings.get("contagion"),
                 weather_settings=orrery_settings.get("weather"),
                 mood_settings=orrery_settings.get("mood"),
+                composition_settings=orrery_settings.get("composition"),
             )
 
         turn_context.orrery_proposal = proposal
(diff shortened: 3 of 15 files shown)


pythagorakase:

Both channels converged on the same P2; fixed in 2b58f1a5.

**[P2] Roster CTE enumerates every simple path (Sol review + Codex bot inline, convergent)** — **Fixed.** The recursive CTE is gone entirely. Roster SQL is now only the cheap member→faction enumeration; reach is filtered in Python through the canonical `state.orbit_distance` mapping the resolver already hydrates every tick (BFS with a proper visited set, linear in edges). `include_rosters=True` without the mapping raises loudly. Cost is now |actors| × |roster rows| dictionary lookups.

**[P3] Duplicate SQL definition of the relationship graph could drift from `relative_orbit_distance`** — **Fixed structurally by the same change.** There is no second graph definition anymore: roster reach reads the *same mapping* the `relative_orbit_distance` evidence predicate reads, so equivalence holds by construction rather than by parity test. The code comment records both the canonicality and the review provenance.

**[P3] Reviewer sandbox could not reach localhost Postgres** — Acknowledged; the review sandbox (`workspace-write`) blocks localhost sockets. The live suites were run independently outside the sandbox after both the build and the fix: `NEXUS_RUN_POSTGRES=1 pytest tests/test_orrery/ tests/test_api/test_orrery_dev_endpoints.py` → **1247 passed, 5 skipped, 0 failed** at 2b58f1a5, including the new regression test proving a distance-3 roster member does not compose at reach 2 against a real hydrated state.

Issue #531, opened 2026-07-21
Orrery: Retrograde Spawn-Sourcing for the Aspiration Band


pythagorakase:
Deferred from #496 (decision recorded on that tracker, 2026-07-21): consume Retrograde's seed machinery as project generators so mature casts bootstrap mid-arc — the last unbuilt idea from #475's original proposal.

Ground truth from the #496 campaign's scouting: this is a **Large, cleanly separable** slice with zero existing seam — `RetrogradeSeedCandidate`/`RetrogradeMechanicalHints` emit tags/pair-tags/relationships/events only, and `retrograde_persistence.py` writes eleven tables, none of which is `character_project_states`. Building it requires:

1. A seed-DTO extension declaring project intent (coverage function → project_type + target binding).
2. A persistence step inserting `character_project_states` rows under the one-open-project budget index and all six per-type CHECK disciplines (migrations 074–087).
3. Wizard-time counterparty resolution for targeted types (and the faction-immutability trigger constrains anything faction-shaped).
4. Replay/reconstruction coverage for wizard-seeded rows, which bypass the resolution ledger's applied-snapshot path (#426 contract).

All six package types now exist (`plan_relocation`, `recruit_ally`, `build_venture`, `pursue_romance`, `court_patron`, `seek_redemption`), so the vocabulary surface is final.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

pythagorakase:
## Stage A Shipped — PR #544 Merged

Wizard-time project seeding is on main: R4 seeds may carry a typed `project_intent`, R6 weaves accepted intents into `project_plan`, and a dedicated persistence writer inserts stage-one `character_project_states` rows under the full validation discipline (CHECK-mirroring target shapes, redemption's TARGET→ACTOR wary-or-worse signal, deterministic ref-level arbitration before stub planning, cast-wide cap, per-type `<type>_started` events, dead-at-story-start participants rejected at both R6 and persistence).

**The load-bearing discovery:** the new-story flow captured no checkpoint at wizard completion — wizard-born rows had no honest replay base. Persistence now captures a `genesis` checkpoint at the synthetic prologue as its final in-transaction write, with a loud invariant against re-execution (a pre-existing genesis checkpoint raises before planning; `capture_state_checkpoint_sync`'s `ON CONFLICT DO NOTHING` `None` also raises). The live seam test proves a seeded project rides the checkpoint through verification with zero drift.

[...]

PR #544, opened 2026-07-22
Let Casts Start Mid-Arc: Wizard-Time Project Seeding (#531 Stage A)

pythagorakase:
Stage A of #531 (Stage B — maturation-time sourcing for declared entities — comes later and shares this stage's writer).

## What This Ships

Retrograde seeds become project generators so mature casts start mid-arc:

- **R4**: `RetrogradeSeedCandidate` gains optional `project_intent` (`project_type` from the six, compact `target_ref`, `rationale`), with a short prompt-policy section — intents are optional and rare, unresolved-ledger seeds may propose court_patron/seek_redemption, trait-bound-hook seeds the venture/romance/ally types.
- **R6**: `RetrogradeExpansionPlanResponse` gains `project_plan` (+ wire twin using the empty-string-for-None convention); the prompt instructs carrying a woven seed's intent forward or dropping it with the thread noted. Target-shape validation is **one shared validator** used by both the R4 intent model and the R6 plan model, mirroring the migration-087 CHECK arms exactly.
- **Persistence**: a dedicated writer step (the `coordinates` structural template) inserts `character_project_states` rows after entity stubs and relationships exist. All validation precedes any INSERT and raises with the seed id named: unresolvable refs, per-type target shapes, seek_redemption's TARGET→ACTOR wary-or-worse signal (checked against this plan's own relationship rows *and* the DB), and an existing-open-project guard so the one-open-project unique index can never fire. Stage-one starts only; deterministic first-wins per-character dedup with loud logs; cast-wide cap; per-type `<type>_started` events with `{seed_ids, project_intent, actor, target}` payloads — zero new event types.

## Genesis-Checkpoint Finding (the Load-Bearing Requirement)

The new-story flow captured **no checkpoint at wizard completion** — wizard-born projection rows had no honest replay base. `persist_retrograde_history` now captures a `genesis` checkpoint at the synthetic prologue as its **final in-transaction write** (whole-world rollback semantics preserved), and the manifest surfaces it. The live seam test seeds a project through the wizard path, applies a real synthetic transition, captures a later checkpoint, and verifies **zero drift**. `_replay_project` is unmodified.

## Config

`[orrery.retrograde.projects]`: `enabled` (model default **false**, shipped on), `max_seeded_projects = 3` (validated positive). Disabled means intents drop with a loud log — the feature is off, not the seeds invalid.

## Evidence

- `NEXUS_RUN_POSTGRES=1 pytest tests/test_orrery/ -q` → **1234 passed, 5 skipped, 0 failed** (rerun independently after the build)
- Full suite: 1602 passed, 406 skipped
- `replay_state.py --slot 5 --verify` → all checkpoint intervals ok
- flake8/black/validate-config clean; pre-commit hooks passed

No migrations. Known pre-existing mypy debt in `retrograde_seed_candidates.py` (18 errors around dynamic `create_model` types) is untouched — that's #528's territory.

### DIFF
--- a/nexus.toml
+++ b/nexus.toml
@@ -543,6 +543,13 @@ transition_timeout_seconds = 900
 # process-local registry. Run the API single-worker, or accept that status
 # polls answered by a non-writing worker report stage "idle" mid-transition.
 
+[orrery.retrograde.projects]
+# Wizard-time long-arc generation is opt-in in the model and enabled in the
+# shipped configuration. Every accepted intent starts on its first stage.
+enabled = true
+# Cast-wide bound after deterministic one-project-per-character arbitration.
+max_seeded_projects = 3
+
 [orrery.retrograde.graph]
 # R3 candidate-graph sampling calibration (issue #442). The graph rolls
 # dangling-edge ingredients the seed model must reconcile; these knobs set
--- a/nexus/agents/orrery/retrograde_expansion.py
+++ b/nexus/agents/orrery/retrograde_expansion.py
@@ -958,6 +958,11 @@ def _project_plan_issues(
     issues: list[str] = []
     threads_by_seed = {thread.seed_id: thread for thread in response.thread_plan}
     projects_by_seed = {project.seed_id: project for project in response.project_plan}
+    dead_character_refs = {
+        normalize_entity_ref(death.entity_ref)
+        for death in response.death_plan
+        if death.entity_kind == "character"
+    }
     for project in response.project_plan:
         candidate = candidates_by_id.get(project.seed_id)
         if candidate is None or project.seed_id not in response.selected_seed_ids:
@@ -986,6 +991,25 @@ def _project_plan_issues(
             issues.append(
                 f"project plan seed {project.seed_id!r} changes target_ref "
                 f"from {intent.target_ref!r} to {project.target_ref!r}"
+            )
+        if normalize_entity_ref(project.actor_ref) in dead_character_refs:
+            issues.append(
+                f"project plan seed {project.seed_id!r} actor_ref "
+                f"{project.actor_ref!r} appears in death_plan"
+            )
+        if (
+            project.project_type
+            in {
+                "recruit_ally",
+                "pursue_romance",
+                "court_patron",
+                "seek_redemption",
+            }
+            and planned_target in dead_character_refs
+        ):
+            issues.append(
+                f"project plan seed {project.seed_id!r} character target_ref "
+                f"{project.target_ref!r} appears in death_plan"
             )
         thread = threads_by_seed.get(project.seed_id)
         if thread is not None and thread.status != "woven":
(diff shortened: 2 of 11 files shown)


pythagorakase:
## Review Rounds 1–2 — Point-by-Point (Independent Sol Review + Codex Bot + Claude Workflow)

### Independent Sol Review

**[P1] Genesis doesn't snapshot `entities.is_active`** — **Rebutted with evidence, filed as #545.** `CHECKPOINT_SECTIONS` has never had an entities section; entity activity sits outside *every* checkpoint on every slot, so this is a pre-existing fleet-wide verification-scope boundary, not a Stage A regression — the genesis checkpoint inherits exactly the interval checkpoints' section list. Fixing it changes checkpoint shape globally (plus a legacy carve-out for stored snapshots), which deserves its own change: issue #545.

**[P1] Stale-genesis fallback reuses an old checkpoint** — **Fixed in 48288710.** Confirmed real and nasty: `capture_state_checkpoint_sync` is `ON CONFLICT (chunk_id, label) DO NOTHING`, so `None` means the pair already exists — the fallback would have blessed a checkpoint that cannot contain this pass's project rows. Persistence now raises on a pre-existing genesis checkpoint *before planning* (whole-world-rollback contract makes that a broken invariant) and raises on a `None` capture; the SELECT fallback is gone. Live test proves repeated execution raises.

**[P2] Project-only stubs survive arbitration** (also flagged by the Codex bot — convergent) — **Fixed in 48288710.** Deterministic ref-level arbitration (first-wins per normalized actor_ref, then cast-wide cap, R6 order) now runs before stub planning and is shared with row planning, so the two phases cannot disagree and dropped projects consume no stub budget. Distinct refs resolving to one entity id raise as a mis-shaped plan.

**[P2] Cap-dropped projects claim their actor key** — **Fixed in 48288710.** Only accepted projects register; `[A, B1, B2]` with cap 1 now reports `[inserted, dropped_cap, dropped_cap]` with truthful counters and no false `winning_seed_id`. (The Claude workflow read the old ordering as intentional two-phase semantics — the restructure makes the statuses truthful outright, so both readings are satisfied.)

**[P3] Reviewer sandbox blocked from localhost Postgres** — Acknowledged; live suites run independently outside the sandbox after every round: `NEXUS_RUN_POSTGRES=1 pytest tests/test_orrery/` → **1239 passed, 5 skipped, 0 failed** at 9779bf00.

### Codex Bot

**[P2] Stubs for arbitration-dropped projects** — Fixed as above (convergent with Sol's P2).

**[P2] Project actor in `death_plan` leaves an invisible open project** — **Fixed in 9779bf00.** Verified: the resolver hydrates projects only for `is_active` actors, so the row would be permanently orphaned. Both the R6 contract validator (generation-time bounce) and the persistence writer (defense in depth, redemption-precedent raise) now reject death-plan matches for project actors *and* for character targets of the four character-targeted types.

PR #546, opened 2026-07-22
Court the Power Itself: Polymorphic Patronage and Introductions (#532 Stage 2)

### LINKED ISSUE #532: Orrery: Faction Patrons and the Candidate-Composition Boundary

pythagorakase:
Closes #532 (Stage 1 = #543; this is the remainder of Arachne ruling seq 6).

## What This Ships

**Migration 096** — the project CHECK constraints re-created with all six arms carried verbatim; the court_patron arms become character-XOR-faction (target and completed-target both). The 081 immutability trigger is untouched.

**`status.bestow`** — the sanctioned state-delta verb for the exclusive status ladder: payload `{"level": ...}`, subject = ACTOR binding, scope = FACTION binding, routed through `apply_status_pair_tag_bestowal` (a new async twin joins the sync helper). All three raw-`status:*` guards **stay** and now name the sanctioned route in their messages. Replay-side, `status.bestow` joins `TAG_DELTA_KEYS` — its effects live in tag provenance tables like `mood.set`.

**Polymorphic COURT_PATRON** — `start_court_patron_faction` is the first and only consumer of Stage 1's dormant `courts_factions` roster source, gated on `NOT has_status_in_scope` (courting a power you already belong to is incoherent). The continuation/completion/abandon/milestone twins bind FACTION from the stored project; character-shaped and faction-shaped families never cross-evaluate. **Completion bestows `status:junior`** — the courted faction becomes a real membership edge the institutional composer already enumerates. That is the C-faction introduction: the composition circle closes by design, and the live test asserts the same actor's faction composes institutionally post-completion.

**MAKE_ACQUAINTANCE** — the C-character introduction: same-*place* off-screen stranger pairs (deliberately narrower than the ruled-out same-zone shared source; the code comment records the distinction) via a dedicated bounded composer behind `forms_acquaintances` + `[orrery.composition].acquaintance_source_enabled` (model default off, shipped on). Six deterministic guards (no relationship either direction, no contact edge either direction, no hostile edge either direction); the outcome writes **mutual** `contact:social` through the existing `contact_made` event — zero new event types — and is self-limiting because the written edge permanently excludes the pair. Next tick, the pair is visible to the romance/ally social-contact opt-ins: introduction feeds composition, proven live.

## Deviations (Recorded, Judged Sound)

- `audit.py` needed no edit: production and audit both delegate to `compose_actor_target_routes`, so the new source has a single routing authority (live parity test proves it).
- The work order's "mundane band" is not a literal `DriveBand`; `ANCHORED_ROUTINE` at priority 5 is the correct vocabulary.

## Evidence

- `NEXUS_RUN_POSTGRES=1 pytest tests/test_orrery/ -q` → **1265 passed, 5 skipped, 0 failed** (rerun independently after the build)
- Full suite: **1618 passed, 421 skipped** (also rerun independently)
- `replay_state.py --slot 5 --verify` → all intervals ok
- flake8 / black / validate-config clean; pre-commit hooks passed
- Migration 096 exercised only against throwaway schemas (`test_polymorphic_patron_migration_pg.py`); **not yet applied to any real slot** — fleet application follows merge per the standard runbook.


pythagorakase:

All four findings fixed in 26e0dd64.

**[P1] Faction starts cannot materialize a draft (Sol review, reproduced)** — **Fixed.** Court_patron `project.start` payload assembly is now polymorphic: a TARGET binding yields the character shape, a FACTION binding the faction shape, anything else raises. One materialization authority (`_materialize_project_delta`) is now shared by production dry-runs *and* audit/dashboard explanations. Root cause of the test gap: the live faction-patron test hand-built its draft object, bypassing exactly the materialization layer that was broken. It now drives the real `resolve_dry_run → draft → commit` path end-to-end, and the circle-closing assertion (post-completion institutional composition) runs against that path.

**[P1] Completion can silently demote standing (Codex bot)** — **Fixed.** `status.bestow` now defaults to **floor** semantics: the level applies only when the subject holds no status in that scope or one ranked strictly below (canonical `STATUS_LEVEL_RANKS`); otherwise it is a recorded no-op. An actor who earned `status:respected` mid-project keeps it at their patronage completion — proven live in both directions. Explicit `{"mode": "set"}` preserves replace semantics for future authored demotions; unknown modes raise. `apply_status_pair_tag_bestowal` itself is unchanged — its direct callers (trait compiler, retrograde) keep deliberate replace semantics.

**[P1] Acquaintance ordering defeats fanout caps / escapes the actor set (Sol review + Codex bot, convergent)** — **Fixed.** The composer now orients every pair so a hydrated actor occupies `Slot.ACTOR` (lower id wins when both are hydrated, for determinism) and emits **at most one pair per entity per tick** via a greedy pass over sorted candidates. The bot's ten-strangers-one-target shape is now the live regression test: one pair emitted, ACTOR inside the hydrated set.

**[P2] Stale column comments (Sol review)** — **Fixed.** Migration 096 now rewrites both target-column comments to state the XOR contract.

**[P3] Reviewer sandbox blocked from localhost Postgres** — Acknowledged; live suites run independently outside the sandbox after build and after fixes: **1269 passed, 5 skipped, 0 failed** at 26e0dd64.

PR #547, opened 2026-07-22
Let Matured Entities Arrive Mid-Arc (#531 Stage B)

### LINKED ISSUE #531: Orrery: Retrograde Spawn-Sourcing for the Aspiration Band

pythagorakase:
Closes #531 (Stage A = #544; this is the maturation-time half).

## What This Ships

When a declared entity matures mid-story (`drain_maturation_jobs_sync` → `_mature_one`), its scoped R4/R5 output may propose **at most one project whose actor is the maturation target** — enforced in code, not prompt: foreign-actor or surplus proposals drop with loud logs naming the job and seed. The insert is **Stage A's writer, refactored for a second caller** (identical validation semantics: per-type CHECK-mirroring target shapes, unresolvable refs raise, redemption's TARGET→ACTOR signal, existing-open-project guard, dead-participant rejection, stage-one starts, policy-derived `next_eligible`), with maturation provenance and the same `[orrery.retrograde.projects]` gating. Character-only targets, deliberately — migration 096 relaxing the DB does not change retrograde seeding discipline.

## The Replay Mechanism (the Load-Bearing Piece)

Maturation-born rows land *between* interval checkpoints, so neither Stage A's genesis-checkpoint answer nor the resolution ledger covers them. The ruled mechanism: the maturation insert emits the type's `<type>_started` world_event whose payload carries Stage A's provenance **plus a complete `applied` projection matching `_replay_project`'s exact contract**. Replay gains one bounded arm — maturation-provenance started events replay through the *same validation path* as resolution-sourced transitions (ladder membership, active status, quantization). Wizard-born `source='retrograde'` starts remain checkpoint-only with a regression test proving they cannot double-insert.

The live seam test: checkpoint → mature → seed → the real aspiration continuation template accepts the project when due → advance on a later tick → later checkpoint → reconstruct from the earlier one → **zero drift, no `uncertain_rows` exemption**.

## Deviation (Recorded, Judged Sound)

The `event_source_kind` enum has no `maturation` label and this PR ships no migration, so rows stay `source='retrograde'` discriminated by `payload.source='maturation'` + the `maturation_job_*` event-ref namespace — both writer-controlled and validated. Replay already accepts a future literal `source='maturation'` label if the enum ever grows one.

## Evidence

- `NEXUS_RUN_POSTGRES=1 pytest tests/test_orrery/ -q` → **1274 passed, 5 skipped, 0 failed** (rerun independently after the build)
- Full suite: **1621 passed, 427 skipped**
- `replay_state.py --slot 5 --verify` → all intervals ok
- flake8 / black / validate-config clean; pre-commit hooks passed

No migrations. No new settings. No template/catalog changes.

### DIFF
--- a/nexus/agents/orrery/replay.py
+++ b/nexus/agents/orrery/replay.py
@@ -113,6 +113,15 @@ MATURATION_PROJECT_STARTED_TYPES = {
     "seek_redemption_started": "seek_redemption",
 }
 
+# Step 8.55 captures the checkpoint before Step 8.6 persists maturation
+# starts in the same accepted-chunk transaction. A base-chunk maturation
+# start is therefore absent from the base checkpoint and must replay, while a
+# target-chunk start is absent from the target checkpoint and must not replay.
+# Ordinary Skald/resolution writes precede Step 8.55, so they use the opposite
+# edges: base-exclusive and target-inclusive.
+CHECKPOINTED_WRITE_WINDOW_SQL = "{column} > %(base)s AND {column} <= %(target)s"
+MATURATION_PROJECT_START_WINDOW_SQL = "{column} >= %(base)s AND {column} < %(target)s"
+
 # Composite natural keys, verbatim column order (faction_relationships
 # enforces faction1_id < faction2_id; character_relationships only c1 <> c2
 # — never re-canonicalize when re-inserting delete pre-images).
@@ -527,12 +536,18 @@ class _Replayer:
 
         if base_chunk < self.target_chunk_id:
             for chunk_id in self._window_chunks(base_chunk):
-                self._apply_skald_rows(chunk_id, characters, places, result)
-                self._apply_orrery_resolutions(
-                    chunk_id, characters, needs, travel, projects, result
-                )
+                if base_chunk < chunk_id <= self.target_chunk_id:
+                    self._apply_skald_rows(chunk_id, characters, places, result)
+                    self._apply_orrery_resolutions(
+                        chunk_id, characters, needs, travel, projects, result
+                    )
                 self._apply_maturation_project_starts(
-                    chunk_id, characters, projects, travel, result
+                    chunk_id,
+                    base_chunk,
+                    characters,
+                    projects,
+                    travel,
+                    result,
                 )
 
         tag_workings, tag_touched_entities = self._replay_tags(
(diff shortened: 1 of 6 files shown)


pythagorakase:

All three findings fixed in 30ee66c2.

**[P1] Checkpoint-boundary window inverted for maturation starts (Sol review + Codex bot, convergent)** — **Fixed.** Confirmed the transactional ordering: Step 8.55 captures the checkpoint before Step 8.6 persists maturation starts, so a checkpoint at chunk C excludes rows born at C. Maturation starts now use their own window — `>= base AND < target`, the exact inverse edges of ordinary checkpointed writes — expressed as two named SQL window constants under one comment explaining why, shared by the chunk-walk and the per-chunk selection so the asymmetry cannot silently regress. Live tests pin all three positions (base chunk, target chunk, mid-interval) at zero drift.

**[P1] Aspiration-band test hand-built its WorldState (Sol review)** — **Fixed.** The recurring #546 trap: the test now drives `resolve_dry_run` on real hydrated state and commits through `commit_orrery_tick_sync`; the entire Stage B live file was audited and no hand-built `WorldState`/`ProjectState` remains. This trap is now recorded as standing doctrine in the working-protocol skill so every future pipeline-feature review checks for it.

**[P2] Maturation started events break `plan_retrograde_summaries()` (Sol review)** — **Fixed, matching Stage A precedent.** Investigation confirmed wizard-born `<type>_started` events carry only mechanical payloads (no canonical summary/chronology), so project-start events are properly outside prose summary backfill; maturation-born starts now share the same explicit event-type exclusion, and a regression test proves the planner runs clean over a slot containing both kinds.

**[P3] Reviewer sandbox blocked from localhost Postgres** — Acknowledged; live suites run independently outside the sandbox after build and after fixes: **1277 passed, 5 skipped, 0 failed** at 30ee66c2.

PR #548, opened 2026-07-22
Fix replay metadata for window-born characters

### LINKED ISSUE #540: Need-Fulfillment Metadata Overwrite Drops the Applicability-Sync Marker (Replay Drift)
Found by tonight's live shakedown (slot 4, a clone of slot 5 grown 16 turns with the full epistemics stack enabled): `replay_state.py --slot 4 --verify` reports **4 drift findings**, all the same shape —

```
section: character_need_states, row_key: NN:thirst, column: metadata
expected: {"synced_by": "need_applicability", "last_fulfillment": {...}}
actual:   {"last_fulfillment": {...}}
```

**Diagnosis:** the need-fulfillment applier writes `metadata` by replacement, not merge. Rows previously stamped `synced_by: need_applicability` (by the applicability sync) lose that key the first time fulfillment fires after the stamp — an un-ledgered mutation, so reconstruction (which faithfully carries the stamp forward) diverges from the live row and `--verify` flags drift.

**Why it was invisible until now:** slot 5's checkpoints postdate its last growth, so no fulfillment ever fired *after* a checkpointed sync stamp. The bug needs post-checkpoint growth to manifest — the shakedown's first contribution.

**Scope:** pre-existing (nothing in the 084–093 campaign touches need metadata); replay-hygiene class, not story-state corruption. Fix shape: the fulfillment writer should merge (`metadata || new`) rather than replace — with symmetric replay-side expectations and a regression test that stamps, fulfills, and verifies.

Shakedown context for the record: the same run validated the full claims stack live (mint → reveal via `trust_earned` → incident promotion with the latent sibling correctly withheld → 14/14 idempotent drift drains → scheduled propagation released by a time skip with mid-skip world-time stamping → depth-correct canonical delivery → `world_knowledge` surfacing with correct acquisition shapes). This finding is the only red.

pythagorakase:
## Summary

- preserve the need-applicability marker when replaying fulfillment for characters born after the base checkpoint
- keep the legacy empty-metadata behavior for missing need rows on pre-existing characters
- add a rollback-only regression using the real character trigger, fulfillment writer, checkpoint capture, and verifier

## Root cause

Production already merges `character_need_states.metadata`; the stored and live slot-4 rows retain `synced_by: need_applicability`. The drift was replay-side: reconstruction seeds window-born characters without their unledgered trigger-created need rows, then replays fulfillment before end-of-window applicability synthesis. That created the replay row with empty metadata.

## Validation

- `NEXUS_RUN_POSTGRES=1 poetry run pytest tests/test_orrery/test_replay.py -q` — 20 passed
- `poetry run pytest tests/test_orrery -q` — 974 passed, 309 skipped
- `poetry run python scripts/replay_state.py --slot 4 --verify` — all checkpoint windows clean
- `poetry run python scripts/replay_state.py --slot 5 --verify` — all checkpoint windows clean
- Black, flake8, and targeted mypy pass

No schema, configuration, or live-data migration changes.

Closes #540

Codex — GPT-5.6-Sol

### DIFF
--- a/nexus/agents/orrery/replay.py
+++ b/nexus/agents/orrery/replay.py
@@ -533,6 +533,14 @@ class _Replayer:
             project.setdefault("target_faction_entity_id", None)
 
         born_entities = self._seed_window_births(characters, places, result)
+        character_entities = {
+            row["entity_id"]
+            for row in characters.values()
+            if row.get("entity_id") is not None
+        }
+        _, made_applicable_at = self._need_applicability_transitions(
+            character_entities, base_chunk, base_state
+        )
 
         if base_chunk < self.target_chunk_id:
             for chunk_id in self._window_chunks(base_chunk):
@@ -545,6 +553,7 @@ class _Replayer:
                         travel,
                         projects,
                         born_entities,
+                        made_applicable_at,
                         result,
                     )
                 self._apply_maturation_project_starts(
@@ -1102,6 +1111,7 @@ class _Replayer:
         travel: dict[int, dict[str, Any]],
         projects: dict[Any, dict[str, Any]],
         born_entities: set[int],
+        made_applicable_at: dict[tuple[int, str], int],
         result: ReplayResult,
     ) -> None:
         self.cur.execute(
@@ -1145,6 +1155,7 @@ class _Replayer:
                     world_time,
                     needs,
                     actor_entity_id in born_entities,
+                    made_applicable_at,
                     result,
                 )
             for project_key in (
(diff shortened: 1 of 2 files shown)


pythagorakase:
Addressed the tag-driven case in `70b2d58a`.

Replay now precomputes immunity transitions before resolution replay. A missing need row gets `synced_by: need_applicability` only when the character was born in the window or a recorded tag transition made that specific need applicable by the fulfillment chunk; arbitrary missing rows on pre-existing characters still receive empty metadata. For a tag initialization in an earlier chunk, pre-fulfillment debt is marked unreproducible because the trigger's `MAX(world_time)` firing-time anchor is not recoverable.

The new rollback-only regression covers base-checkpoint immunity → keyed clearance → later fulfillment → target checkpoint, asserts exact merged metadata, and verifies with zero drift. Both slot 4 and slot 5 remain clean, and the full PostgreSQL-enabled Orrery suite passes.

Codex — GPT-5.6-Sol

PR #551, opened 2026-07-23
Checkpoint entity activity

### LINKED ISSUE #545: Checkpoints Never Snapshot Entity Activity (entities.is_active)
Surfaced by the independent review of PR #544 (#531 Stage A), verified pre-existing and fleet-wide.

`CHECKPOINT_SECTIONS` (nexus/agents/orrery/reconstruction.py) captures every mutable projection the resolver hydrates or the Skald commit path writes — but has never included an `entities` section. `entities.is_active` therefore sits outside every checkpoint on every slot, and `replay_state.py --verify` cannot detect drift in entity activity: a deactivation (e.g. Retrograde death handling, or any future runtime kill switch) that fails to replay would still verify clean.

This is a **verification-scope boundary**, not a Stage A regression — the genesis checkpoint introduced by #544 inherits exactly the same section list as every interval checkpoint. Fixing it means adding an `entities` (or activity-projection) section plus the replay/verify comparison arm, which changes checkpoint shape globally and deserves its own migration-style care: existing stored checkpoints lack the section, so verification must treat its absence as legacy rather than drift.

Acceptance sketch:
- Add an activity section to `CHECKPOINT_SECTIONS` (id, is_active — decide whether kind/name belong).
- Teach verification to compare it, with a legacy-checkpoint carve-out for pre-section snapshots.
- A live test: deactivate an entity mid-window, verify catches a replay that misses the deactivation.

pythagorakase:
## Summary

- add an entities checkpoint section containing exactly id and is_active
- reconstruct in-window entity births from the schema default without copying later live activity
- persist each successful Retrograde death as an exact, idempotent applied_entity_activity projection on its mandatory cause event
- replay namespaced runtime-maturation deaths at the established base-inclusive/target-exclusive maturation boundary, validating the entity is an event participant
- keep raw unledgered deactivation visible as exact drift
- treat a missing entity section as an explicit legacy comparison boundary, including windows containing a maturation death
- update migration 065 genesis shape for fresh installs without rewriting historical checkpoint JSON

There is no numbered/data migration or historical backfill because current activity cannot truthfully reconstruct an old checkpoint.

## Review fix

Claude correctly identified that runtime maturation can create and deactivate a stub inside a replay window. Commit 730f5920 threads the effective recording boundary into the shared death writer, stores the exact false projection on the cause world event in the same transaction, and replays only writer-authored maturation projections. The JSONB append is idempotent and fails if the cause event disappears.

The broader inherited race where a detached maturation worker can finish after an intervening checkpoint is tracked separately in #552.

## Regression proof

- unledgered deactivation still reports entities / entity id / is_active with expected false and replayed true
- a ledgered runtime-maturation death persists one exact projection and verifies with zero drift
- replay rejects a projection whose entity is not a cause-event participant
- a legacy base carrying a later maturation death skips entity comparison without crashing or false drift
- existing window-born entity and need replay remains clean

## Verification

- NEXUS_RUN_POSTGRES=1 poetry run pytest tests/test_orrery/test_reconstruction.py tests/test_orrery/test_replay.py -q — 29 passed
- poetry run pytest tests/test_orrery/test_retrograde_persistence.py -q — 24 passed
- NEXUS_RUN_POSTGRES=1 poetry run pytest tests/test_orrery -q — 1,284 passed, 5 skipped
- poetry run mypy nexus/agents/orrery/ — zero errors in 50 files
- changed-file Black and flake8 — clean
- poetry run python scripts/replay_state.py --slot 4 --verify — six checkpoint windows clean
- poetry run python scripts/replay_state.py --slot 5 --verify — four checkpoint windows clean

Closes #545

Codex — GPT-5.6-Sol

### DIFF
--- a/migrations/065_reconstructability.sql
+++ b/migrations/065_reconstructability.sql
@@ -51,7 +51,7 @@ CREATE TABLE IF NOT EXISTS state_checkpoints (
 );
 
 COMMENT ON TABLE state_checkpoints IS
-    'Point-in-time JSONB snapshots of the mutable world-state surface (tags, pair tags, character scalars, relationships, needs, travel, routine anchors, faction memberships). Auto-taken every [orrery.reconstruction] checkpoint_interval_chunks accepted chunks; "genesis" marks the instrumentation-era boundary for slots older than migration 065.';
+    'Point-in-time JSONB snapshots of the mutable world-state surface (entity activity, tags, pair tags, character scalars, relationships, needs, travel, routine anchors, faction memberships). Auto-taken every [orrery.reconstruction] checkpoint_interval_chunks accepted chunks; "genesis" marks the instrumentation-era boundary for slots older than migration 065.';
 COMMENT ON COLUMN state_checkpoints.chunk_id IS
     'Head chunk at snapshot time; NULL only for empty slots checkpointed before any chunk exists.';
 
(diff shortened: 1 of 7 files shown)


pythagorakase:
Addressed in 730f5920. The shared death writer now appends an exact, idempotent applied_entity_activity projection to the mandatory cause event, including the effective maturation source chunk; replay applies only namespaced runtime projections at the same [base, target) boundary as maturation project starts and verifies the entity is an event participant.

Coverage now proves real PostgreSQL JSONB persistence, zero drift for the ledgered death, continued exact drift for a raw unledgered deactivation, and safe legacy behavior when the base checkpoint lacks entities. Full Orrery validation is 1,284 passed / 5 skipped, and slots 4 and 5 remain drift-free.

The inherited detached-worker/intervening-checkpoint timing race is captured separately as #552.

Codex — GPT-5.6-Sol

PR #560, opened 2026-07-23
Interim description diet on the full storyteller wire schema (#558)

pythagorakase:
Closes #558. Stage 2 (parallel track B) of the schema-weight campaign (Arachne seq 7).

## What Changed
- `diet_storyteller_wire_schema`: pure recursive transform removing `description` values > 70 chars from the serialized wire schema only — models keep their docstrings; non-description keys (even long ones) untouched; input never mutated.
- `storyteller_openai_text_format` reborn registry-independent: strict schema → diet → `text.format` payload, wired into the cached OpenAI branch of `_schema_format_kwargs`. Applies to every OpenAI-path consumer including local chat_completions endpoints. Anthropic branch untouched (SHA-256-verified byte-identical compact schema).

## Measurements (o200k, sorted minified JSON)
| Extended strict schema | Before | After |
|---|---:|---:|
| Tokens | 9,033 | **7,015** |
| Bytes | 40,155 | 30,195 |

Threshold data: 160 chars → only 1,428 tokens saved; 70 chars → 2,018, meeting the ruling's ≥2K acceptance. Tradeoff accepted with eyes open: mid-length semantic hints (70–160 chars) leave the wire; field names + system prompt carry the guidance. Interim by design — #554's wire model supersedes this whole path.

## Verification (run by orchestrator, not just claimed)
- 66 passed, 3 skipped across the validation suite + narrative-generation + logon files.
- New tests: nested traversal (incl. `prefixItems`), input non-mutation, long non-description keys preserved, ≥8,000-byte budget assertion, screed-absence (`"never a bare list of strings"` gone from wire), real `_schema_format_kwargs` wiring.
- black / mypy / flake8 clean on all three touched files.

Implemented by Codex (GPT-5) from a frozen work order; reviewed, gated, and verified by Claude.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01C7JLuuGzH37uY5YwY17hvJ

PR #561, opened 2026-07-23
Retire the inline dossier path from the canonical storyteller turn contract (#557)

### LINKED ISSUE #557: Retire the inline dossier path from the canonical storyteller turn contract
## Summary

Adopted by Arachne ruling (seq 7, 2026-07-23) as an amendment to the schema-weight campaign (#553/#554/#555).

`StorytellerResponseExtended.referenced_entities` still carries the inline new-entity dossier unions — `NewCharacter` (+14-field `CharacterTraits`), `NewPlace` (+16-field `PlaceDetails`), `NewFaction` (+`FactionDetails`) — wizard-era richness demanded synchronously, mid-prose, while the model is also writing the scene. Since #531, new persistent entities arrive through `new_entities` declarations plus asynchronous backstory maturation, which runs with full schema on its own dedicated calls. The compact Anthropic wire already operates without the inline path entirely, so the strongest storyteller path has been living without it in production.

These subtrees are the largest single block in the canonical wire schema (~2.5–3K of the 9K base tokens) and the biggest multiplication sites for the runtime tag-enum injection that #553 removes.

## Proposed Direction

1. **Census first.** Measure how often inline dossiers actually occur in recent live turns across slots (entity rows created by the commit path's inline handling vs. `new_entities` declarations). If nonzero, confirm the declaration + maturation channel covers those cases before removal.
2. Remove the `new_character` / `new_place` / `new_faction` unions (and their Details subtrees) from the canonical turn contract. Entity introduction flows exclusively through `new_entities` declarations.
3. Prune the corresponding inline-creation handling in `nexus/api/db_converters.py`; a legacy payload should fail loudly, not drop silently.
4. Scope is the live turn contract only. `scripts/map_builder.py` (wizard-era set design) keeps whatever models it consumes; this issue does not touch the wizard flow.

## Acceptance Criteria

- [ ] Field-usage census over recent chunks in live slots, reported in this issue before the removal lands.
- [ ] Canonical turn contract no longer accepts inline dossiers; declarations are the only new-entity channel on live turns.
- [ ] Commit path rejects inline-dossier payloads loudly (no silent drop).
- [ ] Wire-schema size regression test reflects the reduction.
- [ ] Replay verification and golden-path suites green.

## Sequencing

Independent of #553 (vocabulary) and compatible with #554/#555 (the sparse wire model simply never grows these fields). Can land before or alongside either.

— Fable (Claude), relaying Arachne seq 7

pythagorakase:
Closes #557. Stage 2 (parallel track A) of the schema-weight campaign (Arachne seq 7).

## Census (Step 1, acceptance criterion)

Ledger-based attribution of post-wizard entity creations (query logic on the issue thread):

| Slot | Total | Declaration/maturation | Retrograde mid-arc | Plausibly inline |
|---|---:|---:|---:|---:|
| save_04 | 39 | 9 | 26 | 4 |
| save_05 | 33 | 7 | 22 | 4 |

The inline path's live footprint is four place rows per slot (one water-infrastructure scene cluster; 3 of 4 timestamp-matched dossier creations). The declaration channel demonstrably carries the load. Reconciliation: the issue's 12/10 maturation-job counts include three jobs targeting pre-cohort entity IDs.

## What Changed

- `apex_schema.py`: `new_character`/`new_place`/`new_faction` unions removed from the Reference models (id-or-name now required); `NewCharacter`, `CharacterTraits`, `NewPlace`, `PlaceDetails`, `NewFaction`, `FactionDetails` deleted; dead `validate_entity_references` removed.
- Both inline commit paths pruned: async `db_converters` **and** the sync mirror in `commit_handler_sync.py` (caller analysis confirmed all six `create_new_*` helpers had no wizard/retrograde/maturation callers; `map_builder`'s same-named locals are unrelated and untouched).
- Prompt retirement: `storyteller_core.md` consolidated to the single declaration channel; `storyteller_bootstrap.md` stopped instructing fields the Bootstrap schema never had (pre-existing dead guidance).
- **Loud rejection proven:** legacy inline payloads fail the entire Extended→Standard→Minimal→Bootstrap fallback chain (test-enforced — no silent degradation).
- Regression: dossier-definition absence + 30 KB schema ceiling test.

## Measurements

Extended strict schema: **8,992 → 6,170 o200k tokens (−31%)**. Combined with #560's diet once both land, the full wire form drops well under half its pre-campaign weight. Touched-file mypy: 262 transitive errors at base → **11** at head (deletion removed debt surface; verified base-vs-head by orchestrator).

## Verification (run by orchestrator)

- Core set: 90 passed, 1 skipped. Wizard/new-story flows: 103 passed, 6 skipped. `-k "declaration or maturation"`: 39 passed, 23 skipped (live-gated).
- black/flake8 clean on all 12 touched files.
- DB-writing suites (replay, golden-path, new_story_setup) intentionally not run under the read-only build order; nothing in the diff touches their code paths.

Implemented by Codex (GPT-5) from a frozen work order; census, diff review, gates, and verification by Claude.


pythagorakase:
## Independent Review (fresh Sol, frozen 22dd767a) + Resolution

**VERDICT: CHANGES_REQUIRED** — 2 P1, 1 P2, 1 P3. All addressed.

**[P1] Same-turn reference+declaration loses the introduction junction row** (references resolved before stubs existed) → **Fixed**: declaration processing now precedes reference resolution in BOTH commit paths. The async handler had the deeper latent defect — it omitted `new_entities` processing entirely — and now creates stubs (geo-resolved places included) before resolution. Real-entry-point junction regressions added for both handlers. *Documented divergence:* the async path creates stubs without maturation enqueue — it has **zero production callers** (`narrative.py:928` sync is the sole live accept path, verified); porting the psycopg2 enqueue machinery to a dormant path was deliberately not done.

**[P1] API silently degraded legacy payloads to Minimal responses** → **Fixed**: `_coerce_story_response` no longer catches dict-validation failures; they propagate loudly per repo doctrine. `create_minimal_response` retained — it has three legitimate non-dict/string-fallback callers.

**[P2] Rejection test bypassed the coercion path** → rewritten to drive the API-level path with full Extended-shaped legacy payloads per entity kind; verified these fail against the pre-fix silent-Minimal behavior.

**[P3] Bootstrap prompt incoherence** → residual tag paragraph removed.

Also from review verification: census SQL now published on #557 (correcting my earlier comment that claimed it was in the PR); reviewer independently reproduced 9/26/4 and 7/22/4 and validated that ledger-flag attribution would *worsen* provenance (late unrelated retrograde events).

**Post-review integration:** rebased onto main with #560; the diet-budget test failed exactly as the parallel-track interaction predicted (dossier deletion removed the descriptions the ≥2K delta was calibrated against, and the screed-anchor itself was deleted text). Recalibrated: ≥6,000 bytes / ≥1,200 tokens relative guards + **≤21,500-byte absolute ceiling on the final dieted wire (measured 20,666 bytes / 4,716 tokens)**. Full combined suite green (99 passed), live populated-registry test green, gates clean. Merging.

Issue #558, opened 2026-07-23
Interim description diet on the full storyteller wire schema


pythagorakase:
## Summary

Adopted by Arachne ruling (seq 7, 2026-07-23). Interim relief while #554 is built: the full strict wire schema carries roughly 3.7K tokens of `description` strings, much of it restating system-prompt guidance verbatim (the OBJECT-not-bare-list warnings appear in both homes, the declare-sparingly guidance likewise). Guidance gets one home — the prompt.

## Proposed Direction

- Mechanical wire transform at the existing hook point (`storyteller_openai_text_format` / the `AnthropicJsonSchemaTransformer` family): strip or truncate long `description` values from the serialized wire schema, keeping a whitelist of short one-liners on genuinely ambiguous leaf fields.
- The Pydantic contract is untouched — model descriptions keep documenting the code; only the wire serialization slims.
- Explicitly a stopgap, superseded by #554's fresh wire model. Tens of lines, no behavior change beyond the wire bytes.

## Acceptance Criteria

- [ ] Wire schema drops ≥2K tokens on the OpenAI path, measured before/after on a live slot with the same tokenizer as the campaign's baseline numbers.
- [ ] No regression in structured-output retry rate over a recorded-context comparison.
- [ ] Every removed guidance string verified present in the system prompt (single home) or explicitly re-homed there.

— Fable (Claude), relaying Arachne seq 7

Issue #555, opened 2026-07-23
Restore full authorial delta fidelity in the compact Skald protocol


pythagorakase:
## Summary

The existing compact Anthropic schema is light partly because it removes authorial capabilities that still exist in the canonical NEXUS contract.

Examples on current `main`:

- The compact `kind` enum permits character, place, and faction updates, while the expansion code also contains a relationship branch that the grammar cannot emit.
- The compact update object permits only `kind`, `name`, `status`, `tag_add`, and `tag_clear`; the expander understands additional character, relationship, location, and faction fields, but `additionalProperties: false` prevents Skald from sending them.
- Compact `chunk_metadata` is forced to `{}`, so Skald cannot declare chronology, world layer, or deliberate scene weather.
- Compact adjudications omit the canonical `replacement_state_delta` shape.

The next compact contract should be both lighter than the full DB-shaped schema and more expressive than today's reduced subset.

## Proposed direction: a sparse semantic-delta protocol

Give Skald a small set of story-domain deltas rather than a mirror of persistence models. A representative shape:

```json
{
  "narrative": "...",
  "choices": ["...", "..."],
  "scene_delta": {"elapsed_minutes": 10},
  "updates": [
    {
      "kind": "character",
      "name": "Brena Tideloft",
      "activity": "following the wet bell-sound",
      "tags_add": ["perceptive"]
    },
    {
      "kind": "relationship",
      "name": "Brena Tideloft",
      "other_name": "Odile",
      "valence": "warm",
      "dynamic": "newly trusting"
    }
  ],
  "new_entities": [],
  "adjudications": []
}
```

Names or compact context references should be resolved to canonical IDs locally. The hydration layer from #554 then constructs the existing rich Pydantic response.

Prefer clear semantic field names and small discriminators over abbreviated keys, positional tuples, or a free-form metadata blob.

## Required expressiveness

The compact protocol should be able to represent, when changed:

- scene chronology, episode transition, world layer, and deliberate weather;
- character activity, location, emotional state, observations, and tag deltas;
- place conditions, notable changes, and tag deltas;
- faction actions, stance changes, and tag deltas;
- relationship endpoints, type, valence, dynamic, and recent events;
- persistent new-entity declarations and tag/pair-tag hints;
- Orrery defer/void/replace adjudications;
- replacement effects using the same semantic-delta vocabulary rather than a second deep schema.

[...]

PR #562, opened 2026-07-23
Introduce SkaldTurnWire, the universal storyteller wire model (#554)

### LINKED ISSUE #554: Introduce a dedicated sparse wire model for Skald turns
## Summary

The current Anthropic compact schema is a hand-built provider override, but the returned JSON is still parsed directly as `StorytellerResponseExtended`. That couples two different contracts and forces Skald to emit fields that the compact grammar can only represent as placeholders:

- `chunk_metadata: {}`
- `referenced_entities: {}`
- `operations: {}`
- a required debug `reasoning` string

Those fields add little byte weight individually, but they blur ownership: the provider wire shape says Skald supplied a full NEXUS response even when runtime defaults or other subsystems supplied the actual values.

## Proposed direction

Introduce a dedicated provider-facing model—provisionally `SkaldTurnWire`—and an explicit deterministic mapper into the canonical `StorytellerResponseExtended` model.

Suggested boundary:

1. Native structured output parses and validates `SkaldTurnWire`.
2. A pure hydration function supplies documented defaults and expands compact fields.
3. The hydrated `StorytellerResponseExtended` undergoes normal Pydantic and semantic validation.
4. Everything downstream of LOGON continues to consume the existing canonical response type.

The first slice should preserve today's behavior exactly while allowing placeholder-only properties to disappear from the wire. Richer scene and entity deltas belong to follow-up #555.

## Design rules

- The canonical NEXUS models remain the application and persistence contract.
- Every omitted wire field must be deterministically defaulted, derived from an authoritative source, owned by another subsystem, or introduced by a later explicit delta.
- Prefer one stable sparse schema with optional capabilities over a combinatorial family of per-turn schemas.
- Keep debug reasoning out of the production contract; expose it only in an explicit diagnostic mode.
- Fail loudly when hydration is ambiguous. Do not infer durable state silently from narrative prose.

## Acceptance criteria

- [ ] A dedicated typed Skald wire model exists for extended narrative turns.
- [ ] Provider-native schema construction uses that wire model rather than a schema that only resembles the canonical response type.
- [ ] A pure, unit-tested mapper hydrates the wire response into `StorytellerResponseExtended`.
- [ ] `chunk_metadata`, `referenced_entities`, `operations`, and `reasoning` are not required placeholder fields on normal compact turns.
- [ ] Existing downstream LOGON, incubator, and commit interfaces continue to receive the canonical response type unchanged.
- [ ] Canonical validation and the slot-bound semantic validator run after hydration and still participate in bounded repair.
- [ ] Golden fixtures prove that representative current compact responses hydrate to the same canonical payloads and commit effects as before.
- [ ] The wire schema has a deterministic size/complexity regression test.

## Dependency and scope

- Builds on #553, which removes live data catalogs from the grammar.
- This issue establishes the boundary and removes no-op scaffolding. It does not yet redesign the state-update vocabulary.
- Bootstrap may retain its existing small contract unless sharing the new boundary clearly simplifies the implementation.

— Codex (GPT-5)

pythagorakase:
Closes #554. Stage 3 of the schema-weight campaign (Arachne seq 7) — the architecture centerpiece. Slice-1 design frozen on the issue thread.

## What Changed

**One wire contract for every provider.** `SkaldTurnWire` (new `nexus/agents/logon/skald_wire.py`) is a thin envelope reusing the canonical component models — no parallel hierarchy. OpenAI gets strict serialization (proven production guarantee); Anthropic + local chat_completions get lenient (`model_json_schema`) where optional fields are simply omitted — a prose+choices-only turn emits nothing else (scaffold-tax proof in tests). Transport selection is automatic by provider wire type; no config knob. Hydration is a pure, explicit, per-field mapper into `StorytellerResponseExtended` with the v1 no-ambiguity contract documented and structurally enforced (a future ambiguous field requires a deliberate rule). No `reasoning` on the production contract per ruling.

**The dialect family dies as deletions:** `orrery_tag_schema.py` (compact Anthropic schema + #558's diet transform + keep-lists) deleted outright; flat-`updates` expander arm and `_compact_*` helpers removed from `apex_schema.py`; repo-wide grep confirms zero surviving references. Anthropic *regains* chronology + referenced-entities by converging on the universal wire — part of #555's restoration lands as a side effect.

**Size-conflict resolution (spec error, mine):** the frozen order demanded ≤21,000 strict bytes *and* untouched component descriptions — unreachable, since the old baseline was achieved by dieting exactly those. Resolution: component Field descriptions authored terse at the source (≤70 chars; production Skald hasn't seen the long ones since #560 stripped them from the wire) with the three orphaned-guidance survivors verbatim (`replacement_event_type`, `declared_entity_role`, `Coordinates` Earth-geography) — allowlist-asserted, not threshold-excepted. Class docstrings stay full for humans.

## Measurements

| Serialization | Bytes | o200k tokens |
|---|---:|---:|
| Strict (OpenAI) | 23,345 | 5,204 |
| Lenient (Anthropic/local) | 22,413 | 5,056 |

vs 20,666 B / 4,716 tok dieted baseline: the +488-token residual is class-docstring weight, deliberately kept; #555's semantic-delta redesign owns the next cut. Campaign arc: 13,559 tokens (with baked-in vocabulary) → 5,204, one contract instead of three schema systems.

**Deleted-test note for the reviewer:** the populated-registry live test died with the compact builder it guarded. `SkaldTurnWire`'s builders take no dbname and don't import the registry — registry dependence is structurally impossible rather than byte-compared. Challenge welcome.

- Combined suite (wire + validation + apex + commit handlers + narrative gen + logon files + mock integration): **132 passed, 4 skipped**.
- Wizard/new-story set: 99 passed, 6 skipped.
- Golden hydration fixtures (rich turn, sparse turn, former-compact-style turn) assert field-level equality with canonical expectations; transport repair tests drive `SkaldTurnWire` through all three provider entry points.
- black / mypy / flake8 clean on all touched files.

Implemented by Codex (GPT-5) from the frozen design; size-conflict resolution, diff review, and verification by Claude.


pythagorakase:
## Independent Review (fresh Sol, frozen 0eeb1515) + Resolution

**VERDICT: CHANGES_REQUIRED** — 2 P2, 0 P1. The architecture held; both findings fixed in the follow-up commit.

**[P2] `LocationStateUpdate` lacked `extra="forbid"`** — pre-existing, but the lenient wire made it load-bearing: a misspelled field on Anthropic/local would validate and silently discard a location change. → Fixed, plus the general guard: a recursive schema-walk regression asserting `additionalProperties: false` on every object node of the lenient wire, so no future component reopens the hole. Direct misspelling-rejection unit added.

**[P2] Dev mock server still emitted `reasoning`** — every non-bootstrap TEST turn would fail wire validation, concealed behind postgres-gated tests. → Mock now emits sparse `SkaldTurnWire` payloads (exercising the lenient omission semantics in the dev flow), with a non-gated validation regression so wire drift surfaces immediately.

Re-verified (orchestrator, pipefail): 126 passed, 6 skipped across the combined suite; black/mypy/flake8 clean. Merging.

Issue #553, opened 2026-07-23
Shrink Skald's native schema by moving live vocabulary catalogs to semantic validation


pythagorakase:
## Summary

The Anthropic/Skald extended-turn schema is already much smaller than the canonical `StorytellerResponseExtended` contract, but most of its remaining weight comes from serializing live registry data as JSON Schema enums.

Measured on current `main` against the live slot registries, using compact JSON and `cl100k_base` only as a stable tokenizer proxy:

| Schema | Bytes | Proxy tokens |
|---|---:|---:|
| Full Anthropic `StorytellerResponseExtended` | 39,534 | 8,509 |
| Current compact schema with live vocabulary | 12,074 | 3,028 |
| Same compact shape with registry values treated as strings | 2,695 | 544 |

The live catalog currently contributes 460 single-entity tags, 27 pair tags, and 109 event types. These are data, not structural grammar. Skald already receives the tag vocabulary in its system prompt, and generation-time registry validation already repairs invalid tag and declaration hints.

## Proposed direction

- Keep small semantic enums in the native schema: entity kind, adjudication action, world layer, relationship valence, and similar closed structural choices.
- Stop embedding the full live tag, pair-tag, and event-type catalogs as schema enums.
- Accept those catalog values as strings on the wire.
- Validate every such value against the slot registry before incubation and route failures through the bounded structured-output retry.
- Extend the existing generation-time validator to cover `replacement_event_type`; retain commit-time validation as defense in depth.
- Make the schema stable across registry growth so schema caching and provider behavior do not depend on catalog size.

## Why this preserves NEXUS fidelity

The canonical Pydantic models, closed registries, and commit-time constraints remain unchanged. This moves a large data catalog from grammar compilation to semantic boundary validation; it does not make unknown values legal.

## Acceptance criteria

- [ ] The compact Skald schema contains no large live-registry enum blocks.
- [ ] Small structural enums remain natively constrained.
- [ ] Invalid single tags, pair tags, declaration hints, and replacement event types fail before incubation and participate in the configured retry budget.
- [ ] Commit-time registry checks remain in place.
- [ ] A deterministic schema-size test or report keeps the live extended-turn schema at or below 3.5 KB / approximately 650 proxy tokens, excluding provider wrappers.
- [ ] Tests cover valid and invalid values for every moved catalog boundary.
- [ ] A live or recorded-context comparison reports first-pass validity, retry rate, and end-to-end latency before and after the change.

## Related work and non-goals

[...]

Issue #556, opened 2026-07-23
Contextualize Skald's tag vocabulary instead of dumping the full registry into every prompt


pythagorakase:
## Summary

After removing registry catalogs from the native schema, the next Skald-facing weight is the system-prompt tag library itself.

Measured on current `main` with live slot data and `cl100k_base` as a tokenizer proxy:

- storyteller core prompt: 17,861 characters / approximately 3,614 tokens;
- rendered tag library: approximately 37.2K characters / 8.2K tokens.

The complete library is valuable for discoverability and closed-vocabulary correctness, but sending every tag description on every turn gives Skald more catalog than most scenes can use.

## Goal

Reduce routine vocabulary load without narrowing NEXUS's registry or silently taking authorial choices away from Skald.

## Options to prototype

1. Always expose the category taxonomy, active entity tags, and the exact tags relevant to current Orrery proposals.
2. Add a deterministic contextual selector for likely tag candidates based on present entities, entity kinds, scene pressure, proposed deltas, and current tags.
3. Preserve a path to the complete registry: an on-demand lookup, cached appendix, explicit expansion request, or repair-time catalog slice.
4. When semantic validation rejects a value, return close registered alternatives from the correct entity kind/category rather than replaying the entire catalog.
5. Version the prompt vocabulary with a registry digest so caches and diagnostics can identify exactly which catalog Skald saw.

These are candidates for a bakeoff, not permission to delete the full vocabulary path.

## Guardrails

- Closed vocabulary remains hard: no auto-registration and no silent normalization to a different semantic tag.
- Contextual selection must not hide the only valid tag for an active mechanical proposal.
- New-entity declarations need enough cross-category vocabulary to avoid collapsing every new entity into generic hints.
- Prompt reduction and schema reduction are measured separately.
- Registry changes must invalidate or version any cached selection.

## Acceptance criteria

- [ ] A reproducible report records current prompt characters/tokens and tag-library composition.
- [ ] At least two contextual exposure strategies are tested against a representative replay corpus.
- [ ] The comparison reports prompt size, first-pass valid-tag rate, structured-output retries, missed valid tag opportunities, false tag applications, latency, and qualitative prose impact.
- [ ] The chosen default reduces routine tag-library prompt weight by at least 50% without a material regression in registry-valid authorial deltas.
- [ ] Skald retains a documented way to reach the complete registered vocabulary when the contextual slice is insufficient.
[...]

PR #563, opened 2026-07-23
Sparse semantic-delta protocol: SkaldTurnWire v2 (#555)

### LINKED ISSUE #555: Restore full authorial delta fidelity in the compact Skald protocol

pythagorakase:
Closes #555. Stage 4 of the schema-weight campaign — the expressiveness payoff. Rulings recorded on the issue: delta-presence (owner, in-session) and the faction-ontology no-migration ruling.

## The Protocol

- **`scene`** — flat delta: `elapsed_minutes` / `transition` / `world_layer` / `weather`, all optional, absent = defaults. The 99% case (primary layer, no transition, derived weather) pays zero tokens.
- **`presence`** — the ruled delta-presence model: character-only `enter`/`exit` with set-idempotent semantics (case-insensitive name keys; enter-of-present and exit-of-absent are tolerated no-ops), per-chunk `mentions` (all kinds) and `transit` (places), `scene_reset` on relocation carrying the new setting + complete roster. **Silence carries the roster and setting forward.** Factions appear only in `mentions` — the ontology ruling: faction junction rows have been untyped since migration 023; presence is a character concept, setting is a place concept, and no consumer has ever read a faction-presence distinction.
- **`updates`** — four-arm discriminated union (character/place/faction/relationship) of semantic deltas hydrating mechanically to canonical `StateUpdates`; tags as plain lists per #553.
- **Hydration stays pure:** `hydrate_skald_turn(wire, presence_baseline=...)` — the baseline (parent chunk's present characters + setting place) is an argument, queried by LOGON via a new read-only helper. Commit paths, junction tables, and replay are untouched; presence deltas are wire-only, persistence keeps full per-chunk fidelity.
- **Loud rules:** presence-without-baseline raises; ontology-invalid kinds raise; a stance without its target raises.
- **Documented deferral:** replacement-vocabulary unification (the issue's "replacement reuses the delta vocabulary" line) — name-keyed deltas cannot map onto the actor/target-keyed `OrreryReplacementStateDelta` without proposal-role context, which pure hydration does not possess. Needs a slice that threads proposal metadata; deliberately not partially implemented.

## Measurements

| Form | v2 | v1 | Campaign start |
|---|---:|---:|---:|
| Strict | **19,097 B / 4,324 tok** | 23,345 / 5,204 | 57,590 / 13,559 |
| Lenient | **18,460 B / 4,212 tok** | 22,413 / 5,056 | — |

The issue's ≤1,200-token vocabulary-free target is **missed by ~3,000 tokens, honestly**: the residual is itemized (update arms ~1,369; retained declarations ~764; retained canonical adjudications ~704; scene/presence ~582; operations ~303; syntax ~490). No design was contorted to chase the number — expressiveness and loud-hydration rules ranked above it per the build order; a future slice can slim declarations/adjudications if the bakeoff says it matters.

- Combined family: **159 passed, 7 skipped** (golden hydration fixtures incl. roster math, scene-reset, idempotency, loud-rule raises; chronology split math; transport repair reruns with v2 shapes; the #562 `additionalProperties` closure walk over the new models).
- **Live-gated presence-baseline tests: 6 passed against the real slot DB** (Sol's sandbox couldn't run these; run by orchestrator).
- Wizard/new-story family: 67 passed, 4 skipped. Gates clean on all 9 touched files.
- Sol's repo-wide sweep: 10 failures, all environmental (8 order-sensitive OpenMP aborts passing independently; 2 macOS Keychain denials inside the sandbox — pass outside it).

## For the Reviewer

One pre-flagged craft concern: the prompt's world-layer trim lost the `extradiegetic` = "user addressing you out-of-game" definition, which is not self-evident from the name. Assess whether the trimmed Time-and-chronology paragraph still teaches the layer semantics adequately.

Implemented by Codex (GPT-5) from the frozen + amended order; faction ruling, live verification, and review by Claude.


pythagorakase:
## Session Pack-Up: State for Resumption

**Independent review (fresh Sol, frozen 3a0ac14a) — VERDICT: CHANGES_REQUIRED, 2 P1 / 4 P2 / 1 P3, all accepted:**

- **[P1]** Name-addressed updates never resolve to canonical IDs — commit path skips id-less rows, silently losing durable changes. *(Pre-existing guard, made fatal by v2's name-first protocol.)*
- **[P1]** `lore_adapter.py` strips most hydrated richness before incubation — observations/tags/conditions/notable-changes/stances/relationships dropped; reads nonexistent `current_status` (place conditions were **always** dropped — pre-existing bug).
- **[P2]** `/api/story/turn` + `/regenerate` supply no parent id → no baseline → presence raises or roster empties.
- **[P2]** Bootstrap never seeds the setting place → first baseline has `setting=None` indefinitely.
- **[P2]** Silent roster precedence: `scene_reset` ignores nonempty enter/exit; enter∩exit resolves silently (exit wins).
- **[P2]** Update arms with only kind+name hydrate to no-op junk rows.
- **[P3]** Prompt lost the world-layer definitions (extradiegetic especially).

**Decided fix semantics** (frozen; full order below): commit-time name resolution in both paths after stub creation, unresolvable = loud transaction abort; adapter switches to full canonical `StateUpdates` serialization (the hand-picking was the bug); parent-id threading from the API routes, missing = loud; bootstrap writes the starting scenario's place as the first setting junction row; roster contradictions become ValidationErrors caught by bounded repair; per-arm substantive-field requirement; one-sentence world-layer definitions restored.

**Branch state:** `fe75b871` is a **WIP commit of Sol's interrupted fix pass — PARTIAL, UNVERIFIED, DO NOT MERGE**. Seven files touched mid-flight when the session ended.

**To resume:** fresh worktree on this branch → relaunch the fix order below against it (deterministic `codex exec` form) → orchestrator verification → point-by-point resolution → merge. Then the final stage: #556 + the two-pass bakeoff arm.

<details><summary>Complete fix order (verbatim)</summary>

```markdown
# Fix Order: PR #563 Review Findings (2 P1, 4 P2, 1 P3 — all accepted)

Worktree branch `codex/555-semantic-deltas` (HEAD 3a0ac14a). NO git
operations; NO poetry install; DB writes allowed ONLY inside tests that
already follow the repo's live-gate convention (none expected — prefer
fakes at the connection seam as existing commit-handler tests do). Tests
UNPIPED via `PYTHONPATH=$PWD poetry run pytest ... -q`.

## P1-1: Name-addressed updates silently lost at commit

Verbatim: "name-addressed updates are never resolved to canonical IDs —
character updates with only a name hydrate with character_id=None, and the
commit path skips characters without IDs, silently losing the durable
change."

**Fix (decided):** resolve names at COMMIT TIME in BOTH commit paths, after
declaration-stub creation (the #561 ordering guarantees declared entities
resolve) and before apply_state_updates: for every state-update row lacking
an id but carrying a name, resolve via the existing lookup helpers
(characters, places, factions; relationships resolve both endpoints).
Unresolvable name → LOUD error aborting the commit transaction (no silent
skip, no partial application). Remove or invert the silent
`if update.character_id:` guards so a name-only row either resolves or
fails. Same semantics sync and async. Tests per path: name-only character/
place/faction/relationship updates land in the DB writes (drive the real
commit entry points with fakes at the connection seam); unresolvable name
aborts loudly.

## P1-2: lore_adapter strips hydrated richness before incubation

Verbatim: "most hydrated v2 fields are stripped before incubation — ...
the adapter reads nonexistent `current_status`."

**Fix (decided):** the adapter must stop hand-picking fields. Serialize the
full canonical `StateUpdates` (`model_dump(mode="json", exclude_none=True)`
or equivalent) into the incubator's entity_updates payload — the commit
side already rehydrates via `StateUpdates(**...)`, so full serialization is
the contract. Verify every field in the PR's mapping table survives
adapter→incubator→commit round-trip with a test that drives the real
adapter and asserts the incubator payload contains observations, tags,
place conditions, notable changes, faction actions, stances, and a
relationship update. Delete the `current_status` misread explicitly.

## P2-1: No baseline without metadata.target_chunk_id

**Fix:** the non-bootstrap generation flow must ALWAYS have the parent
chunk id — thread it from the turn/regenerate API paths into the context
LOGON uses for the baseline query (the incubator write already knows
parent_chunk_id; use the same source). After threading: a non-bootstrap
call without a parent id is a LOUD error, not an empty baseline. Adjust
the API call sites and test both routes' context assembly.

## P2-2: Bootstrap never seeds the setting place

**Fix:** the bootstrap commit writes the starting scenario's place as the
first chunk's SETTING junction row (the bootstrap flow knows the starting
location — trace where and wire it through honestly; no fallback
inference). Test: after a bootstrap commit, the baseline query returns
that place as setting.

## P2-3: Silent roster-operation precedence

**Fix (decided semantics):** contradictions are LOUD —
`scene_reset` with nonempty `enter` or `exit` → ValidationError (mentions/
transit remain legal alongside a reset; they are per-chunk, not roster
ops). The same case-folded name in both `enter` and `exit` →
ValidationError. Both as model validators on PresenceDelta with tests;
update the prompt's presence bullet with one clause ("on a reset, list the
full roster instead of enter/exit").

## P2-4: Empty update arms hydrate to no-op junk

**Fix:** each update arm requires at least one substantive field beyond
kind/name/id (model validator per arm; ValidationError message names the
arm) so bounded repair catches it at generation. Tests per arm.

## P3: World-layer meanings lost from the prompt

**Fix:** restore one compact definitional clause to the Time-and-chronology
paragraph: flashback = a scene set in the past; atemporal = dreams or
time-abnormal realms; extradiegetic = the user addressing out-of-game.
Keep it to a single sentence.

## Acceptance

- All prior suites green plus the new tests; paste verbatim final summary
  lines for: tests/test_skald_wire.py, tests/test_orrery_tag_validation.py,
  tests/test_commit_handler.py, tests/test_commit_handler_sync.py,
  tests/test_api/test_lore_adapter.py, tests/test_api/test_narrative_generation.py.
- Gates clean on every touched file (black --check / mypy with PYTHONPATH /
  flake8); paste verbatim.
- Report: files changed, where the parent-id threading and bootstrap
  setting wiring landed (file:line), deviations.
```
</details>

pythagorakase:
## Review Resolution (all seven findings)

**[P1] Name-addressed updates lost at commit** → commit-time name resolution in BOTH paths, after declaration-stub creation (#561 ordering guarantees declared entities resolve); unresolvable name aborts the transaction loudly; the silent `if update.character_id:` skips are gone. Real-entry-point tests per path.

**[P1] Adapter stripped hydrated richness** → `lore_adapter` now serializes the full canonical `StateUpdates`; the hand-picked field list — including the `current_status` misread that had silently dropped place conditions since before this campaign — is deleted. Round-trip test asserts observations, tags, conditions, notable changes, stances, and relationship updates survive adapter→incubator→commit.

**[P2] No baseline without parent id** → parent id threaded from `/api/story/turn` + `/regenerate` through narrative_generation → LORE → turn_cycle → LOGON; on DB-backed utilities a missing parent id is a loud error. *Orchestrator scoping on top:* the invariant applies where a baseline is structurally possible — a dbname-less utility returns no baseline and hydration's presence-without-baseline guard remains the loud enforcement at the point of consequence.

**[P2] Bootstrap setting** → bootstrap commit writes the starting scenario's place as the first chunk's SETTING junction row (honest FK lookup, no inference); baseline-read-after-bootstrap tested.

**[P2] Silent roster precedence** → `scene_reset` with nonempty enter/exit and case-folded enter∩exit overlap are now ValidationErrors caught by bounded repair; prompt gains the one-clause rule ("on a reset, list the full roster").

**[P2] Empty update arms** → every arm requires a substantive field beyond kind/name/id; repair-loop-visible.

**[P3] World-layer definitions** → restored in one sentence.

Verification (orchestrator, pipefail): combined suite green; **postgres-gated run 53 passed** including live baseline tests and the re-scoped lazy-init test; wizard family 63 passed; black/mypy/flake8 clean on all 18 touched files. The interrupted-WIP reconciliation checklist is in the build report (P1-1 was complete, P1-2's serialization right but its fixture broken, remainder completed fresh). Merging.

PR #564, opened 2026-07-23
Scene-contextual tag vocabulary: names always, descriptions on demand (#556)

### LINKED ISSUE #556: Contextualize Skald's tag vocabulary instead of dumping the full registry into every prompt

pythagorakase:
Closes #556. Final building stage of the schema-weight campaign (the two-pass bakeoff + measurement stage remain, per the owner's descope).

## Design

**Nothing is ever hidden — only descriptions are contextual.** Tier 1 (always): category taxonomy with descriptions, the complete 460-name index grouped kind→category, pair-tag names, event types only when proposals pend. Tier 2 (contextual, full entries): tags active on scene-roster entities (the #563 presence baseline supplies the roster) and tags referenced by pending proposals — the issue's never-hide-the-only-valid-tag guardrail satisfied structurally rather than heuristically. Registry digest line for cache identity/diagnostics.

**Placement win:** the library leaves the system prompt for the per-turn context section, making the system prompt turn-stable — a provider-side prompt-caching improvement the issue didn't ask for. Bootstrap keeps the full library (no roster exists yet; the opening scene introduces the world).

**Config:** `[apex.tag_library]` — `contextual` (the measurement stage's A/B lever, documented as such) + `suggestion_limit`. **Repair UX:** invalid vocabulary values now get kind-correct `did you mean` near-misses in the bounded-repair message, never cross-kind.

## Measurements (save_05, live-gated test prints + asserts)

| Rendering | Bytes | o200k tokens |
|---|---:|---:|
| Full library | 37,359 | 8,276 |
| Contextual | 17,448 | **4,026 (−51.35%)** |

Live context: 5 entities bearing 22 active tags + 1 proposal tag; all 460 registered names verified present in the index.

- 168 passed, 6 skipped across the tag-library/prompt-formatting/validation/settings/wire/lazy-init/narrative/mock suites; live save_05 proof green.
- Completeness invariant (every registered name, exactly once, fake + live), Tier-2 composition, digest stability/invalidation, placement split, A/B switch, cross-kind suggestion exclusion — all tested.
- black/mypy/flake8 clean on **all nine** touched files (two mypy errors surfaced on the full set — a None-guard and an `append(...) or` lambda — fixed orchestrator-side).

## Descoped (recorded)

The two-strategy replay-corpus bakeoff with prose-quality metrics rides the campaign measurement stage; the `contextual` switch exists to run it.

Implemented by Codex (GPT-5) from the frozen order; verification and gate fixes by Claude.


pythagorakase:
## Independent Review (fresh Sol) + Resolution

**VERDICT: CHANGES_REQUIRED** — 1 P1, 2 P2, 1 P3, all accepted and fixed.

**[P1] ID-namespace mismatch** — roster rows carry kind-scoped `characters.id`/`places.id`; `entity_tags_current.entity_id` keys on canonical `entities.id`. On live save_05, Kosi (row 9, entity 18) got entity 9's *place* tags — the central Tier-2 feature silently wrong on real data, invisible to fake-registry tests. → **Fixed structurally:** `TagLibraryContext` now carries typed `EntityRowReference(kind, row_id)` and the lookup translates internally — no caller can pass the wrong namespace again. Live Kosi regression test asserts her actual tags render and the false-positive entity's do not, plus a skewed-namespace fake-registry unit.

**[P2] `mood.set` proposals dodged Tier 2** → covered; the full proposal-delta grammar audited: the five pair-tag/status carriers are enumerated and intentionally index-only (pair tags have no Tier-2 descriptions by design).

**[P2] Expiry-blind active-tag read** → replaced with the resolver's canonical `load_current_entity_tags` under `load_anchor_world_time` — the two-clocks predicate reused, not duplicated; expired-unswept rows no longer render.

**[P3] Duplicate hints stacked suggestions on the first issue line** → per-line bounded suggestions.

Re-verified (orchestrator, pipefail): combined suites green, **live-gated save_05 run 55 passed** (both tag-library proofs incl. Kosi + presence baseline), black/mypy/flake8 clean on all 10 touched files. Merging — this closes the campaign's building phase.

Issue #566, opened 2026-07-24
Measurement stage: first live findings on the optimized schema


pythagorakase:
## Context

First live trial of the campaign-optimized stack (SkaldTurnWire v2 + contextual vocabulary, PRs #559–#564) on both ruled providers, per the seq-7 no-stopgap ruling ("optimize first, then see how Anthropic and Hermes actually perform"). Slot 4 re-cloned from save_05 as the trial bed, API server on `0982702d`.

## Findings

**0. Invalid first pass (methodology, orchestrator error).** The initial three turns ran against a stale main checkout (`8dcb2b24`, pre-campaign) — the API server had been started from it. Symptoms retroactively explained: zero junction rows (old compact Anthropic schema's empty placeholder), 37.6K-token Hermes prompt (old full library + old payload). Those numbers are void. Lesson recorded: verify the serving tree's commit, not just process freshness.

**1. Anthropic native rejects `oneOf`** — Pydantic's spelling for discriminated unions (`SkaldTurnWire.updates`). Fixed: transform rewrites `oneOf`→`anyOf` + strips `discriminator` (app-side Pydantic enforces the discriminator post-parse). **PR #565, review gate pending.**

**2. Anthropic native grammar-size ceiling.** With the fix, the request advances to grammar compilation and fails: *"The compiled grammar is too large."* The ~22KB lenient universal wire exceeds Anthropic's ceiling. The universal wire cannot ride Anthropic's native grammar enforcement as-is. Options for the design round:
   - (a) **Prompted-schema mode** on the Anthropic path: no `output_config`; schema conveyed via prompt; app-side Pydantic + semantic validation + bounded repair as the sole gate (the architecture's validator stack already carries this; needs an explicit no-native-format switch and a bakeoff of first-pass validity).
   - (b) **Further wire slimming** — #563's residual-weight itemization: declarations ~764 tok, canonical adjudications ~704, operations ~303 are the big movable blocks.
   - (c) Anthropic strict-tool path (`anthropic_strict_tool`) — likely the same grammar compiler; cheap to probe.

[...]

PR #570, opened 2026-07-24
Add provider-profile payload budgets for storyteller turns (#566)

pythagorakase:
## Summary

Implements D2 of Arachne ruling seq 8 (relayed on #566): a per-provider-class payload profile so **local storyteller turns assemble a smaller context while frontier turns keep theirs**. Measured motivation from the trials: Hermes 4.3-36B at 6–8 min/attempt on a frontier-sized ~30K-token payload blows the turn ceiling once retries multiply — the schema stopped being the local bottleneck; the payload is.

## Design

- New optional `[lore.token_budget.provider_overrides]` table, keys restricted to the wire classes (`openai` / `anthropic` / `local`), values validated `≥1000` and **shrink-only** (an override above base is rejected loudly). Committed default: `local = 24_000`.
- `resolve_storyteller_context_window(settings, wire_class)` is the one resolution authority — missing config, wrong types, or an unknown class raise instead of defaulting. The old silent `75000` fallback in the memory manager is gone.
- LOGON grows `resolve_provider_wire_type()`, extracted from provider construction into a shared `_resolve_storyteller_route()` — one classification authority, and sizing never constructs a provider.
- The turn cycle resolves the class once per turn, configures the memory-manager budget (per-turn because slots can switch models under a live instance), and passes the effective window **explicitly** into `TokenBudgetManager.calculate_budget`, which now refuses to re-read the base itself. Percent allocations (`[lore.payload_percent_budget.*]`) apply to the effective number unchanged.
- UI slider and settings PATCH endpoints intentionally keep targeting the base value (audited in the consumer table on the build report; no UI changes).

## Test Results

```
197 passed, 13 skipped (postgres-gated) pre-rebase; 191 passed, 13 skipped post-rebase onto the wire-v3 main
mypy: Success · black --check: clean · flake8: clean · config loads: OK
```

Pre-commit's model-drift checker caught one casualty of Black's rewrap (the `gpt-5` family-detection literal separated from its `# pin:` comment) — fixed by keeping the pin on the literal's line.

## Sequencing

Independent of the in-flight Anthropic prompted-transport PR except for `nexus.toml`/settings adjacency; rebased onto the merged wire-v3 main (e5aaecb4).

### DIFF
--- a/nexus.toml
+++ b/nexus.toml
@@ -209,6 +209,11 @@ max_deep_queries = 5
 apex_context_window = 75_000  # Total tokens available for APEX (includes prompt + context)
 system_prompt_tokens = 5000  # Reserved tokens for system prompt
 
+[lore.token_budget.provider_overrides]
+# Effective apex_context_window per storyteller provider class. Absent key =
+# base value. Local models need a smaller assembled payload (see #566).
+local = 24_000
+
 [lore.payload_percent_budget.structured_summaries]
 # Percentage allocation for entity summaries, relationships, events
 min = 70  # Minimum % for structured data
--- a/nexus.toml
+++ b/nexus.toml
@@ -209,6 +209,9 @@ max_deep_queries = 5
 # Context window limits for APEX generation
 apex_context_window = 75_000  # Total tokens available for APEX (includes prompt + context)
 system_prompt_tokens = 5000  # Reserved tokens for system prompt
+# Reserve for LOGON's tag library and formatting after Phase 5; live chunk
+# 1369 measured +2,999 tokens. Keep headroom above that observed overhead.
+prompt_overhead_tokens = 4_000
 
 [lore.token_budget.provider_overrides]
 # Effective apex_context_window per storyteller provider class. Absent key =
--- a/nexus.toml
+++ b/nexus.toml
@@ -124,7 +124,8 @@ theme = "veil"
 [ui.lore_budget_slider]
 # Bounds, step, and preset ladder stops for the LORE budget slider in the
 # settings pane. Values are token counts for lore.token_budget.apex_context_window.
-min = 10_000
+# Keep this floor aligned with the largest committed provider override below.
+min = 24_000
 max = 200_000
 step = 1000
 stops = [75_000, 100_000, 150_000, 200_000]
--- a/nexus/agents/lore/logon_utility.py
+++ b/nexus/agents/lore/logon_utility.py
@@ -274,7 +274,9 @@ class LogonUtility:
         self.provider: Optional[OpenAIProvider | AnthropicProvider] = None
         self._system_prompt: Optional[str] = None
         self._provider_bootstrap_mode: Optional[bool] = None
-        self._provider_wire_type: Optional[str] = None
+        self._provider_wire_type: Optional[Literal["openai", "anthropic", "local"]] = (
+            None
+        )
         self._validation_dbname: Optional[str] = None
         self._schema_format_cache: Dict[type, Dict[str, Any]] = {}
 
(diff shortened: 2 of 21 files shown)


pythagorakase:
**Independent review gate: CHANGES_REQUIRED — all findings confirmed, fixes in flight.**

The reviewer probed live rather than reading prose, and every finding survived my own code inspection of the frozen tree:

- **P1 (window never binds):** confirmed, and partially *pre-existing* — Phase 5 has always packaged warm-slice/entity/retrieval content with token counts as labels; only the Phase-2 retrieval path enforced its budget. But the ruling's point is that local turns actually assemble smaller payloads, so this PR must close it. Decided semantics: `token_counts["total_available"]` becomes a hard ceiling — over it, trim warm-slice oldest-first (never below the most recent chunk: delta-presence baselines on the parent roster), then retrieved passages, then raise loudly. Frontier at 75K rarely binds, so "frontier keeps theirs" holds automatically.
- **P1 (negative allocations):** confirmed via the reviewer's observed `total_available=-15002` — the reasoning-reserve heuristic keyed on the *frontier* model name while the *local* window applied. Fix: the turn's single route resolution threads the actual generating model into `calculate_budget`, and available context below 1000 raises with the arithmetic spelled out. The `gpt-4o`-pinned fixture that masked this gets fixed too.
- **P1 (LOGON-disabled turns break):** confirmed — `ensure_logon()` returns early under `enable_logon=False`, so the new Phase-1 guard fires before the legitimate `logon_disabled` skip at turn_cycle:1081. Fix: explicit base-window configuration path for LOGON-disabled flows; the postgres-gated Pass-2 test will be run for real (`NEXUS_RUN_POSTGRES=1`) as part of acceptance.
- **P1 (route TOCTOU):** confirmed — provider init re-reads mutable slot config after Phase 1 budgeting. Fix: the turn's resolution is authoritative; provider initialization re-resolves and aborts loudly on any drift.
- **P3 (slider range vs shrink-only validator):** confirmed at P3 severity (pre-write validation + UI rollback prevent bricking). Fix: slider floor and preset stops raised to the largest committed override (24_000), with a comment tying them together.

Merge waits on the fix round plus a fresh review pass.

pythagorakase:
**Round-2 review: CHANGES_REQUIRED again — and rightly. Both new P1s only exist because round 1's enforcement now exists; this is the gate converging on deeper strata, not churn.**

- **P1 (trimmed-vs-sent divergence):** accepted. Chunks registered with memory before Phase-5 trimming stay "known" after being dropped, so Pass-2 suppresses ever re-retrieving content the storyteller never saw. Fix invariant: a dropped chunk must be indistinguishable, to memory and retrieval, from one never retrieved — unregistered with balanced budget accounting, tested through the real memory manager.
- **P1 (sync path lacks route-drift guard):** accepted — the round-1 fix wired the expected-route comparison into async only. Sync gets the same guard.
- **P2 (ceiling blind to LOGON's prompt overhead):** accepted, with the reviewer's live measurement (+2,999 tokens on chunk 1369) driving the fix: a configurable `prompt_overhead_tokens` reserve (default 4_000) subtracted at enforcement, plus a last-line loud check at the LOGON call site that the FINAL prompt fits the effective window — the invariant the ruling actually cares about.
- **P3 (hand-pinned regression model):** accepted — the fix order said derive the committed default; the synthetic pin defeats the drift coverage. Now resolved through the real registry.

Reviewer re-ran the suites itself (207 passed focused; postgres-gated Pass-2 passed live). Round-3 fixes in flight; merge follows a clean pass.

pythagorakase:
Round-3 fixes verified and landed as 4f0a632a: trimming now unregisters dropped chunks and refunds their Pass-2 token charges (real retrieve→trim→refund→re-retrieve test), sync path gets the route-drift guard, `prompt_overhead_tokens` (default 4_000, live +2,999 measurement recorded) reserves LOGON headroom with a loud final-prompt-vs-window check, and the negative-allocation regression resolves the committed apex role through the real registry. My verification: 269 passed broad suite, targeted trim/refund tests green, postgres-gated Pass-2 test passed live again. Findings curve decayed 4 P1 → 2 P1 → converged; merging per house convention for review-fix commits.

PR #571, opened 2026-07-24
Switch the Anthropic storyteller to prompted-schema transport (#566)

pythagorakase:
## Summary

Un-breaks Anthropic storyteller turns. Twenty live compile probes (#566) established that Anthropic's native structured-output grammar compiler cannot hold the full universal wire — the budget is cumulative and structural, shared by output_config and strict tools, and survives no spelling including maximal nesting contortion, text stripping, or totalized serialization. Per the decision recorded on #566 (convergent independent consult under the new escalation ladder): **the storyteller default becomes prompted-schema**, with the existing app-side gate as the sole boundary. Bootstrap and wizard schemas compile fine and stay native.

## Changes

- **`AnthropicProvider.structured_transport`** (`"native"` default | `"prompted"`), validated at construction — mirrors `OpenAIProvider`'s existing transport seam. The prompted path: same bounded-repair loop; rejects caller-supplied `output_config`/`output_format`; tolerates exactly one outer markdown fence (both spellings); **carries reasoning effort as effort-only `output_config {"effort": …}` with no `format` key** — the build initially raised here citing docs, but a live probe showed the API accepts the effort-only spelling, so the capability is preserved (rebut-with-evidence; the storyteller call site doesn't pass effort today, making it an explicit measurement-stage lever).
- **`skald_wire_prompt_guide()`** — a deterministic compact guide rendered mechanically from the de-nulled lenient schema (BFS over reachable object defs; enums and field descriptions carried; loud errors on unexpected shapes). 133 lines, **1,391 o200k tokens**, asserted ≤ 1,400. Appended to the turn-stable system prompt in prompted mode only.
- **Config:** `anthropic_storyteller_transport = "prompted"` under `[API Settings].apex` (Literal-validated, loud on illegal values); bootstrap turns force native regardless.
- `_schema_format_kwargs` emits no `output_config` for prompted storyteller turns; wizard/non-wire schemas unchanged (regression-pinned).

## Test Results (unsandboxed)

```
331 passed, 43 skipped in 6.84s   (includes the keychain API test that fails only inside the build sandbox)
mypy: Success (3 files) · black --check: clean · flake8: clean · config loads: OK
```

## Sequencing

Built on the merged wire-v3 main (e5aaecb4). The measurement stage unblocks on this PR: real Anthropic prompted turns, b-prompt vs b-tool arms, contextual-vocab A/B, two-pass arm.

### DIFF
--- a/nexus.toml
+++ b/nexus.toml
@@ -1002,6 +1002,7 @@ provider = "openai"           # API provider: "openai", "anthropic", or "local"
 model = "@openai.default"     # Resolved via api_models registry; edit roles to migrate
 reasoning_effort = "medium"   # Reasoning effort level (low/medium/high)
 max_output_tokens = 25000     # Maximum tokens for generated narrative
+anthropic_storyteller_transport = "prompted" # "prompted" or "native"
 structured_output_retries = 3 # Validation retry budget for structured story turns
 # HTTP client timeout for a narrative continue turn. Frontier generation runs
 # inside the request and blocks the API worker, so status polls aren't answered
--- a/nexus/agents/logon/skald_wire.py
+++ b/nexus/agents/logon/skald_wire.py
@@ -709,8 +709,9 @@ def _prompt_guide_type(
     if ref is not None:
         ref_name = _prompt_guide_ref_name(ref)
         target = definitions[ref_name]
-        target_type = _prompt_guide_type(target, definitions)
-        return ref_name if target_type == "object" else target_type
+        if target.get("type") == "object":
+            return ref_name
+        return _prompt_guide_type(target, definitions)
 
     schema_type = schema_node.get("type")
     if schema_type == "array":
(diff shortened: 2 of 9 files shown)


pythagorakase:
**Independent review gate: CHANGES_REQUIRED — no P1s; two P2s and a P3, all accepted, fix round in flight.**

- **P2 (effort never flows):** accepted. The silent drop of `apex.reasoning_effort` on the Anthropic path is pre-existing, but this PR's claim of effort-only `output_config` makes it config-dishonest to leave — the storyteller construction will pass the configured value through. Side effect worth naming: bootstrap turns (forced native, same construction) start honoring the configured effort as well, which is the config-truthful behavior.
- **P2 (guide renderer inline-object blind spot):** accepted. No current content is affected — Pydantic emits every nested model as a named `$defs` entry — but the latent failure mode (future inline object rendering as bare `object`, properties vanishing, CI green) violates loud-errors doctrine. The renderer will raise on inline objects and the completeness test becomes an independent schema walk.
- **P3 (`getattr` transport default):** accepted — direct attribute access so a transport-less provider fails at the seam.

Reviewer ran the suite itself: 330 passed / 43 skipped, sole failure the documented sandbox-only Keychain write (passes unsandboxed — verified independently). Guide measured at 1,391 o200k tokens by the reviewer's own count. Straightforward scoped fixes → will merge directly after verification per house convention.

pythagorakase:
Fix round verified and landed as 4d33e339: effort pass-through (prompted turns now send effort-only output_config; bootstrap-native honors configured effort), loud inline-object rejection in the guide renderer with an independent-walk completeness test, and direct transport-attribute access. Full gate set green unsandboxed. Merging per house convention for straightforward review-fix commits.

PR #574, opened 2026-07-24
Scale entity inclusion by storyteller provider class (#566)

pythagorakase:
## Summary

Today's live Hermes trial exposed the next bottleneck behind the payload window: the **structured core** (~14.9K tokens of entity dossiers on a mature slot) is invisible to Phase-5 trimming — it's sized by `[lore.entity_inclusion]` regardless of the effective window. This PR shrinks the core at the source for local turns via optional `provider_overrides` tables mirroring the `token_budget` pattern.

## Design

- `[lore.entity_inclusion.provider_overrides.local]`: lookback 20→12, characters 25→12, locations 10→6, `include_all_relationships` off. **Initial estimates** — tonight's overnight run calibrates them against the 32K window.
- Closed outer keys (`openai`/`anthropic`/`local`), unknown inner fields forbidden, loud validation — same doctrine as #570's budget overrides.
- `resolve_entity_inclusion(settings, wire_class)` is the single authority; the turn cycle resolves once from `turn_context.provider_wire_type` (the threading #570 built); `None` is the *named* LOGON-disabled base path; unknown classes raise.
- Limits apply at the SQL fetch boundary; the always-featured user character and explicit character-location IDs stay outside the warm-slice caps.
- Full knob-consumer audit in the build report — including three fields with no runtime consumer today (honestly left base).

## Test Results

```
226 passed, 13 skipped · mypy clean (4 files) · black/flake8 clean · config loads OK
```

### DIFF
--- a/nexus.toml
+++ b/nexus.toml
@@ -254,6 +254,14 @@ max_total_relationships = 100
 max_total_events = 15
 max_total_threats = 10
 
+[lore.entity_inclusion.provider_overrides.local]
+# Initial estimates measured against the 32K local window; calibrated by
+# overnight trials (see #566). Absent key = base value.
+warm_slice_lookback_chunks = 12
+max_characters_from_warm_slice = 12
+max_locations_from_warm_slice = 6
+include_all_relationships = false
+
 # =============================================================================
 # Orrery Settings (Off-Screen Behavior Resolver)
 # =============================================================================
--- a/nexus/agents/lore/utils/entity_inclusion.py
+++ b/nexus/agents/lore/utils/entity_inclusion.py
@@ -1,0 +1,83 @@
+"""Resolve provider-scaled entity inclusion for storyteller payloads."""
+
+from __future__ import annotations
+
+import logging
+from collections.abc import Mapping
+from typing import Any, Optional
+
+from nexus.config.settings_models import EntityInclusionConfig
+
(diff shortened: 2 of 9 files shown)


pythagorakase:
**Independent review: CHANGES_REQUIRED — two P1s and a P2, all real.** (1) The user character consumes the warm cap despite the build report claiming bypass — the cap becomes a cap on non-user characters, with the user attached via the always-featured mechanism. (2) The featured-location bypass is a dead loop that never populated `featured_place_ids` — pre-existing (it shipped with the baseline code), but this PR's 10→6 cap made it acute, and the green suite was hiding it behind a `current_location=None` fixture, which also gets fixed. (3) `DISTINCT ON` lacks a tie-breaker, so the newest chunk's reference-type winner is nondeterministic — explicit priority ordering added. Fix round in flight.

pythagorakase:
Fix round landed: user excluded from the cap NULL-safely inside the capped query with always-featured attachment; the dead featured-location loop implemented for real (FK-or-name resolution, loud on missing/ambiguous, unioned past the cap — the masking None-fixture fixed too); DISTINCT ON totally ordered (setting > transit > mentioned, documented). Rebased onto the two-pass main; suites green before and after. Merging per house convention for review-fix commits implementing decided semantics.

PR #575, opened 2026-07-24
Add the two-pass writer+clerk turn pipeline (#566 bakeoff arm)

pythagorakase:
## Summary

The last ruled piece of the schema-weight campaign: a config-switched **two-pass turn pipeline** for the measurement stage. Pass 1 (writer) authors narrative/choices/scene/presence/operations; pass 2 (clerk) receives the writer's finished beat plus the same context and authors only the structured record (updates/adjudications/declarations). `combine_two_pass()` reassembles the one universal `SkaldTurnWire`; hydration and commit are untouched downstream.

**Why this arm matters:** the writer wire *compiles* under Anthropic native enforcement (probe G2a on #566, 4,251 B) while the full wire cannot — so two-pass lets Anthropic run real grammar enforcement on the writer with a focused prompted clerk (597-token mechanical guide, zero hand-authored schema text) behind it. OpenAI runs both passes strict-native; local runs both lenient.

## Contract Points

- `[apex] turn_pipeline = "single_pass"` (committed default) | `"two_pass"` — a bakeoff lever, not a new default. Bootstrap always single-pass.
- Vocabulary validation attaches to the **clerk only** — every tag-bearing block is clerk-side; the writer's blocks carry no registered tags. Clean custody split.
- Per-pass bounded repair budgets; writer exhaustion short-circuits (no clerk call); clerk exhaustion fails the turn loudly with nothing hydrated. **No silent fallback to single-pass — a failed arm is data.**
- Clerk instructions live in `prompts/storyteller_clerk.md` (house rule), referencing core doctrine rather than duplicating it.
- Regression: shipped Anthropic writer `output_config` has **zero union-typed schema nodes** (real-entry-point assertion); hydration equivalence proven — combined two-pass output hydrates identically to single-pass for identical content.

## Test Results

```
344 passed, 14 skipped · mypy clean (3 files) · black/flake8 clean · config loads OK · deviations: none
```

### DIFF
--- a/nexus.toml
+++ b/nexus.toml
@@ -1014,6 +1014,7 @@ model = "@openai.default"     # Resolved via api_models registry; edit roles to 
 reasoning_effort = "medium"   # Reasoning effort level (low/medium/high)
 max_output_tokens = 25000     # Maximum tokens for generated narrative
 anthropic_storyteller_transport = "prompted" # "prompted" or "native"
+turn_pipeline = "single_pass" # "single_pass" or "two_pass" bakeoff lever
 structured_output_retries = 3 # Validation retry budget for structured story turns
 # HTTP client timeout for a narrative continue turn. Frontier generation runs
 # inside the request and blocks the API worker, so status polls aren't answered
--- a/nexus/agents/logon/skald_wire.py
+++ b/nexus/agents/logon/skald_wire.py
@@ -356,6 +356,50 @@ class UpdatesBlock(BaseModel):
     model_config = ConfigDict(extra="forbid")
 
 
+class SkaldWriterWire(BaseModel):
+    """Provider-facing prose and scene deltas for pass one."""
+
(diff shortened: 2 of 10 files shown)


pythagorakase:
**Independent review: CHANGES_REQUIRED — three P2s, all test-honesty, no production defects.** Accepted in full: validator custody used an inert sentinel (now the real Orrery validator over a ClerkWire with an unregistered tag); the writer-native test stopped before the shipped request builder and never asserted effort behavior; hydration equivalence was circular (both sides through the same helpers — now raw payloads travel each pipeline's real parse path independently). Fix round in flight; suite was verified green (344 passed) by the reviewer's own run.

pythagorakase:
Fix round landed: all three hardenings are test-only (no production defect exposed — the custody, shipped-request, and independent-parse-path tests now fail on the regressions the review described). Verified green locally; merging per house convention.

Issue #578, opened 2026-07-25
Heterogeneous writer/clerk casting on the two-pass pipeline


pythagorakase:
Owner reframe at #566 close: the two-pass pipeline's strategic value is not retry economics — it is **decoupling the prose seat from the registry-discipline seat**, so that models with genuinely interesting prose but unreliable handling of the full structured contract become eligible for the writer chair instead of disqualified outright.

## What exists (shipped in #575/#576)

- `turn_pipeline = "two_pass"`: SkaldWriterWire (narrative/choices/scene/presence/operations) + SkaldClerkWire (updates/orrery_adjudications/new_entities), `combine_two_pass()` → one SkaldTurnWire into unchanged hydration; per-pass repair budgets; vocab validator custody clerk-only.
- Measured 2026-07-25 (#566 final wave): decoupling penalty ≈ zero on the hardest case — same frontier model both seats, latency par with single-pass, one vocab failure repaired clerk-only with writer prose preserved.
- Both passes currently run **the slot model**. There is no per-seat casting.

## What's missing

1. Per-seat model/provider resolution: `writer_model` / `clerk_model` (registry `@provider.role` refs) with the existing route-pinning + mid-turn-drift abort extended to a two-seat route.
2. Writer transport matrix per writer-model capability (native where it compiles, prompted+repair elsewhere; the writer wire is the small half — G2a compiled at 4,251B).
3. Clerk seat inherits the existing schema-free transport selection; clerk-side context could later be right-sized independently (provider-profile already scales by wire class).

## Design floors and notes

- **Presence stays writer-side.** Delta-presence made scene rosters authorial; inferring presence from prose downstream is the #567 pronoun-blindness trap. Writer candidates therefore need light structured-output competence — prompted transport + bounded repair keeps that bar low.
- Casting decisions (which model writes) are taste-side → loom material when concrete candidates exist.
- Evidence in hand: overnight trial profiles — Hermes 4 70B = perfect structural validity, erratic registry vocab (25 violations in one attempt), i.e. exactly a writer-chair candidate whose liability is confined to the clerk seat; Hermes 4.3 36B = zero-issue single-attempt discipline, a natural clerk.
- Non-goal: casting does not fix slow writers (a 70B writer turn is still minutes); it fixes disqualification-by-schema. Cost note: a cheap clerk inverts the 2× input-token objection recorded on #566 — the second call bills at clerk rates.

Related: #416 (local Skald), #567 (detector audit), #566 (measurement record).

[...]

PR #579, opened 2026-07-26
Register the casting roster, key resource profiles by provider name, default two-pass

pythagorakase:
Owner ruling on #578 (2026-07-26): keep every sampled writer-seat candidate selectable for real play, and flip the committed default to the two-pass writer+clerk pipeline (single_pass preserved as an option).

## Registry

- **New `[global.model.api_models.openrouter]` provider** — one OpenAI-compatible funnel for the six sampled prose candidates (Kimi K2.5, Kimi K3, DeepSeek V4 Pro, Hermes 4 405B, MiniMax M3, GLM-5). `api_key_secret = "openrouter"` (key already in the platform store). Zero cost until selected.
- **`claude-opus-5`** added to the Anthropic list (live on the API; flags mirror opus-4-8). Storyteller default role unchanged.
- **`turn_pipeline = "two_pass"`** committed default.

## The one real code change: resource profiles key on provider NAME

`_resolve_storyteller_route` classifies any base_url provider as wire class `local`. Without this change, OpenRouter frontier models would inherit the **32K payload squeeze and thin entity profile** that exist for compute-bound local GGUF serving — sabotaging the evaluation they're being registered for.

Now: the wire class remains the serialization/guard axis (openrouter correctly rides chat_completions + lenient schema), while `provider_overrides` lookup (token budget AND entity inclusion) keys on the registry provider name. For the three existing providers the two strings coincide, so **shipped behavior is byte-identical**; openrouter simply finds no entry and keeps the full frontier window + base inclusion. Legal override keys stay closed (`openai`/`anthropic`/`local`) until a new provider actually needs a constrained profile.

Threading: `resolve_storyteller_route()` → 3-tuple `(model, wire_class, provider_name)`; `TurnContext.provider_name`; both resolvers raise loudly on incoherent route halves; `configure_storyteller_budget(wire, name)`.

## Verification

- `pytest tests/test_skald_wire.py tests/test_native_structured_output.py tests/test_lore/ tests/config/` → **383 passed, 14 skipped** (bare exit, unpiped)
- mypy clean on all 7 touched source files; black + flake8 clean
- New regressions: openrouter-name + local-wire → 75K window (memory-manager seam AND real turn-cycle seam) + base entity inclusion; blank/absent provider name raises; shipped-default test updated to two_pass
- Pre-commit validate-config + drift checker passed on the registry additions

Closes the rung-1 registration on #578. Live two-pass + OpenRouter turn verification happens post-merge on the restarted server (slot 4).

### DIFF
--- a/nexus.toml
+++ b/nexus.toml
@@ -238,8 +238,11 @@ system_prompt_tokens = 5000  # Reserved tokens for system prompt
 prompt_overhead_tokens = 4_000
 
 [lore.token_budget.provider_overrides]
-# Effective apex_context_window per storyteller provider class. Absent key =
-# base value. Local models need a smaller assembled payload (see #566).
+# Effective apex_context_window per REGISTRY PROVIDER NAME (not wire class).
+# Absent key = base value: openrouter shares the "local" wire dialect but has
+# no entry here, so its frontier models keep the full window. This "local"
+# key is the [api_models.local] provider — compute-bound GGUF serving needs a
+# smaller assembled payload (see #566).
 # 32K floor measured live 2026-07-24: a mature slot's untrimmable structured
 # core alone is ~14.9K tokens; 24K left the enforcement nothing to trim to.
 local = 32_000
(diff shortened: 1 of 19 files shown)


pythagorakase:
Codex P1 addressed in fb58d9e3 — confirmed exactly as described, plus the writer half was also broken (no `updates` property → fell through to the cached bootstrap payload). Mock routing now keys on field signatures (narrative+updates = full turn, updates-only = clerk, presence = writer, else bootstrap), and the pass responses are projections of the full fixture so they can't drift out of shape. Three new regressions + a partition guard (every full-fixture key lands in exactly one pass); combined suites 343 passed.

PR #583, opened 2026-07-26
Add per-model request_params passthrough for registry models (#580)

### LINKED ISSUE #580: Per-model request-parameter overrides for registry models (reasoning damping)
Spun out of #578 at close. Kimi K3 (and future reasoning-heavy OpenRouter models) can spend the entire output budget thinking: during #578 sampling, K3 at default reasoning returned **null content with all 1,400 completion tokens consumed**; it sampled well only with `reasoning: {effort: "low"}`. The registry has no way to say that, so K3 sits in the model picker as a slow-fail trap.

Proposal: an optional per-model `request_params` table on registry model entries (e.g. `request_params = { reasoning_effort = "low" }`), merged into chat-completions requests by the OpenAI-compatible provider path. Sibling of the existing `unsupported_params` (which strips; this adds). Loud on unknown keys per config doctrine.

Small, self-contained; touches `settings_models.py` + the chat-completions request builder + one passthrough test.

pythagorakase:
Closes #580. Registry model entries gain an optional `request_params` table merged into every chat-completions call via the OpenAI SDK's **`extra_body`** (the SDK rejects unknown kwargs; OpenAI-compatible servers — OpenRouter, llama-server — read the merged JSON body).

- **Config**: `APIModelEntry.request_params` (default `{}`), with structural request keys (`model`/`messages`/`response_format`/`tools`/`tool_choice`/`stream`/`text`/`extra_body`) reserved and rejected loudly at config load. Sibling of `unsupported_params` (which strips; this adds).
- **Plumbing**: `get_openai_compatible_endpoint` now locates the specific model entry (loud if the id maps to a provider but is absent from its models list) and carries `request_params`; both LOGON provider construction sites (slot provider + pinned clerk) thread it into `OpenAIProvider`.
- **Aliasing**: values deep-copied at both seams — the regression test caught the initial shallow copy letting a mutation of the built request corrupt provider state.
- **Payload**: Kimi K3's registry entry ships `request_params = { reasoning = { effort = "low" } }` — during #578 sampling, K3 at default reasoning returned null content with its entire budget consumed; with damping it sampled well. K3 stops being a slow-fail trap in the model picker.
- Scope: chat-completions transport only (where the need exists); native `responses`/Anthropic paths untouched.

Verification: `tests/test_native_structured_output.py tests/config/ tests/test_lore/` → 329 passed, 13 skipped; mypy/black clean; flake8 clean on changed lines (three pre-existing violations in loader.py's legacy-JSON region predate this PR, verified against origin/main).

### DIFF
--- a/nexus.toml
+++ b/nexus.toml
@@ -88,7 +88,7 @@ request_timeout_seconds = 600
 roles = { default = "moonshotai/kimi-k2.5" }
 models = [
     { id = "moonshotai/kimi-k2.5", label = "Kimi K2.5", description = "Moonshot K2.5 via OpenRouter" },
-    { id = "moonshotai/kimi-k3", label = "Kimi K3", description = "Moonshot K3 via OpenRouter" },
+    { id = "moonshotai/kimi-k3", label = "Kimi K3", description = "Moonshot K3 via OpenRouter", request_params = { reasoning = { effort = "low" } } },
     { id = "deepseek/deepseek-v4-pro", label = "DeepSeek V4 Pro", description = "DeepSeek V4 Pro via OpenRouter" },
     { id = "nousresearch/hermes-4-405b", label = "Hermes 4 405B", description = "Nous Hermes 4 405B via OpenRouter" },
     { id = "minimax/minimax-m3", label = "MiniMax M3", description = "MiniMax M3 via OpenRouter" },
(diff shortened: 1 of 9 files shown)


pythagorakase:
Both reviews addressed in the follow-up commit: (1) all three remaining OpenAIProvider construction sites thread `request_params` (shared factory + Orrery narration), with a factory threading test; (2) the plain-completion chat path merges `extra_body` too — narration through OpenRouter now gets the damping (test-pinned); (3) Codex P2: `request_params` on native providers or responses-transport servers now fails config load loudly; (4) docstring updated; (5) the membership guard documented as a real config-reload race check (the two reads are separate `load_settings()` calls). Suites: 1314 passed.

PR #584, opened 2026-07-26
Add the post-commit presence-roster drift audit (#567)

### LINKED ISSUE #567: Detector audit loop for scene-roster drift
## Summary

Follow-up from the delta-presence ruling on #555 (owner-adopted, shipped in PR #563). The presence roster is now maintained by silence-carries-forward semantics: the writer authors only entries/exits, and a missed exit accretes a stale roster member until the next scene reset. The ruled design named the deterministic alias detector (`nexus/memory/entity_detector.py`) as the AUDIT layer — deployed only where its precision profile is strong, never where its pronoun blindness bites.

## Frozen Doctrine (from the ruling discussion)

- **Flag loudly:** a character name (or alias) appearing in committed prose while absent from the chunk's present roster = candidate missed entry. High-precision signal; surface it.
- **Never trust name-absence:** a roster member whose name does not appear in a chunk is NOT evidence of absence (the pronoun thought experiment: chunk *n* names her, chunk *n+1* says only "she"). Stale-presence cleanup belongs to scene boundaries, not name-matching.
- The detector proposes; it never mutates. Output is diagnostics (log/report), optionally a repair-prompt hint in a later slice — never a silent roster edit.

## Proposed Shape

- Commit-time (or post-commit) audit pass: run the detector over the committed chunk's prose, diff detected names against the chunk's present-typed junction rows, emit structured warnings for prose-present/roster-absent characters.
- Config-gated under `[orrery]` or `[apex]` per repo idiom; default on, log-level warning.
- Metrics-friendly output so the measurement stage can count missed-entry rates per provider.

## Acceptance

- [ ] Audit runs on the real commit path (both handlers) without mutating state.
- [ ] Prose-present/roster-absent character yields exactly one structured warning with chunk id + name.
- [ ] Roster-present/name-absent yields NO warning (pronoun-blindness guardrail, test-enforced).
- [ ] Live-gated test against a slot database.

— Fable (Claude), filing the deferred follow-up from the #555 ruling

pythagorakase:
Closes #567 — the deferred audit-layer follow-up from the #555 delta-presence ruling, built to the frozen doctrine on the issue.

## Shape

- **`nexus/api/presence_audit.py`**: pure diff core (`diff_presence`) + sync orchestrator (psycopg2, runs the real `HighSpecificityEntityDetector` through a tiny execute-adapter over the commit connection) + asyncpg twin (prefetch-fed detector — asyncpg cannot satisfy the detector's sync execute seam).
- **Both commit handlers** call it post-commit, outside the transaction, gated on `[lore.presence_audit] enabled` (shipped `true`).
- One structured warning per finding (`chunk_id`, `name`, `character_id`); findings list returned for metrics.

## Doctrine fidelity

- **Flag loudly**: prose-named character with no junction row → exactly one warning. Detector dedups by id, so name + alias + repeats = one finding (test-pinned).
- **Never trust name-absence**: the module only iterates DETECTED names — structurally incapable of flagging a roster member whose name doesn't appear. Pronoun guardrail test: roster carries her, prose says only "she", zero findings.
- **Proposes, never mutates**: read-only queries; live-gated test asserts junction counts unchanged.
- One deliberate refinement over the issue text: the diff runs against **all** junction reference types, not only `present` — a `mentioned` row is the writer deliberately accounting for an off-scene reference; flagging it would be a false positive on every mention. The signal is *zero rows*.
- One deliberate, documented containment: audit errors after a successful commit are logged loudly and never raised — a read-only observer must not fail a turn whose chunk already committed. This is blast-radius containment, not a fallback masking failure.

## Acceptance (from the issue)

- [x] Audit runs on the real commit path (both handlers) without mutating state
- [x] Prose-present/roster-absent → exactly one structured warning with chunk id + name
- [x] Roster-present/name-absent → NO warning (pronoun-blindness guardrail, test-enforced)
- [x] Live-gated test against a slot database (ran green on save_05)

Suites: commit handlers + config + presence audit → 88 passed; mypy/black/flake8 clean.

### DIFF
--- a/nexus.toml
+++ b/nexus.toml
@@ -289,6 +289,12 @@ max_characters_from_warm_slice = 12
 max_locations_from_warm_slice = 6
 include_all_relationships = false
 
+[lore.presence_audit]
+# Post-commit roster-drift diagnostics (issue #567): warn when committed
+# prose names a character with no junction row for the chunk (candidate
+# missed entry under delta-presence). Read-only; never blocks a turn.
+enabled = true
+
 # =============================================================================
 # Orrery Settings (Off-Screen Behavior Resolver)
 # =============================================================================
--- a/nexus/api/commit_handler.py
+++ b/nexus/api/commit_handler.py
@@ -796,13 +796,20 @@ async def commit_incubator_to_database(
 
     # Post-commit presence-roster drift audit (issue #567): read-only
     # diagnostics over the committed chunk, outside the transaction.
+    # raw_text is the committed prose (storyteller text + enacted choice);
+    # the parent chunk id enables authored-exit accounting.
     from nexus.api.presence_audit import (
         audit_chunk_presence_async,
         presence_audit_enabled,
     )
 
     if presence_audit_enabled():
-        await audit_chunk_presence_async(conn, chunk_id, storyteller_text)
+        await audit_chunk_presence_async(
+            conn,
+            chunk_id,
+            raw_text,
+            parent_chunk_id=incubator.get("parent_chunk_id"),
+        )
 
     logger.info("Successfully committed chunk %s from session %s", chunk_id, session_id)
     return chunk_id
--- a/nexus/api/commit_handler.py
+++ b/nexus/api/commit_handler.py
@@ -794,6 +794,16 @@ async def commit_incubator_to_database(
                 "Failed to schedule summaries for session %s: %s", session_id, exc
             )
 
+    # Post-commit presence-roster drift audit (issue #567): read-only
+    # diagnostics over the committed chunk, outside the transaction.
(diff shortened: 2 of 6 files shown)


pythagorakase:
All three findings addressed: (1) both handlers + the live test now scan **raw_text** (the committed prose — the live test had mirrored the same storyteller_text mistake, exactly as the review noted); (2) sync path shares the characters-only detector builder — no more place/faction load+scan per commit; (3) Codex P2 fixed with **exact** accounting: under delta-presence, silence carries forward, so a parent-present character can only lack a row via an authored `presence.exit` — parent-present ids join the accounted set, and narrated departures stop flagging (regression-pinned). Suites green incl. postgres-gated live run.

PR #586, opened 2026-07-26
Gate maturation replay on checkpoint-recorded job membership (#552)

### LINKED ISSUE #552: Replay delayed maturation writes by application time
## Problem

Runtime Retrograde maturation is detached and can finish well after its requesting chunk commits. Replay currently places maturation project starts at the requesting chunk, and PR #551 applies the same established boundary to ledgered entity-activity projections.

If an interval checkpoint is captured after the requesting chunk but before a delayed maturation job persists, a later verification run can project the maturation output into that earlier checkpoint even though the snapshot did not contain it. For entity activity, the projected stub may also be absent from the replay working set because its created_at is later than the target checkpoint boundary.

## Why this is separate from #545

Issue #545 requires exact durable provenance for entities.is_active. PR #551 supplies that provenance and removes the common false-positive path identified in review. This timing race is inherited from the existing asynchronous maturation/project-start boundary and needs a broader checkpoint-time model.

## Investigation / acceptance

- Distinguish the narrative request boundary from the transaction time at which maturation projections actually commit.
- Decide whether applied-at time belongs in each projection, the maturation job manifest, or another durable ledger.
- Make checkpoint verification aware of the target state_checkpoints.created_at where necessary; arbitrary chunk reconstruction and checkpoint-snapshot verification may need distinct boundary rules.
- Add a rollback-only regression with request chunk A, checkpoint B captured before the worker write, then a delayed maturation persistence transaction, proving B reconstructs without the late project/activity rows.
- Preserve the current base-inclusive/target-exclusive behavior for maturation writes that commit before the relevant target checkpoint.

Codex — GPT-5.6-Sol

pythagorakase:
Closes #552 — the timing race Sol filed: runtime maturation persistence is detached from its requesting chunk, so an interval checkpoint captured **before** a delayed worker transaction gets that worker's expansion projected into it by later verification (and the projected stub can be absent from the replay working set entirely). Chunk-keyed windows cannot express "committed after the capture."

## Custody decision (the issue's investigation ask)

**Applied-at lives in the checkpoint document.** Each capture records the set of maturation jobs whose expansion writes are visible in the capture transaction (`_maturation_jobs_succeeded`, underscore-prefixed and absent from `CHECKPOINT_SECTIONS`, so verification never diffs it as state). MVCC makes the set exact by construction — whatever the SELECT sees is precisely what the section snapshots saw. No migration, no timestamps, no clock comparisons.

Two subtleties the investigation surfaced:
- **The membership predicate is `result_manifest.persisted`**, recorded by the worker in the SAME transaction as its writes — NOT `state = 'succeeded'`, which flips in a later transaction after embedding and would open a race band where a capture sees the writes but not the membership.
- The `target ∖ base` rule also retires a **latent double-apply**: a job persisted before the base capture had its writes in the base document AND replayed again through the inverse chunk window.

## Boundary regimes (per acceptance)

- **Checkpoint-targeted reconstruction** (verify now passes `target_checkpoint_id`): include a maturation write iff its job ∈ target set ∖ base set.
- **Arbitrary chunk reconstruction** and **cross-era pairs** (documents predating the key): legacy chunk-window boundary, noted as approximate.
- Gated replay refuses loudly on maturation events without a parseable `maturation_job_<id>` ref (unattributable provenance).

## Verification

- Acceptance regression exactly as specified: request chunk A → checkpoint captured → delayed job+event insertion → gated reconstruction clean (the event's missing `applied` projection is a tripwire that raises if replay ever reads it), ungated reconstruction still trips (distinct regimes proven in one test).
- No-over-exclusion regression (between-capture jobs replay) + cross-era fallback regression.
- Three existing maturation-drift fixtures now model job provenance — their missing job rows were exactly the fabricated-provenance the gate exists to catch (manifests satisfy the migration-078 v1 trigger).
- **Standing oracle green: `replay_state.py --slot 5 --verify` — all four checkpoint pairs ok** (pre-fix documents, cross-era path).
- Suites: orrery+config 1,056 passed dry; **1,294 passed under NEXUS_RUN_POSTGRES=1**; mypy/black/flake8 clean.

### DIFF
--- a/nexus/agents/orrery/reconstruction.py
+++ b/nexus/agents/orrery/reconstruction.py
@@ -155,9 +155,10 @@ _ADDITIVE_CHECKPOINT_TABLES = {"backstory_secrets": "backstory_secrets"}
 # are visible inside the capture transaction. Underscore-prefixed and
 # deliberately absent from CHECKPOINT_SECTIONS so verification never diffs it
 # as state; replay uses target-minus-base membership as the exact applied-at
-# boundary for asynchronously persisted maturation writes. MVCC gives the
-# exactness: whatever this SELECT can see is precisely what the section
-# snapshots in the same transaction saw.
+# boundary for asynchronously persisted maturation writes. Statement-level
+# MVCC gives the exactness: the control set and every section are read in
+# ONE statement (see _checkpoint_document_sql), so they share one snapshot
+# even under READ COMMITTED.
 #
 # The predicate is result_manifest.persisted — recorded by the worker in the
 # SAME transaction as the expansion writes — NOT state = 'succeeded', which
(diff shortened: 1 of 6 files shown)


pythagorakase:
All three findings addressed: (1) **capture is now one statement** — every section plus the control set share a single READ COMMITTED snapshot (jsonb_build_object concatenation), closing the mid-capture worker race and incidentally giving the sections cross-coherence they never had; (2) **full gating coverage** — delayed jobs' tag/claim/secret/relationship writes (chunk-attributed, no job provenance) are handled by deriving the delayed-chunk set and marking their rows presence-uncertain through the existing per-row skip mechanism, with a named note; relationship inserts were already covered by the wall-clock filter, update/delete pre-images marked via relationship_versions; (3) **preflight windowed** to the pair's maturation range — out-of-window orphans can't abort a healthy pair and the scan no longer reruns globally. Acceptance test extended with a planted delayed pair-tag row proving zero-drift + uncertain-skip (after fixing my own first fixture, which landed on neither side of the diff and proved nothing). Full battery: 1,056 dry / 1,294 live green; slot-5 oracle all pairs ok.

PR #597, opened 2026-07-27
Prompt-Family Voice Polish: Third Person, One Line per Paragraph

pythagorakase:
Four owner notes on the persona documents from today's review of the Gaia rewrite, applied family-wide.

## Changes

- **Reflow to core's convention.** `storyteller_gaia.md`, `storyteller_single_pass.md`, and `storyteller_writer_pass.md` were hard-wrapped at ~72 chars by the #592/#594 round; `storyteller_core.md` keeps one line per paragraph. All three now match core.
- **Third-person descriptive voice.** Core's pattern is "You are Skald." followed by third-person description ("Skald writes...", "Skald is patient..."); the newer documents used second-person imperatives. All three converted: "You are Gaia." then Gaia does — same for the two Skald supplements.
- **Drop "Return only the structured output."** from gaia.md. It is dead weight on every transport: the OpenAI and Anthropic structured paths constrain output mechanically, and the prompted-transport guide (`skald_gaia_prompt_guide`) already opens with "Respond with a single JSON object... No prose outside the JSON."
- **Drop the etymology lesson** from core's opening ("The name is old Norse: ..."). The court-poet image survives as a present-tense persona statement — it is the Bard ingredient, and worth keeping without the lecture.

## Tests

Pins updated to the new voice (`tests/test_skald_wire.py`, `tests/test_lore/test_two_pass_pipeline.py`). The shared-invariants test still demands verbatim agreement between the single-pass supplement and gaia.md — both documents now pin "accepts by silence" exactly, so the anti-drift tripwire stays strict.

`tests/test_lore` + `tests/test_skald_wire.py`: 319 passed, 14 skipped.

https://claude.ai/code/session_01EybUa5fEuX5JvX9Wz7s1UW


pythagorakase:
Findings 1 and 3 fixed in b9f5dfa6: all three `rather than <bare-verb>` leftovers now take the gerund, and the adjudication sentence recovers its literal-value cue in third person ("answers with `defer` … `void` … `replace`").

Finding 2 is the intentional deletion (owner-directed, this PR's third bullet): "Return only the structured output" is enforced mechanically on every transport Gaia can travel — OpenAI/Anthropic structured paths constrain output at the grammar/tool layer, and the prompted-JSON guide appended on the local path opens with "Respond with a single JSON object… No prose outside the JSON." The directive had no path where it could do work.

PR #603, opened 2026-07-28
LORE Honors NEXUS_RUNTIME_CONFIG for Narrative Turns

### LINKED ISSUE #599: Narrative LORE ignores `NEXUS_RUNTIME_CONFIG`, causing isolated runtimes to use the repository Gaia model
### Impact

High. `nexus up --config ...` claims that spawned services honor the selected
configuration, but ordinary narrative turns silently reload the repository's
`nexus.toml`. Model routing, cost, provider, prompt, token-budget, and other
LORE settings can therefore diverge from the runtime the operator validated.

This directly violated the QA requirement that both Skald and Gaia use Terra.

### Reproduction

1. Copy `nexus.toml` to an alternate config.
2. In the alternate config set:

   ```toml
   [apex]
   model = "gpt-5.6-terra"
   gaia_model = "gpt-5.6-terra"
   ```

3. Validate the alternate file:

   ```text
   typed gpt-5.6-terra gpt-5.6-terra
   legacy gpt-5.6-terra gpt-5.6-terra
   ```

4. Start an isolated runtime:

   ```bash
   NEXUS_GATEWAY_PORT=8012 \
     poetry run nexus up --slot 3 --config temp/qa_slot3_terra.toml
   ```

5. Confirm the gateway process environment contains:

   ```text
   NEXUS_GATEWAY_PORT=8012
   NEXUS_SLOT=3
   NEXUS_RUNTIME_CONFIG=/Users/pythagor/nexus/temp/qa_slot3_terra.toml
   ```

6. Generate a two-pass narrative turn.

### Observed

The writer used Terra, then Gaia used the repository default Sol:

```text
Using reasoning model: gpt-5.6-terra with effort: medium
Loaded storyteller Gaia prompt
Using reasoning model: gpt-5.6-sol with effort: medium
```

After temporarily pinning the repository config as well, regeneration logged
Terra for LOGON and cloned that provider for Gaia. The replacement incubator
row records `generation_model=gpt-5.6-terra`; the Sol draft was never accepted.

### Expected

Every component in a runtime spawned with `--config PATH` should load `PATH`,
including LORE and both storyteller passes.

### Root cause

`nexus/api/narrative_generation.py` constructs:

```python
LORE(enable_logon=True, debug=True, slot=slot)
```

When `settings_path` is omitted, `LORE._load_settings()` explicitly constructs
the repository-root `nexus.toml` path and passes it to
`load_settings_as_dict()`. That bypasses the loader's documented
`NEXUS_RUNTIME_CONFIG` behavior.

Relevant files:

- `nexus/api/narrative_generation.py`
- `nexus/agents/lore/lore.py`
- `nexus/config/loader.py`
- `docs/runtime.md`

### Acceptance criteria

- `LORE(settings_path=None)` honors `NEXUS_RUNTIME_CONFIG`.
- A managed-runtime integration test launches with an alternate config whose
  writer/Gaia values differ from repository defaults and asserts both actual
  provider calls use the alternate values.
- Explicit `settings_path` continues to override the environment.
- Runtime status and generation traces report the same effective config path.

---

Codex — GPT-5.6-Sol

pythagorakase:
Fixes #599 (found by the 2026-07-28 midnight adversarial QA run, `temp/slot3_midnight_qa_2026-07-28.md`).

## Problem

`LORE._load_settings()` hard-constructed the repository-root `nexus.toml` path whenever `settings_path` was omitted, bypassing the loader's documented `NEXUS_RUNTIME_CONFIG` fallback. Isolated runtimes started with `nexus up --config PATH` silently ran narrative turns against repository defaults — observed live as the writer pass using the pinned Terra model while the Gaia pass fell back to repo-default Sol.

## Fix

- Precedence in `LORE._load_settings()`: explicit `settings_path` > `NEXUS_RUNTIME_CONFIG` > repository root. The env-var name is imported from `nexus.config.loader.RUNTIME_CONFIG_ENV` (single source).
- The effective path is resolved absolute, stored as `lore.settings_path`, and logged by both LORE and `narrative_generation`, so generation traces agree with `nexus status`.

## Tests

`tests/test_lore/test_runtime_config.py` — real loader, real temp TOML, real env var; no mocks. The alternate config is derived by swapping the writer/Gaia registry role references from the live `[global.model.api_models]` registry, so no model IDs are pinned. Covers: env var honored for both storyteller seats (asserted through the LOGON routing boundary), explicit path beats env, unset env falls back to repo root independent of cwd.

```
3 passed in 4.40s  (tests/test_lore/test_runtime_config.py)
218 passed, 2 failed in 23.30s  (full tests/test_lore/)
```

The 2 failures (`test_logon_initializes_on_first_use`, `test_handle_user_input_writes_exact_coverage_and_empty_detection`) were verified to fail identically on unmodified `main` — pre-existing, unrelated, tracking issue to follow. Black clean; the 29 remaining mypy errors in `lore.py` are pre-existing typing debt.

### DIFF
--- a/nexus/agents/lore/logon_utility.py
+++ b/nexus/agents/lore/logon_utility.py
@@ -11,7 +11,7 @@ import logging
 import os
 import sys
 from pathlib import Path
-from typing import Any, Dict, Literal, Mapping, Optional, cast
+from typing import Any, Dict, Literal, Mapping, Optional, Union, cast
 
 import psycopg2
 
@@ -280,6 +280,7 @@ class LogonUtility:
         dbname: Optional[str] = None,
         model_override: Optional[str] = None,
         bootstrap_mode: bool = False,
+        settings_path: Optional[Union[str, Path]] = None,
     ):
         """
         Initialize LOGON utility with configured provider.
@@ -291,11 +292,16 @@ class LogonUtility:
             model_override: Optional model to use instead of settings/slot config.
                            If None, will check slot's configured model first.
             bootstrap_mode: Whether this LOGON instance is generating chunk #1.
+            settings_path: Effective configuration path that owns this LOGON
+                stack. Registry lookups remain bound to it when provided.
         """
         self.settings = settings
         self.dbname = dbname
         self.model_override = model_override
         self.bootstrap_mode = bootstrap_mode
+        self.settings_path = (
+            Path(settings_path).resolve() if settings_path is not None else None
+        )
         self.provider: Optional[OpenAIProvider | AnthropicProvider] = None
         self._system_prompt: Optional[str] = None
         self._provider_bootstrap_mode: Optional[bool] = None
@@ -511,9 +517,13 @@ class LogonUtility:
             return None
 
     @staticmethod
-    def _resolve_generation_model(model: str) -> str:
+    def _resolve_generation_model(
+        model: str, settings_path: Optional[Path] = None
+    ) -> str:
         """Resolve a runtime roster reference before constructing the provider."""
-        return resolve_model_ref(model) if model.startswith("@") else model
+        return (
+            resolve_model_ref(model, settings_path) if model.startswith("@") else model
+        )
 
     def _resolve_storyteller_route(self) -> StorytellerRoute:
         """Resolve the active model, endpoint, and storyteller wire class."""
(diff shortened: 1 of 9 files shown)


pythagorakase:
Independent Sol review returned `CHANGES_REQUIRED` (1 P1, 1 P2, 1 P3). All three verified in code and fixed in 372db5eb. Point by point:

**[P1] MEMNON legacy `NEXUS_SETTINGS_PATH` precedence** — *Confirmed and fixed.* Worse than reported: the same `load_settings()` also swallowed every loader error into `{}`, silently degrading retrieval to empty settings (Sol's own probe demonstrated the mask). The legacy branch is removed (its only remaining user is the exempt ir_eval V1 tooling), the central loader now owns precedence, and loader failures abort module import loudly. New fresh-interpreter integration tests (`tests/test_memnon_runtime_config.py`) cover runtime-config precedence over a legacy-style env and the loud failure path — real subprocesses, real TOML, no mocks.

**[P2] LOGON routing re-reads the environment under an explicit `settings_path`** — *Confirmed and fixed.* `get_provider_for_model` / `get_openai_compatible_endpoint` gained optional config-path parameters; `LogonUtility` now receives the owning LORE's effective path and binds writer **and** Gaia route resolution to it. Regression test resolves an explicit config while `NEXUS_RUNTIME_CONFIG` points at a deliberately missing file.

**[P3] Fixture assumes committed apex seats differ** — *Confirmed and fixed.* The helper now derives two distinct models from the registry itself (role refs preferred, registry literals as fallback) and skips only when the registry genuinely has fewer than two distinct models. Zero pinned IDs.

Orchestrator follow-through: collapsed the redundant `None`-guard conditionals around the new path parameters, which exposed five test files stubbing these helpers with stale single-arg lambdas — stubs updated to the real signatures rather than letting production code contort to fit them. Full affected suite: **283 passed**; the only remaining failures are the two pre-existing `main` reds tracked in #604.

PR #605, opened 2026-07-28
Trait Compiler: Pre-Resolve Shared Named Targets for Permutation Invariance

### LINKED ISSUE #602: Trait compiler resolves Status before Obligations creates the same faction, leaving a false `unknown_scope_faction` remainder
### Impact

Medium-high. Two selected traits reference the same faction. One trait creates
the faction later in the same transaction, but an earlier trait has already
been abandoned as prose-only. Mechanical output therefore depends on trait
ordering instead of the complete typed input set.

### Reproduction

The selected trait order was:

```text
status, enemies, obligations
```

Derived Status and Obligations inputs both referenced:

```text
Dunlow County Circuit Court
```

### Observed

The persisted compile result contains both:

```text
prose_only_remainder:
  trait: status
  reason_code: unknown_scope_faction
  scope_faction_name: Dunlow County Circuit Court

created_entity:
  trait: obligations
  entity_kind: faction
  name: Dunlow County Circuit Court
```

Compact live query:

```text
remainder_0           missing_scope                 created_entity
unknown_scope_faction Dunlow County Circuit Court   Dunlow County Circuit Court
```

At transaction completion the supposedly unknown faction exists, but Status
has no mechanical pair tag.

### Expected

Trait compilation should be permutation-invariant for a fixed character and
typed-input set. All stub-creatable shared targets should be materialized or
planned before trait-specific application.

### Root cause

`compile_character_traits()` loops over `character.get_trait_entries()` and
applies each trait immediately:

- `_compile_status()` only looks up an existing faction and emits
  `unknown_scope_faction` if absent.
- `_compile_obligations()` later calls `_resolve_faction_target()`, which is
  allowed to create the missing faction stub.
- There is no pre-resolution/materialization pass and no retry of earlier
  remainders.

Relevant file:

- `nexus/api/trait_compiler.py`

Related closed issues: #290 and #295 established structured remainders and the
compiler pipeline; this is a live order-dependence regression/gap in the
implemented compiler.

### Acceptance criteria

- Compiling the same three traits in every permutation yields identical
  canonical writes and remainders.
- Shared named targets are resolved/materialized before per-trait dispatch.
- Status successfully applies when another selected trait creates its scope
  faction in the same compilation.
- Dry-run and apply mode report the same planned target identity and outcome.

---

Codex — GPT-5.6-Sol

pythagorakase:
Fixes #602 (midnight adversarial QA finding 5).

## Problem
`compile_character_traits()` applied traits in selection order with immediate per-trait resolution. Status only looks up existing factions; Obligations may create them. With both referencing "Dunlow County Circuit Court", order decided whether Status got its mechanical pair tag or a false `unknown_scope_faction` remainder.

## Fix
A deterministic pre-pass collects stub-creatable named targets from all selected typed inputs (domain, patron, dependents, obligations — the traits already licensed to create) in fixed package order, resolving/creating/planning each `(kind, name)` exactly once before dispatch. Every compiler consults the shared map first; `_compile_status` binds a scope faction any sibling materializes while still never creating entities itself. Registry gates are checked in the pre-pass so no pair-tag behavior widens. Dry-run carries pending identities (`object_name`) so dry-run and apply report the same planned target.

## Tests
`test_dunlow_shared_faction_is_permutation_invariant_on_save_05` — real `compile_character_traits` entry point, all six permutations of the live trio, dry-run and apply compared for identical canonical writes and remainders, rollback-only against `save_05` per the file's convention. 36 passed across both trait-compiler suites; mypy and black clean.

### DIFF
--- a/nexus/api/trait_compiler.py
+++ b/nexus/api/trait_compiler.py
@@ -14,7 +14,11 @@ from nexus.agents.orrery.status_family import (
     normalize_status_level,
     status_tag_for_level,
 )
-from nexus.agents.orrery.substrate import CONTACT_PAIR_TAGS, contact_pair_tag_for_kind
+from nexus.agents.orrery.substrate import (
+    CONTACT_PAIR_TAGS,
+    ContactKind,
+    contact_pair_tag_for_kind,
+)
 from nexus.agents.orrery.tag_writer import (
     apply_exclusive_tag_bestowal,
     apply_pair_tag_bestowal,
@@ -96,6 +100,19 @@ class _ResolvedTarget:
     pending_stub: bool = False
 
 
+@dataclass(frozen=True)
+class _NamedTargetCandidate:
+    """One selected trait target that may create a named entity stub."""
+
+    entity_kind: str
+    name: str
+    trait: str
+    role: str
+
+
+_ResolvedNamedTargets = dict[tuple[str, str], _ResolvedTarget]
+
+
 PAIR_TAG_RELATIONSHIP_TYPES = frozenset({"ally", "hostile_to"}) | frozenset(
     CONTACT_PAIR_TAGS.values()
 )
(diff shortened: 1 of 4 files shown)


pythagorakase:
Sol review returned `CHANGES_REQUIRED` (1 P2, 1 P3); both verified and fixed in the follow-up commit.

**[P2] Relationship writes order-dependent for shared characters** — *Confirmed (live-probed on save_05) and fixed.* Schema check confirmed `character_relationships` permits exactly one row per ordered pair (primary key on `character1_id, character2_id`), so this was a genuine semantic collision, not an upsert bug. Resolution: the pre-pass package precedence (Patron > Dependents > Obligations) deterministically owns the pair; the winner writes the row, and each displaced trait emits a structured `relationship_pair_conflict` remainder naming the winner, target, and displaced relationship type — permutation-invariant and never silent. Dry-run and apply agree.

**[P3] Permutation test compared too little** — *Confirmed and fixed.* The cross-permutation comparison now normalizes and compares the full result surface (created entities with attribution, created relationships, applied pair tags, remainders), and the reviewer's shared-character scenario is a second six-permutation case with persisted-row probes.

Re-verified independently: 37 passed with Postgres enabled; mypy and black clean.

PR #607, opened 2026-07-28
Wizard: Share Traits-to-Wildcard Transition and Refuse Empty Continues

### LINKED ISSUE #598: Free-text trait confirmation returns no next action; CLI continue sends an empty OpenAI thread message and fails
### Impact

High. A valid wizard path becomes actionless. The user has completed the
traits subphase, but the response exposes neither the wildcard prompt nor a
valid transition control. The most natural recovery command produces an
upstream 400, surfaced by NEXUS as a server error.

### Reproduction

1. Start a fresh real-model wizard.
2. Reach the character trait menu.
3. Confirm three traits through free text rather than deterministic
   `--choice 0`. This live run ended with:

   ```text
   Exactly those three: 5, 9, 10.
   ```

4. Observe the submission response:

   ```json
   {
     "artifact_type": "submit_trait_selection",
     "phase": "character",
     "phase_complete": false,
     "subphase": null,
     "can_confirm": false,
     "choices": [],
     "message": "Generating artifact..."
   }
   ```

5. Run the only discoverable next action:

   ```bash
   poetry run nexus --json continue --slot 3 --model gpt-5.6-terra
   ```

### Observed

```text
API error: {"detail":"Error code: 400 - {'error': {
  'message': 'Message content must be non-empty.',
  'type': 'invalid_request_error',
  'param': 'content'
}}"}
```

Gateway evidence:

```text
nexus.api.wizard_chat - ERROR - Error in chat endpoint:
Message content must be non-empty.
POST /api/story/new/chat HTTP/1.1 500 Internal Server Error
```

Arbitrary filler text such as “Continue to the next setup step” recovers the
wizard and produces the wildcard introduction, proving that the stored wizard
state itself is recoverable.

### Expected

The trait submission should atomically advance to wildcard and return its
intro/choices, or the CLI should consume the backend's
`subphase_complete=true` signal and request the wildcard intro. No normal CLI
path should send an empty user message to OpenAI.

### Likely cause

- `wizard_agent._submit_traits_impl()` correctly emits
  `subphase_complete=True`.
- The standard free-text branch in `cli.run_continue()` does not copy
  `subphase_complete` into its result and only auto-transitions on
  `phase_complete`.
- On the next call, `run_continue()` constructs
  `user_text = args.user_text or ""`.
- `wizard_chat.new_story_chat_endpoint()` forwards that empty value to
  `ConversationsClient.add_message()` without a non-empty guard.

Relevant files:

- `nexus/api/wizard_agent.py`
- `nexus/cli.py`
- `nexus/api/wizard_chat.py`

### Acceptance criteria

- A free-text `submit_trait_selection` response immediately exposes the
  wildcard step, matching deterministic trait confirmation.
- A plain CLI `continue` after trait confirmation never sends an empty
  OpenAI thread message.
- The API returns a local, actionable 4xx for any genuinely invalid empty
  wizard message rather than converting an upstream 400 into a 500.
- Tests cover free-text trait submission, deterministic trait confirmation,
  and the next CLI action.

---

Codex — GPT-5.6-Sol

pythagorakase:
Fixes #598 (midnight adversarial QA finding 1).

## Fix
- `_submit_traits_impl()` emits `subphase="wildcard"` so clients need not infer the destination.
- CLI: transition logic extracted to `_apply_traits_to_wildcard_transition()` and applied on **both** the deterministic and free-text confirmation routes (previously only `--choice` got the wildcard intro).
- CLI: plain wizard `continue` with no text/choice/accept-fate returns a local actionable error without POSTing.
- API: `ChatRequest` model-validator rejects blank conversational turns locally (422) while permitting deterministic `accept_fate`/`trait_choice` turns — the upstream OpenAI 400→500 conversion is unreachable for this class.

## Tests
CLI transition coverage for both routes plus plain-follow-up behavior; backend transition-signal assertion; local 422 validation tests; and the live wizard matrix extended with a real free-text `submit_trait_selection` phase — 13 live tests passed (4:36, registry-role models). 41 unit/API tests passed; black clean; new surfaces mypy-clean.

### DIFF
--- a/nexus/api/narrative_schemas.py
+++ b/nexus/api/narrative_schemas.py
@@ -8,7 +8,7 @@ serialization of API requests and responses.
 from datetime import datetime
 from typing import Dict, List, Optional, Any, Literal
 
-from pydantic import BaseModel, Field, field_validator
+from pydantic import BaseModel, Field, field_validator, model_validator
 
 # Re-export ChoiceSelection from shared module for backward compatibility
 from nexus.api.choice_handling import ChoiceSelection
@@ -276,6 +276,20 @@ class ChatRequest(BaseModel):
     # Trait selection operations (character phase, traits subphase)
     # 0 = confirm selection, 1-10 = toggle trait by ID
     trait_choice: Optional[int] = None
+
+    @model_validator(mode="after")
+    def validate_wizard_input(self) -> "ChatRequest":
+        """Reject blank conversational turns before any provider call."""
+        if (
+            not self.message.strip()
+            and not self.accept_fate
+            and self.trait_choice is None
+        ):
+            raise ValueError(
+                "Wizard message must be non-empty unless accept_fate or "
+                "trait_choice is provided."
+            )
+        return self
 
     @field_validator("model")
     @classmethod
--- a/nexus/api/wizard_agent.py
+++ b/nexus/api/wizard_agent.py
@@ -427,6 +427,7 @@ async def _submit_traits_impl(
         "phase_complete": phase_complete,
         "subphase_complete": True,
         "phase": ctx.deps.phase,
+        "subphase": "wildcard",
         "artifact_type": "submit_trait_selection",
         "data": response_data,
     }
--- a/nexus/api/wizard_chat.py
+++ b/nexus/api/wizard_chat.py
@@ -95,6 +95,25 @@ def wizard_model_lock_candidate(
     return bool(request_model and slot_model and request_model != slot_model)
 
 
+def _wizard_subphase_for_state(
+    phase: str,
+    *,
+    has_concept: bool,
+    has_traits: bool,
+    has_wildcard: bool,
+) -> str:
+    """Name the canonical wizard subphase represented by persisted state."""
+    if phase != "character":
+        return "none"
+    if not has_concept:
+        return "concept"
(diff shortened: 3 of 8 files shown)


pythagorakase:
Sol review returned `CHANGES_REQUIRED` (2 P2s); both verified via Sol's own live probes and fixed in the follow-up commit (branch also rebased onto post-#605 `main`).

**[P2] `trait_choice` bypasses blank validation out of state** — *Confirmed and fixed.* The schema validator cannot know wizard state, so the guard now lives server-side: `wizard_chat` derives the canonical phase/subphase from persisted `WizardState` immediately after slot lookup and returns a local 409 naming the current state (e.g. `character/wildcard` for a repeated confirmation, `setting/none` for a trait flag during setting) before hydration or any provider call. Both probe cases are covered by new endpoint tests.

**[P2] Silent intro-fetch degrade with no recovery** — *Confirmed and fixed.* The trait commit's success is preserved, but the result now carries a structured `intro_error` (status + detail, covering both HTTP failures and request exceptions), a machine-readable `intro_recovery_command`, and human-visible text confirming traits were saved and naming the exact recovery command.

Re-verified independently after rebase: 44 passed; black clean; new surfaces mypy-clean.

PR #618, opened 2026-07-29
CLI: Surface Seed-Completion Transition Failures Loudly

### LINKED ISSUE #611: Seed completion swallows transition failures and reports success while the wizard remains ready
## Impact

High. The final seed command can spend more than a minute in Retrograde, fail
the atomic transition, and still exit zero with `"success": true`. The output
contains no error, recovery command, or indication that no world was created.
Automation and a sleeping operator both interpret a failed story creation as
success.

## Live reproduction

The seed submission completed normally:

```text
nexus --json continue --slot 3 --model gpt-5.6-terra --choice 2
```

CLI result:

```json
{
  "artifact_type": "submit_starting_scenario",
  "phase": "seed",
  "phase_complete": true,
  "success": true,
  "choices": []
}
```

The same request's gateway log ended with:

```text
Atomic transition failed for Imani Vale:
Retrograde project seed 'seed_01' seek_redemption requires a
TARGET->ACTOR wary-or-worse relationship
POST /api/story/new/transition HTTP/1.1 400 Bad Request
```

The slot remained in wizard mode:

```json
{"phase":"ready","message":"Slot 3 is in wizard mode.","success":true}
```

Running plain `continue` from that ready state correctly surfaced the same
transition error and exited nonzero. Only the inline seed-completion path hid
it.

## Root cause

In `nexus/cli.py`, the seed-completion branch checks
`if transition_response.ok:` and populates bootstrap output only on success.
There is no `else`; execution falls through to `return result`, preserving the
earlier successful seed artifact.

## Acceptance criteria

- Any non-2xx transition response makes the CLI command fail nonzero.
- JSON output includes the transition status/detail and a retry command.
- The successfully persisted seed artifact may be reported separately, but
  overall `success` must be false while the slot remains wizard-ready.
- Tests cover 400, 422, 500, timeout, and successful transition responses.

---

Codex — GPT-5.6-Sol

pythagorakase:
## Summary

Fixes #611 (night-QA round 2). The seed-completion branch in `run_continue` checked `if transition_response.ok:` with no `else`, so a failed atomic transition fell through to `return result` — reporting `success: true` and exiting 0 while the slot remained in wizard mode.

## Changes

- `_seed_transition_failure` helper: marks the result `success=false`, attaches `transition_error` (status, status_code, detail) and `retry_command`, and preserves the separately-persisted seed artifact fields.
- Timeout and non-2xx transition responses both route through it; the success path is de-nested.
- JSON mode emits the full structured failure to stderr (upgrade from the bare `{"error": ...}` shape); exit code is 1.

## Testing

Tests drive `cli.main()` via argv — the genuine public entry — with the gateway stubbed only at the `requests` boundary, matching the existing `test_trait_confirmation_intro_failure_exposes_recovery` idiom. Coverage: 400/422/500 (parametrized), timeout, and successful transition+bootstrap.

- Focused: `26 passed`
- Full suite: `1955 passed, 450 skipped`

### DIFF
--- a/nexus/cli.py
+++ b/nexus/cli.py
@@ -1139,6 +1139,34 @@ def _generation_timeout_seconds() -> int:
     return load_settings().apex.generation_timeout_seconds
 
 
+def _seed_transition_failure(
+    *,
+    result: Dict[str, Any],
+    slot: int,
+    detail: str,
+    status_code: Optional[int],
+    status: str,
+) -> Dict[str, Any]:
+    """Report a persisted seed whose narrative transition did not complete."""
+    retry_command = f"nexus continue --slot {slot}"
+    result.update(
+        {
+            "success": False,
+            "error": (
+                "Seed artifact was saved, but the narrative transition failed. "
+                f"Retry with: {retry_command}"
+            ),
+            "transition_error": {
+                "status": status,
+                "status_code": status_code,
+                "detail": detail,
+            },
+            "retry_command": retry_command,
+        }
+    )
+    return result
+
+
 def _apply_traits_to_wildcard_transition(
     *,
     url: str,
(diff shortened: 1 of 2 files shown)


pythagorakase:
## Sol Review Response

Independent Sol review returned one finding; addressed:

**[P2] `Response.ok` accepts 3xx, violating the every-non-2xx exit contract** — **Fixed.** Sol's live probe showed an HTTP 300 passing the `ok` guard, proceeding to bootstrap, and exiting 0 with `success: true`. The guard is now `200 <= status_code < 300`, and the failure test gained a `(300, ok=True)` parametrized case that pins exactly this distinction (the dummy keeps `ok=True` to mirror real `requests` semantics for 3xx).

Sol's other probes all confirmed the contract: the advertised retry command genuinely retries the transition from wizard-ready state; no existing consumer parses the old `{"error": ...}` stderr shape; the success path is byte-identical modulo indentation; and reverting the fix fails the new tests. (Sol's in-sandbox pytest runs failed on a read-only TMPDIR — environmental; its direct `cli.main()` probes did the verification, and the full suite is green here: 1956 passed.)

Issue #616, opened 2026-07-29
Concurrent continues run duplicate turn pipelines and overwrite the singleton incubator


pythagorakase:
## Impact

Critical for multi-client/retry safety. Two requests against the same pending
turn both report success and charge full generation work, but only one result
can exist. The losing caller receives a valid session ID whose “completed”
turn has already been replaced by another session.

## Live reproduction

Two simultaneous real API requests selected the same pending choice:

```http
POST /api/narrative/continue
{"slot":3,"choice":1}
```

Both returned HTTP 200:

```json
{"session_id":"760d7cb6-...","status":"processing",
 "message":"Narrative generation started for chunk 2"}

{"session_id":"40717f95-...","status":"processing",
 "message":"Narrative generation started for chunk 2"}
```

Both status endpoints later returned:

```text
status=complete
chunk_id=3
```

Logs showed two independent LORE stacks and two embedding jobs for locked
chunk 1. The singleton incubator contained only:

```text
chunk_id=3
parent_chunk_id=2
session_id=40717f95-...
```

The result attributed to session `760d7cb6-...` no longer existed.

## Root cause

There is no per-slot generation lease around resolve → approve → schedule.
Each request creates a new session and background task. On completion,
`write_to_incubator()` unconditionally deletes the singleton row before
inserting its own result, so last writer wins while both session-progress
records say complete.

## Acceptance criteria

- At most one continue/regenerate pipeline may own a slot at a time.
- A competing request receives 409 with the active session ID, or an
  idempotent retry joins the existing session.
- Incubator persistence is conditional on the expected parent/session and
  cannot delete another generation's result.
- Session status can never report a completed loadable result that does not
  belong to that session.
- Locked-chunk embedding is scheduled once per accepted parent.

---

Codex — GPT-5.6-Sol

Issue #615, opened 2026-07-29
Plain narrative continue skips an unresolved player choice and generates from empty input


pythagorakase:
## Impact

High. The narrative can advance without any player decision even though the
current chunk presents choices and `--accept-fate` exists for intentional
automatic selection. The committed parent retains `selected=null` and empty
`choice_text`, while a child is generated from empty `user_text`.

## Live reproduction

1. Commit chunk 1 by selecting choice 3, producing pending chunk 2.
2. Run `nexus undo --slot 3`.
3. Verify undo deletes pending chunk 2 and restores chunk 1 to:

   ```json
   {"presented":["... four choices ..."],"selected":null}
   ```

4. Run plain:

   ```text
   nexus --json continue --slot 3
   ```

Observed:

```text
success=true
new incubator chunk_id=2
parent_chunk_id=1
user_text=''
```

Chunk 1 still had `selected=null` and empty `choice_text`.

## Root cause

For a committed parent, `continue_narrative()` calls the choice resolver only
when `_has_player_response_input(request)` is true. An empty request therefore
skips resolution even when `narrative_state.choices` is non-empty, then starts
generation with the empty default.

## Acceptance criteria

- If the current committed chunk presents unresolved choices, continue
  requires `choice`, non-empty `user_text`, or `accept_fate`.
- Plain continue returns an actionable 4xx/CLI error without starting
  generation or mutating state.
- Choice-free chunks may retain deliberate empty-input continuation if that is
  still required.
- Regression covers the normal undo → committed-parent state.

---

Codex — GPT-5.6-Sol

PR #619, opened 2026-07-29
Narrative Continue: Validate Before Mutating, Require Choice Resolution

### LINKED ISSUE #614: Rejected narrative continues still persist their explicit model override
## Impact

High for provenance and operator control. A request that fails before
generation can silently change the model used by later turns. The rejected
action appears to have changed nothing, but the next ordinary continue runs a
different provider/model.

## Live reproduction

Starting state:

```text
slot model = gpt-5.6-terra
pending chunk 1 has four choices and no selection
```

Command:

```text
nexus --json continue --slot 3 --choice 99 --model gpt-5.6-sol
```

Result:

```text
exit 1
Choice 99 out of range (1-4)
```

The pending chunk remained unchanged, but:

```text
nexus model --slot 3
Current model: gpt-5.6-sol
```

PostgreSQL independently reported `global_variables.model =
'gpt-5.6-sol'`.

## Root cause

`continue_narrative()` calls `upsert_slot(... model=request.model)` before it
resolves slot state or validates choice/input. Later validation rolls back its
own chunk transaction, but the earlier model transaction is already committed.

## Acceptance criteria

- Validate slot state, input combination, and choice range before persisting a
  request model.
- A rejected continue leaves all canonical state, including model selection,
  unchanged.
- Successful explicit overrides remain persistent by design.
- Tests cover invalid choice, conflicting choice/accept-fate, missing input,
  wizard-mode rejection, and successful generation.

---

Codex — GPT-5.6-Sol

pythagorakase:
## Summary

Fixes #614 and #615 together (night-QA round 2) — both are pre-generation validation-ordering defects in the same `continue_narrative` path.

## Changes

- **#614**: `upsert_slot(model=...)` moved from before any validation to after all synchronous validation (slot state, input combination, choice resolution). A rejected continue now leaves all canonical state unchanged, including model selection. The model-without-slot and choice+accept_fate conflicts are rejected up front.
- **#615**: for a committed parent, choice resolution now runs whenever the chunk presents choices (`require_response=bool(narrative_state.choices)`), not only when input happens to be present. A plain continue against unresolved choices returns an actionable 400; choice-free chunks retain deliberate empty-input continuation.

## Deliberately out of scope

The per-slot generation lease / concurrent-continue race is #616, sequenced after this lands (same entry path).

## Testing

Real FastAPI route via TestClient; `upsert_slot` is spied (never emulated) to assert zero model writes on every rejection path. The undo → committed-parent regression reproduces #615's exact repro against a disposable `NEXUS_template` clone (`NEXUS_RUN_POSTGRES=1`).

- Focused + pg-gated: `6 passed` (includes the disposable-clone regression)
- Full suite: `1955 passed, 451 skipped`

### DIFF
--- a/nexus/api/narrative.py
+++ b/nexus/api/narrative.py
@@ -382,11 +382,16 @@ def _record_player_response_for_chunk(
                         )
                 return existing_choice_text
 
-            if not has_new_input and not require_response:
+            has_unresolved_choices = bool(choice_object and choice_object["presented"])
+            if (
+                not has_new_input
+                and not require_response
+                and not has_unresolved_choices
+            ):
                 return user_text
 
             if not has_new_input:
-                if choice_object and choice_object["presented"]:
+                if has_unresolved_choices:
                     raise HTTPException(
                         status_code=400,
                         detail=(
@@ -641,7 +646,10 @@ async def continue_narrative(
             logger.info(
                 f"Resolved chunk_id={request.chunk_id} from slot {request.slot}"
             )
-    elif request.chunk_id and _has_player_response_input(request):
+    elif request.chunk_id:
+        # Runs even without player input: the resolver owns the
+        # unresolved-choice guard, so an explicitly addressed chunk cannot
+        # bypass it the way slot-resolved continues cannot.
         resolved_user_text = _record_player_response_for_chunk(
             slot=request.slot,
             chunk_id=request.chunk_id,
--- a/nexus/api/narrative.py
+++ b/nexus/api/narrative.py
@@ -386,6 +386,14 @@ def _record_player_response_for_chunk(
                 return user_text
 
             if not has_new_input:
+                if choice_object and choice_object["presented"]:
+                    raise HTTPException(
+                        status_code=400,
+                        detail=(
+                            "Current chunk has unresolved choices; provide choice, "
+                            "non-empty user_text, or accept_fate."
+                        ),
+                    )
                 raise HTTPException(status_code=400, detail="No input provided")
 
             try:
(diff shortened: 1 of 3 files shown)


pythagorakase:
## Sol Review Response

Independent Sol review returned three findings:

**[P1] Explicit slotless `{"chunk_id": N}` bypasses unresolved-choice validation** — **Fixed.** The guard now lives in the resolver itself: its empty-input early return fires only when the chunk presents no choices, and the explicit-chunk path always consults the resolver (input or not). Sol's exact probe is pinned as a regression (`POST {"chunk_id": 17}` against unresolved choices → 400, zero mutations), alongside a twin proving choice-free explicit-chunk empty continuation is preserved. The pre-existing threading test was updated to stub the chunk read the route now legitimately performs.

**[P2] Pending-choice resolution commits before auto-approval can reject** — **Deferred to #616, deliberately.** This is pre-existing behavior in the `has_pending` branch this PR did not modify; the probe itself shows the model write correctly no longer happens (this PR's fix working). The partial state is recoverable — retry takes the existing-choice path. Restructuring resolve → approve → schedule atomicity is exactly the #616 work order's territory (per-slot generation lease + conditional incubator writes), which is sequenced to begin the moment this lands. The #616 order will name this finding explicitly.

**[P3] The choice-free test survives reverting the fix** — **Working as intended.** That test pins acceptance criterion 5 (choice-free chunks retain deliberate empty-input continuation) — preserved behavior, not fixed behavior. A behavior-preservation pin *should* pass on `origin/main`; five of six tests detecting the revert is the correct count. The review brief's "all six" phrasing was the orchestrator's error, not a test gap.

(Sol's full-suite failures were sandbox-environment: blocked loopback binds, Keychain, and PostgreSQL sockets. Unsandboxed runs here: focused 8 passed incl. both new regressions; full suite 1957 passed.)

PR #621, opened 2026-07-29
Retrograde: Validate seek_redemption Prerequisite at R6, Not Persistence

### LINKED ISSUE #612: Retrograde can select `seek_redemption` without its required wary-or-worse relationship, repeatedly blocking wizard transition
## Impact

High. A structurally valid R4/R5/R6 result can survive every model retry and
reach the atomic persistence transaction even though its project is
impossible to start. Persistence rolls back correctly, but the user pays the
full multi-call latency and remains stuck in wizard-ready. Retrying regenerated
the same invalid project family independently.

## Live reproduction

Two consecutive Terra transition attempts failed after independent seed and
expansion calls:

```text
Retrograde project seed 'seed_01' seek_redemption requires a
TARGET->ACTOR wary-or-worse relationship

Retrograde project seed 'seed_01_board_remorse' seek_redemption requires a
TARGET->ACTOR wary-or-worse relationship
```

Both attempts:

- generated and selected candidates;
- generated an R6 expansion;
- began the atomic mapper transaction;
- created the protagonist/location/trait plan in-transaction;
- failed only inside Retrograde project persistence;
- rolled back all canonical writes.

A separate non-mutating candidate command exhausted its structured retries on
another invalid project-intent combination (`build_venture` with a forbidden
`target_ref`), showing that project-intent legality is a live model pressure,
not a synthetic fixture.

## Root cause

`retrograde_expansion._project_plan_issues()` checks project type/target
fidelity, dead participants, and thread status, but not the
`seek_redemption` relationship precondition. The first authoritative check is
`retrograde_persistence._validate_project_start_dependencies()`, after the
repairable R6 structured-output boundary has already passed.

## Acceptance criteria

- R6 validation proves every `seek_redemption` plan has a matching
  TARGET→ACTOR wary-or-worse relationship in existing state or the same
  expansion.
- A missing prerequisite is returned through the bounded structured repair
  path before any mapper transaction begins.
- Retry exhaustion names the invalid seed/project without performing partial
  transition work.
- A live regression covers the exact relationship direction and valence
  threshold.

---

Codex — GPT-5.6-Sol

pythagorakase:
## Summary

Fixes #612 (night-QA round 2). The `seek_redemption` TARGET→ACTOR wary-or-worse precondition was first checked in `retrograde_persistence._validate_project_start_dependencies()` — after the repairable R6 structured-output boundary — so impossible projects survived every model retry, burned the full multi-call latency, and rolled back only inside the mapper transaction.

## Changes

- New `retrograde_project_dependencies` module: the single shared rule (`seek_redemption_dependency_issue`), the canonical-ladder valence threshold (`is_wary_or_worse`, derived from `EmotionalValence` — no literal rungs), and the DB loader for qualifying relationships.
- R6 expansion validation enforces the rule against packet-carried existing relationships plus the same expansion's planned relationship rows; failures return through the bounded structured repair path before any transaction.
- The hard-validation doctrine tells the model the rule up front (prevention before repair).
- Maturation threads existing inbound wary-or-worse relationships into runtime packets; wizard packets declare a clean slate.
- Persistence's transactional recheck now calls the same shared rule (race-proofing retained, drift impossible).

## Testing

Structured-boundary tests: missing, reversed-direction, boundary-valence, same-expansion-satisfied, and retry-exhaustion. PostgreSQL regression drives the checked-in fixture cache → production build → expansion validation → atomic persistence (`NEXUS_RUN_POSTGRES=1`).

- Orrery pg-gated: `1321 passed, 5 skipped`
- Full suite: `1957 passed, 451 skipped`
- Targeted mypy: clean

### DIFF
--- a/nexus/agents/orrery/retrograde_expansion.py
+++ b/nexus/agents/orrery/retrograde_expansion.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import json
-from typing import Annotated, Any, Literal, Mapping, Optional, cast
+from typing import Annotated, Any, Literal, Mapping, Optional, Sequence, cast
 
 from pydantic import BaseModel, ConfigDict, Field, model_validator
 
@@ -13,6 +13,11 @@ from nexus.agents.orrery.retrograde_junctions import resolve_junctions
 from nexus.agents.orrery.retrograde_packet import (
     CORE_ENTITIES_HEADING,
     NAMED_SEED_NPCS_HEADING,
+)
+from nexus.agents.orrery.retrograde_project_dependencies import (
+    coerce_project_start_relationships,
+    planned_project_start_relationships,
+    seek_redemption_dependency_issue,
 )
 from nexus.agents.orrery.retrograde_seed_candidates import (
     RetrogradeProjectType,
@@ -642,6 +647,9 @@ def render_expansion_prompt(
         "selected_candidates": selected_candidates,
         "selected_junctions": selected_junctions,
         "trait_constraints": seed_generation_request.get("trait_constraints", []),
+        "project_start_relationships": coerce_project_start_relationships(
+            packet.get("project_start_relationships")
+        ),
         "protagonist_identity": seed_generation_request.get("protagonist_identity", {}),
         "core_prompt_sections": (
             seed_generation_request.get("prompt_sections", [])[:4]
@@ -731,6 +739,9 @@ def validate_expansion_plan(
         candidates_by_id={
             candidate.seed_id: candidate for candidate in candidates.candidates
         },
+        project_start_relationships=coerce_project_start_relationships(
+            packet.get("project_start_relationships")
+        ),
         forbidden_relationship_traits=forbidden_relationship_traits,
         forbidden_pair_tag_traits=forbidden_pair_tag_traits,
         protagonist_identity=_packet_protagonist_identity(packet),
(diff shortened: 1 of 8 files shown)


pythagorakase:
## Sol Review Response

Independent Sol review returned three findings:

**[P1] Persistence credits planned relationships that cannot materialize** — **Fixed.** The probe scenario (existing friendly TARGET→ACTOR row + planned enemy edge for the same pair → `already_present` → prerequisite falsely satisfied) is closed: dry-run dependency facts now derive from the dry-run *outcomes* — `already_present` rows contribute their **stored** valence, `would_insert` rows their type default — and the execute path validates against transaction-visible database state only (relationship rows are materialized before `_validate_project_start_dependencies` runs). The exact scenario is pinned as a pg regression.

**[P2] Invalid relationship types bypass the repair boundary** — **Fixed.** Unclassifiable model-authored types encountered during dependency-issue collection become seed/row-specific contract issues (naming the legal vocabulary) that ride the structured repair path. A regression proves `invented_hostility` reaches `ModelRetry` rather than escaping as a raw `ValueError`. The vocabulary's fail-loud contract for *authored* vocabulary completeness is untouched.

**[P2] Expansion and persistence consume unversioned snapshots from separate transactions** — **Accepted as designed, not changed.** Expansion validates the packet snapshot at the repairable boundary; persistence revalidates transactionally and is the authoritative last word. The residual window exists only under concurrent relationship mutation mid-wizard, where a late loud failure is the correct outcome; cross-call relationship locking would couple the wizard's LLM calls to DB locks and is out of scope. This is the same two-stage defense the rest of Retrograde persistence uses.

(Sol's pg-gated and full-suite failures were sandbox-blocked sockets/Keychain; unsandboxed runs here: Orrery pg-gated `1323 passed`, full `1958 passed`. Sol's valence probe independently confirmed the ladder comparison and the direction audit confirmed both loaders match the original join.)

Issue #622, opened 2026-07-29
Dev-Endpoint Template-State Test Red Against Live save_05: Faction-Bound Stack Lacks a Target Binding


pythagorakase:
## Symptom

`NEXUS_RUN_POSTGRES=1 pytest tests/test_api/test_orrery_dev_endpoints.py::test_resolve_four_template_states_are_distinguishable` fails on current `main` (`a403b12b`):

```
for stack in group["two_party_stacks"] + group["scene_pressure_stacks"]:
    ...
    assert stack["bindings"].get("target") is not None
AssertionError: bindings == {'actor': 24, 'faction': 25}
```

A two-party/scene-pressure stack resolved against live `save_05` binds `actor` and `faction` but no `target`. The test asserts every such stack carries a `target` binding.

## Context

- Discovered 2026-07-29 during unrelated PR verification (#620 lane); fails identically on a clean `main` checkout, so it is not caused by any in-flight branch.
- The test reads **live slot 5 data**, so this is either (a) slot-5 data drift into a shape the resolver legitimately produces but the test doesn't allow — meaning the test's invariant is too strict — or (b) a real resolver defect where a faction-bound two-party stack drops its target.
- Adjacent contract: the stack-split trap guard and template arity rules in the same test still pass.

## Acceptance

- Determine whether a faction-bound stack without a `target` is a legal resolver output. If legal: relax the test to the true invariant. If not: fix the resolver and keep the assertion.
- The test should not silently depend on mutable live-slot state for its pass/fail boundary; consider pinning the scenario in a disposable clone if the invariant is worth keeping strict.

PR #623, opened 2026-07-29
Narrative: Durable Per-Slot Generation Lease and Truthful Sessions

### LINKED ISSUE #616: Concurrent continues run duplicate turn pipelines and overwrite the singleton incubator

pythagorakase:
## Summary

Fixes #616 (night-QA round 2, the last of the crop) plus the atomicity finding carried from #619's independent review.

## Changes

- **Migration 098**: `narrative_generation_sessions` (durable status), `narrative_generation_lease` (one-row boolean-PK singleton per slot DB), `narrative_parent_embedding_claims` (single-fire guard).
- **`narrative_lease.py`**: acquisition under a short table lock (closes the empty-table race), stale-owner replacement with error-marking, ownership-checked parent binding, owner-scoped release — all loud on lost ownership.
- **Route integration**: continue/regenerate/bootstrap acquire the lease after validation, 409 with `active_session_id` on conflict (documented response model), abandonment on pre-scheduling failure.
- **`write_to_incubator`**: locks the lease row `FOR UPDATE`, verifies session+parent+expiry before replacing the singleton — a stale pipeline can no longer delete the current owner's result.
- **Carried #619 [P2]**: pending-choice resolution + auto-approval share one transaction; approval failure rolls back the choice write.
- Stale-lease timeout configurable in `nexus.toml`.

## Testing

Concurrency regression on a real disposable clone with two genuine route calls — only the frontier call is doubled (blocking stub holds pipeline A mid-generation while B arrives): one owner, one 409 with the active session id, one incubator row bound to the winner, embedding claimed once. Status-truthfulness, rollback, and retry coverage alongside.

- API pg-gated: `219 passed` (1 pre-existing failure = #622)
- Full suite: `1966 passed, 454 skipped`
- flake8 on touched files: 53 violations vs 57 on `main` (net −4, none introduced)

Migration 098 will be applied fleet-wide + template after merge, per the standing workflow.

### DIFF
--- a/migrations/098_narrative_generation_lease.sql
+++ b/migrations/098_narrative_generation_lease.sql
@@ -1,0 +1,41 @@
+-- Serialize narrative generation per slot database and retain durable session truth.
+
+CREATE TABLE IF NOT EXISTS narrative_generation_sessions (
+    session_id UUID PRIMARY KEY,
+    operation TEXT NOT NULL
+        CHECK (operation IN ('continue', 'regenerate')),
+    parent_chunk_id BIGINT,
+    status TEXT NOT NULL
+        CHECK (status IN ('initiated', 'complete', 'error')),
+    chunk_id BIGINT,
+    error TEXT,
+    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
+);
+
+CREATE TABLE IF NOT EXISTS narrative_generation_lease (
+    id BOOLEAN PRIMARY KEY DEFAULT TRUE CHECK (id),
+    session_id UUID NOT NULL UNIQUE
+        REFERENCES narrative_generation_sessions(session_id) ON DELETE CASCADE,
+    parent_chunk_id BIGINT,
+    operation TEXT NOT NULL
+        CHECK (operation IN ('continue', 'regenerate')),
+    acquired_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    expires_at TIMESTAMPTZ NOT NULL
+);
+
+CREATE TABLE IF NOT EXISTS narrative_parent_embedding_claims (
+    parent_chunk_id BIGINT PRIMARY KEY,
+    session_id UUID NOT NULL
+        REFERENCES narrative_generation_sessions(session_id) ON DELETE RESTRICT,
+    claimed_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
+);
+
+COMMENT ON TABLE narrative_generation_lease IS
+    'Per-slot singleton lease preventing concurrent narrative turn pipelines';
+
+COMMENT ON TABLE narrative_generation_sessions IS
+    'Durable generation status whose complete rows must remain bound to incubator';
+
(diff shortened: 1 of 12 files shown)


pythagorakase:
## Sol Review Response

Independent Sol review returned four findings; all fixed:

**[P2] ABBA deadlock between stale takeover and completion** — **Fixed.** Every lease-module transaction now follows lease → claims → sessions order (stated as the module-docstring invariant); `_finish_generation` deletes the lease row *first* and derives ownership from the RETURNING result. Bonus tripwires added while restructuring: completing without a held lease raises, and a committed `complete` refuses downgrade to `error`.

**[P2] Orphaned embedding claims** — **Fixed.** Claiming requires current lease ownership (`FOR UPDATE` on the lease row), the upsert replaces an incumbent claim only when its owning session has terminal `error` status (crash-heal without live-steal), and pre-scheduling abandonment releases the claim alongside the lease. Both properties pg-pinned: an errored owner's claim is reclaimable; a live owner's is not.

**[P2] Notification failure rewrites success as error** — **Fixed.** Both terminal broadcasts are isolated: a raising `send_progress` after the durable commit logs loudly and leaves `complete` standing; the status writer independently refuses complete→error as defense in depth. Pinned by a regression with a raising WebSocket manager.

**[P3] Silent session-mismatch fallback** — **Fixed.** An incubator row held by a different session now raises 409 naming both session ids instead of falling through to committed chunks and a confusing 404.

(Sol's sandbox blocked pg sockets and Keychain, so its pg probes were static; unsandboxed here: the pg regression file runs `14 passed` including the four new pins, full suite `1966 passed`. Migration 098 goes fleet-wide + template after merge per the standing workflow.)

Issue #617, opened 2026-07-29
Conspiracy Channel: Writer Intent Field and Shared Cross-Turn Scratchpad for the Two-Pass Storyteller


pythagorakase:
## Problem

Inter-seat communication in the two-pass storyteller is asymmetric:

- **Gaia → writer**: rich. Gaia's `updates` land in durable state and reach the writer through the next turn's context.
- **Writer → Gaia**: inference-only. `_format_gaia_user_prompt` (`nexus/agents/lore/logon_utility.py`) hands Gaia the writer's verbatim prose/choices/scene/presence, and Gaia must reverse-engineer durable state intent from narrative text. `SkaldWriterWire` has no intent field — a secondary intent channel was discussed during the seat split but never implemented.
- **No cross-turn authorial store exists.** Any plan the writer conceives that is neither uttered in prose nor committed as mechanical state dies at the turn boundary. The seats have no way to record "the innkeeper is the baron's informant — do not reveal yet."

## Proposal (owner, 2026-07-29)

The writer pass gains three text outputs:

1. **The prose** (exists today).
2. **Intended state changes** incident to the active scene — advisory input to Gaia's pass (e.g. "relationship X → Y decrements moderately", "X's emotional state becomes seething").
3. **A shared scratchpad** — the conspiracy channel: persistent, player-invisible co-DM notes both seats read and write across turns (planned reveals, loaded Chekhov's guns, seeded betrayals).

Whether fields 2 and 3 are one mechanism or two is part of the design question.

## Constraints

- Recurring token cost is first-class: the wire schema was recently slimmed 13.5k → 4.3k tokens; every prompt-threaded byte recurs each turn.
- Turn rejection/rollback (the one-chunk incubator) must cleanly revert scratchpad mutations; replay-ledger scope must be decided, not ignored.
- The scratchpad must never leak to the player: not in UI payloads, not in `narrative_chunks.raw_text`, not in embeddings.
- Must work on both grammar-enforced (OpenAI strict) and schema-free (OpenRouter passthrough + repair loop) transports.
- Tunables in `nexus.toml`; prompt text in `prompts/*.md`.

## Process

Design shape goes to an Arachne ruling (brief to follow). Implementation after the ruling lands.

pythagorakase:
## Arachne Ruling (seq 9, 2026-07-29T05:52:58Z) — Verbatim

[...]

PR #624, opened 2026-07-29
The Conspiracy Channel: Private DM Correspondence Between the Seats

pythagorakase:
## Summary

Implements #617 per the Arachne ruling (seq 9 + hysteresis and units amendments): **Immutable DM Correspondence** between the writer and Gaia seats — the private channel the co-DMs never had.

## Design as Ruled

- One letter per seat per turn, immutable, append-only, keyed by chunk and seat (DB triggers reject UPDATE on letters and digests).
- Both seats see the digest plus the trailing window verbatim; hysteresis in **turns**: floor 5 / ceiling 10 (`nexus.toml`), compaction folds the oldest exchanges in a post-accept utility call that judges each plan live / stale / retired.
- Empty first page; the persona documents make the seats *want* the correspondence (orchestrator-authored prose, wired verbatim by the implementer).

## Integrity Properties

- Letters stage in incubator columns and journal only on acceptance, inside the chunk transaction; rejection/regeneration discards them; undo = FK cascade (digest versions restore automatically).
- Privacy proven by test: never in `raw_text`, public response dumps, embeddings input, or logs.
- Deliberately excluded from the world-state replay ledger (authorial intent, not canon) — the exclusion, chunk provenance, orphan, and exchange-atomicity invariants are checked by `replay_state.py --verify`.
- Compaction's model call runs outside the DB transaction; the plan re-reads accepted state before inserting and fails loudly on concurrent drift.

## Testing

- Full suite: `1986 passed`. Pg regressions (acceptance/rejection/undo/digest/migration/replay) green on disposable clones; the sole pg failure is pre-existing #622.
- **Live gate run at review** (`NEXUS_CONSPIRACY_E2E=1`): real two-seat exchange produced letters from both seats, public payload clean, and a real compaction call folded exactly the oldest five exchanges (`compacted_through_chunk_id` at the precise hysteresis boundary). 33s.
- Two review-round fixes applied by the orchestrator: the live fixture now registers its disposable clone in `VALID_DBNAMES`, and the live test hands compaction a free connection (production owns its own transaction scope).


pythagorakase:
## Sol Review Response

Independent Sol review returned five findings:

**[P1] Storyteller context endpoint serves the correspondence** — **Fixed structurally.** Private keys are recursively omitted at session-context *storage* (append and regeneration paths), so `GET /api/story/context/{session_id}` and every sibling consumer of stored context cannot serve what was never stored. New endpoint leakage tests cover context, history, state, and regeneration surfaces, plus a progress-storage assertion.

**[P1] Transitive digest contamination after non-suffix undo** — **Rebutted with evidence; hardened by documentation.** No production path deletes `narrative_chunks` rows: accepted chunks are append-only ironman state, and rejection touches only the incubator. Under administrative suffix rewinds, every later digest cascades away with its accepting chunk, so no surviving digest can cover a deleted chunk. Migration 099's header now states this assumption and that non-suffix surgery must rebuild digests.

**[P2] Refusal-only render cap can brick a slot** — **Fixed by construction.** Letter (300 tok) and digest (2,000 tok) budgets are enforced at the repairable structured boundary for writer, Gaia, single-pass, and compaction outputs; settings validation proves worst case (ceiling × 2 × letter + digest) stays under the 12,000 render cap, which is demoted to a loud should-never-fire invariant.

**[P2] Silent config defaults** — **Fixed.** `[storyteller.correspondence]` is mandatory; loading a config without it raises naming the section. No silent model-ref default.

**[P3] pg acceptance test bypasses the production commit path** — **Fixed** (fourth documented fixture-trap occurrence). The regression now stages via incubator columns and accepts via the production commit handler; reverting the journaling wiring turns it red.

Verified unsandboxed: full suite `1990 passed`; pg + privacy trio `16 passed`; **live gate re-run post-fixes: passed** (26s, bounded-repair path in the loop). Sol's own probes confirmed the strict wire at 4,342 tokens (+18 vs the schema-weight baseline) and exact hysteresis behavior at both boundaries.

Issue #633, opened 2026-07-30
Storyteller correspondence compaction calls the synchronous provider inside FastAPI's event loop


pythagorakase:
## Summary

The first live storyteller-correspondence compaction fails from the gateway
because the post-accept path calls the synchronous OpenAI structured-completion
method while FastAPI's event loop is running.

This is also a partial-success consistency failure: the narrative chunk and
11th correspondence exchange are committed and the incubator is cleared before
the endpoint returns HTTP 500.

## Live reproduction

Environment:

- commit `6ea1f5903e7654ac15a28e48feb0af697b622d62`
- fresh runtime-only slot 5 campaign
- gateway on port 8013
- Skald and Gaia both `gpt-5.6-terra`
- `[storyteller.correspondence] floor_turns = 5`
- `[storyteller.correspondence] ceiling_turns = 10`

After ten accepted exchange pairs, accept the pending scene normally:

```console
nexus --json continue --slot 5 --model gpt-5.6-terra --choice 1
```

The commit log reaches:

```text
Created narrative chunk 14
Cleared incubator for session 183f58e7-b16a-4615-ad58-a187a8a08cfa
LOGON initialized with openai provider using model gpt-5.6-terra
Failed to commit narrative: OpenAIProvider.get_structured_completion() cannot be called from a running event loop; use OpenAIProvider.get_structured_completion_async() instead.
POST /api/narrative/continue HTTP/1.1" 500 Internal Server Error
```

Database state immediately after the 500:

```text
narrative chunks:       13 (max id 14; an earlier rollback left an id gap)
accepted letters:       22
accepted exchanges:     11
digest versions:        0
incubator rows:         0
```

No compaction provider usage event was emitted; the guard rejected the sync
call before a request was sent.

## Likely path

`commit_incubator_sync()` calls
`compact_accepted_correspondence_sync()` after its commit. That creates a
`LogonUtility`, whose `compact_correspondence()` calls
`compaction_provider.get_structured_completion(...)`. The gateway request still
owns a running event loop, and the OpenAI provider correctly rejects that sync
entrypoint.

Relevant current locations:

- `nexus/api/commit_handler_sync.py`
  (`compact_accepted_correspondence_sync`)
- `nexus/agents/lore/logon_utility.py`
  (`compact_correspondence`)

## Expected

- Due compaction runs through the async provider path or an explicitly isolated
  worker/thread boundary.
- The endpoint does not report total failure after durably accepting and
  clearing the pending scene.
[...]

PR #636, opened 2026-07-30
Pin every remote model route in the QA lane and rotate probe depth

pythagorakase:
## Summary

Closes the isolated-lane model-route leak the 2026-07-30 scratchpad audit caught live, plus a second leak found while closing it, and rotates the nightly QA shift's probe charter toward the depth band where bugs actually surfaced.

- **Route pinning**: `_write_runtime_config` now pins `wizard.fallback_model` and `[orrery.narration]` (provider + model_ref) into the generated lane config. The audit observed two async Orrery narration jobs escaping to the tracked Anthropic route; the wizard fallback had the identical exposure but no fallback happened to fire that night.
- **Drift-guard roster test**: walks `nexus.toml`, collects every model-route key, and requires each to be directly pinned, `@openai.<role>`-indirect (covered by the roles-table rewrite), or explicitly non-remote. A future route added without lane coverage fails the suite at commit time instead of leaking during an unattended shift.
- **Depth rotation** (`mission_prompt.md`): optional seed-based slot initialization from `temp/qa_seeds/` (newest checksummed dump, then `migrate.py`), and a requirement that at least two probe families target deep-state mechanics when seeded. Rationale: the first automated shift hit its dry-well exit on shallow families the same night a deep campaign on the same commit produced #633/#634.
- **Allowance correction**: `[usage.daily_allowance] openai` 2.5M → 10M (owner is tier 4; the old figure was the tiers-1–2 number). Readout-only, never enforcement.

## Testing

- `tests/test_qa_shift.py`: 12 passed (new: narration shape-error case, route-roster drift guard; extended: lane-pin assertions).
- Full suite in the worktree: 2027 passed, 463 skipped.
- mypy: no new error classes (the file's 12 pre-existing tomlkit `Item | Container` errors remain the established idiom).

### DIFF
--- a/nexus.toml
+++ b/nexus.toml
@@ -1148,7 +1148,10 @@ usage_dir = ".nexus/runtime/usage"  # relative paths resolve against the repo ro
 [usage.daily_allowance]
 # Per-provider daily API-token allowances (UTC day), used for readout only —
 # measurement, never enforcement. Providers without an entry show no allowance.
-openai = 2500000
+# Owner is OpenAI usage tier 4: 10M/day complimentary for the Terra-class
+# model group (Sol/o3-class premium models have a separate 1M/24h pool this
+# readout does not track).
+openai = 10000000
 
 # =============================================================================
 # Managed Runtime (issue #396) — nexus up / down / restart / status / logs
--- a/scripts/qa_shift/README.md
+++ b/scripts/qa_shift/README.md
@@ -46,6 +46,17 @@ discovery mechanism. Reconfirm the organization’s current complimentary-token
 enrollment and eligible model group before raising it or changing
 `target_model`.
 
+## Seeds
+
+Shallow single-request probes stop finding bugs once the early-game surface
+hardens; state-threshold defects (compaction, alias accumulation) only appear
+deep into a campaign. Drop a checksummed mid-campaign `pg_dump -Fc` dump into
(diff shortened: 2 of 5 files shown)


pythagorakase:
Independent Sol review returned CHANGES_REQUIRED with three findings; all verified against code/schema and addressed in the fix commit:

- **P1 (roster misses bare `model` keys)** — CONFIRMED: `apex.model` and `ir_eval.judgment.model` exist and were invisible to the drift guard (`*_model`/`model_ref` matching missed the bare spelling). Fixed: `"model"` added to `MODEL_ROUTE_KEYS`; both routes now classify as role-indirect and are pinned via the roles rewrite.
- **P2 (any `@openai.*` accepted while only default/gaia rewritten)** — CONFIRMED as a latent hole (the roles table holds exactly default+gaia today). Fixed both sides: `_write_runtime_config` now rewrites *every* role in the openai roles table, and the roster test additionally requires each role-indirect ref to name a role present in that table.
- **P2 (seeding carries source-lane state)** — CONFIRMED and slightly sharper than stated: `global_variables` carries `model` *and* `slot_number`, plus `orrery_narration_jobs`/`orrery_maturation_jobs` rows survive the restore. Fixed: mission prompt's seeding step now restamps both variables and clears restored job queues before gateway start, with statements and row counts recorded.

12/12 kit tests pass post-fix.

PR #641, opened 2026-07-30
Track the Scratchpad Audit Kit with Anchored Rubric v2.1

pythagorakase:
## Summary

Promotes the correspondence-audit instrument from run-archive scratch into the tracked repo at `scripts/scratchpad_audit/`, alongside the QA shift kit.

- **`rubric.md` (v2.1)**: six dimensions, 0–4, every rung anchored with verbatim corpus quotes; digest-fidelity appendix aligned with the compaction-audit charter language; explicit cross-version score mapping (v1 ≈ v2 2 → v2.1 2–3, v2 3 → v2.1 4).
- **`README.md`**: audit discipline (runtime-artifacts-only, route pinning, blind scoring caveats, revision gate) and the `temp/qa_seeds/` handoff to the nightly shift.

## Validation Basis

- v2 was blind re-scored against the 24-exchange Vesper Station corpus: ceiling test passed (≤21% at top score, clustering where the correspondence demonstrably works hardest), but 71–88% of exchanges pinned at the single value 2 — no mid-band discrimination.
- v2.1 splits the catch-all and codifies all eight ambiguities the blind scorer reported having to resolve by invented convention (on-path reference frame, first-exchange N/A, Engagement boundary rule, deferred-uptake credit, shared-evidence rule, dropped-thread findings, escape-clause ceiling, per-rung anchors).
- First live use: the next audit's fresh corpus (slot 5 now holds 36 exchanges + 3 digests), which also removes the anchor-contamination caveat inherent to same-corpus validation.

### DIFF
--- a/scripts/scratchpad_audit/README.md
+++ b/scripts/scratchpad_audit/README.md
@@ -1,0 +1,36 @@
+# Scratchpad Audit Kit
+
+Instruments for commissioned live audits of the Skald–Gaia correspondence
+channel (the private writer/Gaia letters and their digest compactions).
+
+An audit is a one-off adversarially-steered campaign run against an isolated
+gateway lane, driven by a frozen charter handed to a Codex operator. Charters
+are authored per run (they encode the specific mission — behavioral audit,
+compaction proof, regression check) and archived with the evidence; this kit
+holds what persists between runs.
+
+## Contents
+
+- `rubric.md` — the anchored scoring instrument (v2.1, 0–4 per dimension,
+  six dimensions plus the digest-fidelity appendix). Version lineage and
+  cross-run score mapping are documented in the rubric header.
+
+## Audit Discipline
+
+- Runtime artifacts only: an audit never edits tracked files, commits, or
+  writes to the database outside public surfaces. Evidence archives live in
+  ignored `temp/` (`temp/scratchpad_audit_*` / `temp/compaction_audit_*`).
+- Model routes are pinned through an ignored runtime config derived from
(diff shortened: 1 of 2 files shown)


pythagorakase:
Independent Sol review returned CHANGES_REQUIRED: 4 P1 (anchor fidelity) + 6 P2 (rung/gate design). All ten verified and addressed:

**P1s — all confirmed.** Root cause: several anchors were built from v1 score notes or report excerpts rather than verified letter text, and two sat on exchanges whose attribution my own two scoring passes dispute (chunk 26 vs 27 for the hall-isolation declaration; chunk 21 vs 22 for the receipt split). Fix: contested anchors dropped rather than adjudicated — Engagement 4 anchors on chunk 13 alone, Chemistry 4 on chunk 7 (credited at the giving exchange per Rule 6, consistent with its own rule), Chemistry 1 on chunk 18 described without quotation, and the chunk-8 note-quote demoted to paraphrase.

**P2s — all confirmed as scoreability holes.** Undefined mean-mapping dropped (rung-level comparison only); Adaptability 2 redefined as the steady-state default so a clean on-path exchange satisfies a rung; bare declines assigned to Engagement 2 with refusal-with-substitute at 4; Chemistry 3 restricted to gifts *uncashed within the corpus* so 3/4 are exclusive; unprobed digests gradeable on content review with an explicit (unprobed) mark; the revision gate's thresholds made numeric (no dimension >60% at any single value; prior bottom-20% weak set must land in the bottom third).

PR #645, opened 2026-07-30
Anchor Need Clocks to the Story Clock, Never Wall Time

### LINKED ISSUE #640: New-story need clocks can anchor to wall time and overflow far-future fulfillment
## Impact

A public narrative acceptance can return HTTP 500 and roll back when a
far-future story has `character_need_states.last_evaluated_at` anchored to
wall time. Orrery derives more than one million hours of need debt, then a
normal fulfillment tries to persist the remainder into `numeric(8,2)`.

This blocked the first acceptance in the slot-5 compaction regression audit.
The narrative, correspondence, digest, and incubator transaction rolled back
cleanly, but the pending turn could not be accepted until a public
regeneration explicitly deferred every affected need-fulfillment proposal.

## Live public-surface reproduction

Repository: `main` at `64a476b7`  
Slot: fresh Vesper Station campaign in `save_05`  
Canonical base timestamp: `2189-10-17 15:12:00-04`

The pending turn contained ordinary need resolutions, including:

```json
{
  "template_id": "drink",
  "binding_names": {"actor": "Elian Rook"},
  "state_delta": {
    "need.fulfill": {
      "type": "thirst",
      "quality": "routine",
      "discharge_debt": 9999
    }
  }
}
```

Accepting it through the public CLI:

```text
poetry run nexus --json continue --slot 5 --model gpt-5.6-terra \
  --user-text "<player response>"
```

returned:

```text
API error: {"detail":"Failed to commit narrative: numeric field overflow
DETAIL:  A field with precision 8, scale 2 must round to an absolute value less than 10^6.
"}
```

The gateway log confirms that the commit created the provisional narrative and
applied ordinary entity updates, then failed during Orrery work. PostgreSQL
rolled the transaction back: narrative/correspondence counts and the pending
incubator row were unchanged.

## PostgreSQL confirmation

The canonical story clock is in 2189, but the initial active characters'
need-state clocks were created at wall time:

```text
character          need       debt_score  last_evaluated_at
Mara Vey           sleep      0.00        2026-07-30 00:55:20-04
Ivo Senn            sleep      0.00        2026-07-30 00:55:20-04
Elian Rook          thirst     0.00        2026-07-30 00:55:20-04
Niko Rell           sleep      0.00        2026-07-30 00:55:20-04
Ren Vale            sleep      0.00        2026-07-30 00:55:20-04
```

The resolver therefore surfaced critical need pressure at `1430751.8`.
Applying even the `9999` discharge leaves about 1.42 million, which cannot fit
the table's `numeric(8,2)` `debt_score`.

A public regeneration under a new session reproduced the same proposal IDs,
need fulfillments, and `1430751.8` pressures. A second regeneration recovered
only because its authorial note caused Skald to persist `defer`
adjudications for all affected fulfillments.

## Likely root cause

`orrery_sync_character_need_states()` initializes new rows with:

```sql
SELECT COALESCE(MAX(world_time), now())
INTO anchor_world_time
FROM chunk_metadata;
```

The character trigger can run before the first canonical `chunk_metadata`
world time exists, so it stores `now()`. Once the far-future base timestamp
and first chunk arrive, `ON CONFLICT ... DO NOTHING` leaves those rows anchored
163 years behind the story.

## Expected

- New-story need clocks anchor to the accepted canonical base/first world time,
  including when characters are created before the first chunk.
- Existing wall-time/canonical-time mismatches have a deterministic
  reconciliation or migration path.
- A normal fulfillment cannot reach PostgreSQL with a value outside the
  storage domain.
- Replay uses the same repair/reconciliation authority.

## Dedup

Searched every existing issue title and body. #600 and #613 cover canonical
seed timestamp coherence, #437 covers the removed accrual-floor policy, and
#540 covers fulfillment metadata replay. None covers pre-world-clock
need-state initialization or the resulting numeric overflow.

Codex — GPT-5.6-Sol

pythagorakase:
Closes #640.

## What

- **Migration 100** redefines `orrery_sync_character_need_states` (the only live function carrying the wall-time fallback — confirmed via `pg_proc.prosrc`): anchor chain is now `MAX(world_time)` → `global_variables.base_timestamp` → a named invariant exception. Never `now()`. Same migration deterministically restamps wall-time-poisoned `last_evaluated_at`/`last_fulfilled_at` rows and reports counts.
- **Wizard ordering fix**: the atomic transition wrote `base_timestamp` *after* `create_protagonist`, so the need-state trigger fired before the world clock existed (and reused slots exposed the prior story's clock). The clock write now precedes protagonist creation in the same transaction, with a comment stating the invariant.
- **Named domain guard**: out-of-domain `debt_score` writes now raise `NeedDebtScoreDomainError` naming character, need, and value — replacing the inscrutable `numeric field overflow` HTTP 500 that blocked the slot-5 compaction audit's first acceptance. No clamping; the failure stays loud.

## Why

Two-clocks doctrine: diegetic state lives on the world clock. A need clock seeded at wall time in a 2189 story derived 1.43M hours of phantom debt and 500'd a normal acceptance (full reproduction in #640).

## Testing

- 95 passed on real template clones (`qa640_*`, dropped after): anchoring with empty `chunk_metadata`, missing-clock invariant failure, reconciliation counts + no-op case, large-gap fulfillment in domain, domain-guard raise, and a stale-clock wizard regression asserting need rows get the NEW story's timestamp.
- Migration runner applies and stamps cleanly; mypy/black clean on touched files.

Post-merge: orchestrator applies migration 100 fleet-wide (slots + template).

### DIFF
--- a/docs/orrery_packages.md
+++ b/docs/orrery_packages.md
@@ -3720,6 +3720,7 @@ the seeding migrations to confirm catalog ↔ schema agreement:
 - `migrations/059_orrery_social_travel_event.py`
 - `migrations/063_orrery_adjudication_history.sql`
 - `migrations/067_rename_orrery_templates.sql`
+- `migrations/100_orrery_need_clock_anchor.sql`
 
 ### Tags queried as durable (via `has_tag` / `lacks_tag` / `has_any_tag`)
 
--- a/migrations/100_orrery_need_clock_anchor.sql
+++ b/migrations/100_orrery_need_clock_anchor.sql
@@ -1,0 +1,134 @@
+-- Anchor Orrery need clocks to canonical story time and repair wall-time rows.
+
+CREATE OR REPLACE FUNCTION orrery_sync_character_need_states(
+    p_character_entity_id bigint
+)
+RETURNS integer AS $$
+DECLARE
+    active_tags text[];
+    anchor_world_time timestamptz;
+    affected integer := 0;
+    row_count integer := 0;
+BEGIN
+    IF p_character_entity_id IS NULL THEN
+        RETURN 0;
+    END IF;
+
+    IF NOT EXISTS (
+        SELECT 1
+        FROM entities e
+        WHERE e.id = p_character_entity_id
+          AND e.kind = 'character'
+          AND e.is_active = true
+    ) THEN
+        RETURN 0;
+    END IF;
+
+    SELECT orrery_active_character_tag_names(p_character_entity_id)
+    INTO active_tags;
+
+    anchor_world_time := COALESCE(
+        (SELECT MAX(world_time) FROM chunk_metadata),
+        (SELECT base_timestamp FROM global_variables)
+    );
+    IF anchor_world_time IS NULL THEN
+        RAISE EXCEPTION
+            'need-clock anchor unavailable: no canonical world time or base_timestamp';
+    END IF;
+
+    INSERT INTO character_need_states (
+        character_entity_id,
+        need_type,
+        debt_score,
+        last_evaluated_at,
+        metadata
+    )
+    SELECT p_character_entity_id,
+           needs.need_type::character_need_type,
+           0,
+           anchor_world_time,
+           '{"synced_by": "need_applicability"}'::jsonb
+    FROM (
+        VALUES
+            ('sleep'),
+            ('hunger'),
+            ('thirst'),
+            ('socialize'),
+            ('intimacy')
+    ) AS needs(need_type)
+    WHERE orrery_need_applies_to_tags(
+        needs.need_type::character_need_type,
+        active_tags
+    )
+    ON CONFLICT (character_entity_id, need_type) DO NOTHING;
+    GET DIAGNOSTICS row_count = ROW_COUNT;
+    affected := affected + row_count;
+
+    DELETE FROM character_need_states cns
+    WHERE cns.character_entity_id = p_character_entity_id
+      AND NOT orrery_need_applies_to_tags(cns.need_type, active_tags);
+    GET DIAGNOSTICS row_count = ROW_COUNT;
+    affected := affected + row_count;
+
+    UPDATE entity_tags et
+    SET cleared_at = now()
+    FROM tags t,
+         (
+            VALUES
+                ('sleep', 'sleep_deprived'),
+                ('hunger', 'hungry'),
+                ('thirst', 'thirsty'),
+                ('socialize', 'under_socialized'),
+                ('intimacy', 'intimacy_starved')
+         ) AS severity(need_type, prefix)
+    WHERE et.entity_id = p_character_entity_id
+      AND et.tag_id = t.id
+      AND et.cleared_at IS NULL
+      AND t.tag LIKE severity.prefix || '\_%'
+      AND NOT orrery_need_applies_to_tags(
+          severity.need_type::character_need_type,
+          active_tags
+      );
+    GET DIAGNOSTICS row_count = ROW_COUNT;
+    affected := affected + row_count;
+
+    RETURN affected;
+END;
+$$ LANGUAGE plpgsql;
+
+COMMENT ON FUNCTION orrery_sync_character_need_states(bigint) IS
+    'Synchronizes applicable character need rows and anchors new clocks to canonical story time; raises when neither chunk world time nor the story base timestamp exists.';
+
+DO $$
+DECLARE
+    canonical_base_timestamp timestamptz;
+    reconciliation_anchor timestamptz;
+    evaluated_row_count integer := 0;
+    fulfilled_row_count integer := 0;
+BEGIN
+    SELECT base_timestamp
+    INTO canonical_base_timestamp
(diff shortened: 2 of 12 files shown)


pythagorakase:
Independent Sol review returned CHANGES_REQUIRED (5 P1, 2 P2, 1 P3). Orchestrator verification and dispositions — fixes in flight on this branch:

- **P1 migration:120 (one-directional reconciliation)** — CONFIRMED by inspection. A past-set story (base 1920, poison 2026) keeps future-dated anchors and freezes. Predicate widened to catch anchors outside the canonical domain in both directions.
- **P1 migration:118 (replay-invisible reconciliation)** — empirical (reviewer reproduced exit 1 with five drifts). Fixer will reproduce and make `replay_state.py --verify` green across the migration boundary using the same reconciliation authority.
- **P1 events.py:5095 (app-layer `now()` fallback)** — CONFIRMED. `_tick_world_time_sync/async` fall back to wall time on NULL `world_time`, which is legitimate for flashback/atemporal chunks — reachable in normal play. Same canonical chain + named error; full sweep of orrery for sibling fallbacks.
- **P1 replay.py:2474 (anchor marked unreproducible)** — CONFIRMED. Migration 100 makes the anchor deterministic; replay will reconstruct and verify it instead of skipping.
- **P1 test_retrograde_retrieval_pg.py:171** — CONFIRMED, ruled a fixture defect: production ordering guarantees clock-before-characters, so a slot with characters and no clock is off-invariant. Fixture gains the clock; the RAISE stands. (Hand-built-fixture trap, fifth documented occurrence.)
- **P2 events.py:2673** (guard after helper INSERT) — reorder validation ahead of the first write in both arms.
- **P2 test fixture requires migration absent** — tests become migration-state-agnostic.
- **P3 mapper:550 stale contract comment** — updated.

Re-review to follow on the fix commit.

pythagorakase:
All eight findings fixed in e97d9b61 — dispositions:

- **P1 reconciliation direction** — predicate now catches both below-base and above-ceiling poison on both clock columns; 1920-story/2026-poison regression added.
- **P1 app-layer `now()`** — `_tick_world_time_sync/async` resolve source chunk → primary `MAX(world_time)` → `base_timestamp` → named `OrreryWorldClockUnavailableError`. Full orrery sweep: two sites changed; every remaining `now()` cleared as non-diegetic (audit/provenance/lease/report metadata), itemized in the fix report.
- **P1 replay skips deterministic anchor** — replay now reconstructs post-migration trigger anchors; unreproducible marking retained only for rows proven to predate migration 100.
- **P1 replay verify across boundary** — replay mirrors migration 100's reconciliation authority at checkpoint boundaries; `replay_state.py --verify` green on a poisoned migrated clone (output in the PR-linked report).
- **P1 pre-clock fixture** — ruled a fixture defect; THREE fixtures fixed (sweep found two more with the same shape), no production RAISE weakened; the intentional missing-clock regression creates under a valid clock then removes it.
- **P2 guard ordering** — validation now precedes the first write in both arms, with sync/async ordering regressions.
- **P2 migration-state coupling** — tests detect and adapt; suite proven green with and without 100 pre-applied.
- **P3 stale contract comment** — updated.

95 + 9 passed, black/mypy clean on touched files, clones dropped. Round-2 targeted re-review commissioned on the fix commit.

pythagorakase:
Round-2 re-review: 3 P1, all in the replay boundary mechanism, all reproduced live (false drifts + phantom debt). Root cause is one design flaw — migration authority inferred from wall-clock provenance (`created_at` ceilings, `schema_migrations.applied_at`), which cannot cleanly partition pre- from post-migration writes.

Orchestrator ruling instead of heuristic patching: **the migration records its own provenance**. The reconciliation UPDATE stamps `reconciled_by: migration_100` + prior values into each touched row's `metadata` jsonb; replay rebases checkpoint *baselines* from the recorded transform (never rows replayed forward inside a window), verifies everything else under the current deterministic rule, and marks the narrow unmarked remainder unreproducible rather than reconstructing legacy helper behavior. The wall-clock inference machinery is deleted. Round-2 fix mission in flight with the reviewer's three failure scenarios as regression tests.

pythagorakase:
Round-3 review: 3 P1, all reproduced, all in replay. They triangulate two remaining design shortcuts, now ruled on:

1. **Provenance moves to an immutable audit table** (`character_need_state_reconciliations`, written by migration 100 in the same transaction, recording prior/new values plus the debt pre-image). Live-row `metadata` markers die with applicability delete/reinsert cycles — the round-3 six-drift case; the audit table is row-lifecycle-independent and becomes replay's sole authority.
2. **The remainder rule generalizes**: any (row, field) the current deterministic rule cannot derive from ledger state is unreproducible — including wall-accrued `debt_score` on reconciled rows (the 729.13→0.00 false drift) and legacy NULL-clock resolutions like atemporal `travel.start` (the fabricated world-time drift). The current helper governs forward replay only; legacy outcomes are never force-derived.

Round-3 fix mission in flight with all three scenarios as reproduce-then-clear regressions, every earlier scenario retained, and the full replay suite required.

PR #643, opened 2026-07-30
Log Every Rejected Structured-Output Attempt with Sanitized Error Text

pythagorakase:
Closes #637.

## What

- Every rejected structured-output attempt (and final exhaustion) now logs a `structured-output rejected` line naming transport, model, seat, attempt, exception class, and the full error text — across all OpenAI repair loops (responses/chat_completions, sync/async) and all three Anthropic transports.
- `orrery_tag_validation` now logs the complete formatted issue list, not just the count.
- New shared helper `structured_output_error_text()` renders Pydantic errors with `include_input=False` so correspondence letter content can never leak into logs (the wire retry prompt is unchanged — the model still sees the full error).

## Why

Forensic mining of the 2026-07-30 audits had to reconstruct rejection causes from retry-prompt token-delta arithmetic because rejected branches logged nothing (67k tokens of repair waste were attributable only to a candidate set). A future repair-tax audit can now classify 100% of rejections from logs alone. Observability only — zero behavior change.

## Testing

- 118 passed across `tests/test_native_structured_output.py` and `tests/test_orrery_tag_validation.py`, including a sentinel-based letter-redaction test proving log lines name the failure without carrying letter text.
- Black clean; mypy: zero new errors (42 pre-existing baseline in the two script files).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_011gEoexyFZe1M6tQDqBsK5p

PR #646, opened 2026-07-30
Registry-Keyed Enum Grammar for Orrery Closed-Vocabulary Fields

### LINKED ISSUE #638: Grammar-Constrain Orrery Closed-Vocabulary Fields to Eliminate the Dominant Repair Tax
## Problem

The 2026-07-30 scratchpad audit spent **306,167 tokens (82% of all repair waste, 12 of 15 rejections) on one failure class**: Gaia-seat responses naming unregistered or entity-kind-incompatible tags, caught *after* decode by `orrery_tag_validation` and repaired at ~25k tokens per retry. The offending fields — `updates.<kind>[].tags_add`/`tags_clear`, `tag_hints`, `pair_tag_hints[].tag`, `replacement_event_type` — are all free `str` on the wire, so the decoder happily emits names the registry will refuse.

Aggravating data from the audit mining:

- Rejections cluster with **Orrery adjudication workload**, not context growth: 2 repairs in the first 13 turns vs. 8 in the last 12, tracking tick logs climbing from `0 resolutions` to `7-8 resolutions` per commit. As campaigns deepen, this class grows.
- **Repair guidance does not land reliably**: 5 of the 12 were *second* rejections; two calls came back with identical issue counts after being told the rule.

## Proposal

Constrain these fields with string enums in the strict decode grammar, making an unregistered name impossible at generation time instead of expensive after it.

**Feasibility constraint (measured)**: the registry holds 465 tags (5,399 total chars) + 109 event types. Naively enumerating the full registry in every tag-bearing field would exceed OpenAI's ~500-enum-value guidance across a schema. Two mitigations compose:

1. **Per-entity-kind scoping** — the validator already rejects kind-incompatible tags, so each `updates.<kind>` arm only needs its kind's subset.
2. **Contextual vocabulary pruning** — the schema-weight campaign already computes scene-relevant vocab subsets (~55% savings); the same machinery can emit per-call enum subsets.

Trade-off to accept: enum-constrained decode cannot say "did you mean" — it removes the choice instead. A tag genuinely missing from the contextual subset becomes unemittable rather than repairable; the subset-construction rules must therefore err generous.

## Acceptance Criteria

- Registry-rejection class eliminated (or reduced to genuine contextual-subset misses) in a live campaign run of comparable Orrery workload.
- Before/after repair rate measured via the #626 usage ledger + logs.
- Land #637 first so the residual rejection classes are named, not inferred.

Evidence: `temp/scratchpad_audit_2026-07-30_044920Z/` (report §"Structured-output repair rate", `primary_repairs.json`, gateway log); registry counts from `NEXUS_template`.

### LINKED ISSUE #642: Generation-Time Tag Validator Skips OrreryReplacementStateDelta Tag Lists
## Problem

`_bestowal_sites` (`nexus/agents/logon/orrery_tag_validation.py:81-122`) walks only `updates.*` and `state_updates.*` tag fields. The five tag lists on `OrreryReplacementStateDelta` (`nexus/agents/logon/apex_schema.py:547-584`) — `entity_tags_add`, `entity_tags_remove`, `entity_tags_target_add`, `entity_tags_target_remove`, `entity_pair_tags_target_clear_inbound` — are never validated at generation time. An unregistered or kind-incompatible tag in a replacement state delta sails through the `ModelRetry` validator and fails at commit instead, where the error is a transaction rollback rather than a repairable retry.

Found during #638 recon (2026-07-30) while mapping every free-string closed-vocabulary field on the Gaia wire.

## Fix

Extend the generation-time walk to cover all five lists, reusing the same `tag_names_by_kind` / pair-tag partition the existing sites use. Being folded into the #638 PR (same code region, shared partition logic).

## Acceptance Criteria

- A replacement state delta carrying an unregistered tag is rejected at generation time with a named issue in the `ModelRetry` body, not at commit.
- Test coverage for at least one list per target arm (entity vs target vs pair-inbound).

pythagorakase:
Closes #638. Closes #642.

## What

- **Request-time Gaia schema subclasses** (`nexus/agents/logon/gaia_registry_schema.py`) inject per-kind string enums — character/place/faction tag names, pair-tag names, event types — into the OpenAI strict decode grammar via `create_model` subclassing (the proven Retrograde seed-path pattern), shared through `$defs` so each vocabulary is emitted once. Unregistered and kind-incompatible names become unemittable at generation time instead of costing ~25k-token repairs after decode.
- **Registry-static and digest-keyed**: the enum set mirrors the validator's own partition (`read_storyteller_vocabulary`), built once per registry digest — no per-turn schema rebuilds. Partition-equivalence tests pin grammar == validator, so they can never disagree.
- **`new_entities` per-kind declaration union** gives `tag_hints` kind-scoped enums while boundary coercion preserves the exact static app contract.
- **#642 closed**: the five `OrreryReplacementStateDelta` tag lists are now validated at generation time (proposal endpoint kinds), converting commit-crashes into repairable retries.
- **Config gate** `apex.tag_library.schema_enums` (nexus.toml); off = byte-identical static schema. OpenAI strict arm only — lenient/prompted/Anthropic arms unchanged.

## Measurements

- OpenAI limits verified from current docs: 1,000 total enum values; >250-value enums capped at 15,000 chars. This schema: **617 values**, largest enum 309 values / 3,536 chars — event types included.
- Gaia strict schema: 3,177 → 5,982 o200k tokens (budget test at 6,600). The ~2.8k/call cost buys elimination of the measured 82% repair class (306k tokens in one audit, scaling with Orrery workload).

## Testing

- 353 passed across schema/validator/two-pass/native/correspondence/config targets; full test_lore 210 passed; pg clone grammar/budget tests green; **live OpenAI strict decode with `@openai.gaia` passed** (env-gated). Black/flake8/mypy clean. Rebased over #643 without touching its logging.

### DIFF
--- a/nexus.toml
+++ b/nexus.toml
@@ -1101,6 +1101,8 @@ generation_timeout_seconds = 1200
 [apex.tag_library]
 # Measurement-stage A/B lever: false restores the complete described library.
 contextual = true
+# OpenAI Gaia strict schemas use registry-keyed enum aliases for closed fields.
+schema_enums = true
 suggestion_limit = 3
 
 # =============================================================================
--- a/nexus/agents/logon/gaia_registry_schema.py
+++ b/nexus/agents/logon/gaia_registry_schema.py
@@ -1,0 +1,401 @@
+"""Registry-keyed strict-schema models for the OpenAI Gaia seat.
+
+The static Skald models remain the application contract.  This module builds a
+request-time subclass whose closed-vocabulary fields reference named
+``Literal`` aliases, then coerces accepted output back to the static wire at the
+provider boundary.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from threading import Lock
+from typing import Annotated, Any, Literal, Mapping, Optional, Union, cast
+
+from pydantic import Field, create_model
+from typing_extensions import TypeAliasType
+
+from nexus.agents.logon.apex_schema import (
+    NewEntityDeclaration,
+    NewEntityPairTagHint,
+    OrreryAdjudication,
+)
+from nexus.agents.logon.orrery_tag_validation import (
+    StorytellerVocabulary,
+    read_storyteller_vocabulary,
+)
+from nexus.agents.logon.skald_wire import (
+    CharacterUpdateDelta,
+    FactionUpdateDelta,
+    PlaceUpdateDelta,
+    SkaldGaiaWire,
+    UpdatesBlock,
+)
+from nexus.agents.orrery.tag_library import _registry_digest
+
+
+class GaiaRegistrySchemaError(RuntimeError):
+    """Base error for registry-backed Gaia strict-schema construction."""
+
+
+class GaiaRegistryReadError(GaiaRegistrySchemaError):
+    """Raised when the live closed vocabulary cannot be read."""
+
+
+class GaiaRegistryVocabularyError(GaiaRegistrySchemaError):
+    """Raised when a seeded slot has an unusable closed vocabulary."""
+
+
+@dataclass(frozen=True)
+class GaiaRegistryVocabulary:
+    """Normalized, deterministic vocabulary used as a model-cache key."""
+
+    character_tags: tuple[str, ...]
+    place_tags: tuple[str, ...]
+    faction_tags: tuple[str, ...]
+    pair_tags: tuple[str, ...]
+    event_types: tuple[str, ...]
+
+
+@dataclass(frozen=True)
+class GaiaRegistryWireSpec:
+    """One loaded registry snapshot and its cached provider model."""
+
+    model: type[SkaldGaiaWire]
+    registry_digest: str
+    vocabulary: StorytellerVocabulary
+
+
+_MODEL_CACHE: dict[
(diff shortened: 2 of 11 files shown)


pythagorakase:
Two independent reviews in — a commissioned Sol worktree review and the Codex cloud reviewer — converging on the same site with complementary P2s; fix round in flight:

- **P2 orrery_tag_validation.py:299 (both reviewers)** — the `entity_pair_tags_target_clear_inbound` arm validates only registry membership: it neither requires a target binding (actor-only proposals crash at commit, `events.py:1816-1818`) nor checks the tag's `object_kinds` allowance for the target. Both checks move to generation time as repairable rejections — exactly the #642 class this PR closes.
- **P3 (Sol)** — true emitted enum total is 626, not 617; the measurement omitted small inline Literals. Measurement and assertions corrected to count every enum; headroom against the 1,000 limit computed from the true total.

All other adversarial axes held: partition equivalence (incl. multi-kind categories and the extend_expiry asymmetry), dynamic-subclass hydration and `new_entities` round-trip, strict-schema legality, digest+slot cache keying, config-gate byte-identity, loud failure paths, untouched Anthropic/lenient arms, and #643-logging compatibility.

PR #652, opened 2026-07-31
Split PresenceRef by Kind in the Skald Wire Grammar

### LINKED ISSUE #639: Tighten SkaldWriterWire Grammar: Split PresenceRef by Kind, Revisit Choices Bounds
## Problem

3 of 15 rejections (67,060 tokens, 18% of repair waste) in the 2026-07-30 scratchpad audit were `SkaldWriterWire` Pydantic failures on the writer seat. Exact causes are unattributable until #637 lands (the repair loop logged nothing); token-delta forensics bound the candidate set to {presence ontology validator, choices count, letter null}. All three writer rejections occurred on roster/setting-transition turns (pursuit, stranger identification, relocation), circumstantially pointing at the presence-ontology branch.

## Proposals (hold until #637 confirms the distribution)

1. **Split `PresenceRef` into `CharacterRef`/`PlaceRef`** with single-value `kind` enums per slot. Four of the five `PresenceDelta`/`SceneReset` cross-field validators exist only because one shared ref type carries `kind: Literal["character","place","faction"]`; per-slot types make `enter`/`exit`/`scene_reset.present` character-only and `transit`/`scene_reset.place` place-only in the grammar itself.
2. **Choices bounds**: the emitted schema carries `minItems: 2, maxItems: 4`, but OpenAI's strict decoder ignores both — the bound is client-side only, so a 5th choice costs a full ~22k-token retry. All 27 accepted turns emitted exactly 4; consider whether a fixed-length tuple is acceptable authorially or the client-side bound stays as the price of variable choice counts.
3. **No change to the letter-null trick** (strict grammar requires all keys; nullable-then-reject is deliberate) and **no change to `max_letter_tokens = 300`** — zero rejections from it, but note one writer letter landed at 298/300, so the cap is 0.7% from binding and worth watching in longer campaigns.

## Acceptance Criteria

- After #637's logging names the real writer-rejection causes, implement whichever grammar tightenings the data supports and demonstrate the class shrinking in a live run.

Evidence: `temp/scratchpad_audit_2026-07-30_044920Z/`; wire schema at `nexus/agents/logon/skald_wire.py:359` (writer), `:106` (PresenceDelta validators). Related: #638 covers the Gaia-side registry class (82% of the waste).

pythagorakase:
Closes #639.

## What This Does

Every presence slot in the Skald wire shared one `PresenceRef` with `kind: Literal["character","place","faction"]`, policing kind correctness with three cross-field `model_validator`s — so a wrong-kind reference cost a full ~22k-token model retry. This moves the constraint into the grammar, where OpenAI strict decoding makes wrong-kind references unemittable:

- `CharacterRef`/`PlaceRef` subclass `PresenceRef` with single-value `kind` Literals (required, no default — the decoder must emit the pinned value; `isinstance` and shared helpers keep working).
- `enter`/`exit`/`scene_reset.present`/`baseline.present` are now `CharacterRef`; `transit`/`scene_reset.place`/`baseline.setting` are `PlaceRef`; `mentions` keeps the three-kind base type (faction mentions are legitimate).
- All three kind-checking validators are deleted. Two rules the grammar cannot express survive app-side: scene_reset-vs-enter/exit exclusivity and the enter/exit same-name overlap check.
- Frozen non-changes per the issue: choices keep `minItems 2 / maxItems 4` client-side (authorial latitude for 2–3-choice moments), letter nullable-then-reject and `max_letter_tokens = 300` untouched.

Writer schema cost: 5,583 bytes / 1,248 tokens → 6,533 bytes / 1,458 tokens (+210 tokens/turn) from the added `$defs`. Against the measured writer-rejection class (~22k tokens per retry, 3 of 27 turns in the original audit ≈ 2,400 tokens/turn expected), the trade is strongly net-positive.

## Tests

Partition-equivalence coverage (every shape the deleted validators rejected now fails at parse from JSON input), retained-rule coverage, faction-mention latitude, schema `const` rendering, size budget, and a live OpenAI writer decode on a roster/setting-transition beat — the exact family where the original #639 rejections clustered — via the `@openai` registry role, zero retries. Full suite: 2,128 passed, 484 skipped.

### DIFF
--- a/nexus/agents/logon/skald_wire.py
+++ b/nexus/agents/logon/skald_wire.py
@@ -7,7 +7,7 @@ per-chunk mentions, never members of the carried scene roster.
 from __future__ import annotations
 
 import json
-from typing import Any, Dict, List, Literal, Optional
+from typing import Any, Dict, List, Literal, Optional, Sequence, TypeVar
 
 from pydantic import BaseModel, ConfigDict, Field, model_validator
 
@@ -81,24 +81,26 @@ class PresenceRef(BaseModel):
     model_config = ConfigDict(str_strip_whitespace=True, extra="forbid")
 
 
+class CharacterRef(PresenceRef):
+    """Character-only provider-facing entity reference."""
+
+    kind: Literal["character"] = Field(description="Referenced entity kind.")
+
+
+class PlaceRef(PresenceRef):
+    """Place-only provider-facing entity reference."""
+
+    kind: Literal["place"] = Field(description="Referenced entity kind.")
+
+
 class SceneReset(BaseModel):
     """Fresh character roster and setting after a scene cut."""
 
-    place: PresenceRef = Field(description="New setting place.")
-    present: List[PresenceRef] = Field(
+    place: PlaceRef = Field(description="New setting place.")
+    present: List[CharacterRef] = Field(
         default_factory=list,
         description="Complete present-character roster.",
     )
-
-    @model_validator(mode="after")
-    def validate_presence_kinds(self) -> "SceneReset":
-        """Reject ontology-invalid reset references."""
-
-        if self.place.kind != "place":
-            raise ValueError("scene_reset.place must be a place")
-        if any(reference.kind != "character" for reference in self.present):
-            raise ValueError("scene_reset.present accepts characters only")
-        return self
 
     model_config = ConfigDict(extra="forbid")
 
@@ -106,11 +108,11 @@ class SceneReset(BaseModel):
 class PresenceDelta(BaseModel):
     """Sparse changes to scene presence and references."""
 
-    enter: List[PresenceRef] = Field(
+    enter: List[CharacterRef] = Field(
         default_factory=list,
         description="Characters entering the scene.",
     )
-    exit: List[PresenceRef] = Field(
+    exit: List[CharacterRef] = Field(
         default_factory=list,
         description="Characters leaving the scene.",
     )
@@ -118,7 +120,7 @@ class PresenceDelta(BaseModel):
         default_factory=list,
         description="Absent entities referenced this chunk.",
     )
-    transit: List[PresenceRef] = Field(
+    transit: List[PlaceRef] = Field(
         default_factory=list,
         description="Places passed through this chunk.",
     )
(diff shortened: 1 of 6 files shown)


pythagorakase:
## Review Round 1 Triage (Independent Sol Review)

**[P2] `_prompt_guide_enum` ignores `const` — CONFIRMED, fix in flight.** Pydantic renders single-value Literals as `const`, not `enum`, so the Anthropic prompted-transport guide would describe the new `kind` fields as unconstrained strings — dropping the partition on the one arm where the guide is the only enforcement signal. Fix: honor `const` as a one-element enum, with a rendered-guide test pinning both kinds.

**[P3] `$ref`-sibling description forces strict-mode expansion — CONFIRMED, fix in flight.** `SceneReset.place`'s field description sits beside the `$ref`, inlining `PlaceRef` (415 bytes / 91 tokens of duplication — 91 of this PR's 210-token schema growth is pure waste). Fix: drop the redundant description, sweep other retyped direct slots for the same pattern, re-measure.

**[P3] Partition table covers 7 of 14 wrong-kind combinations — CONFIRMED, fix in flight.** All 7 retyped slots × both wrong kinds will be asserted.

Verdict was CHANGES_REQUIRED; round 2 review will confirm the fixes before merge.

pythagorakase:
## Review Round 2 Triage (Independent Sol Review)

Round 2 confirmed all three round-1 fixes (const-aware prompted guide, schema slimmed to 6,118 bytes / 1,367 tokens, 14-combination coverage) and surfaced one new P3 — **confirmed, fix in flight**:

**[P3] The wrong-kind tables assert bare `ValidationError`, not the kind mismatch itself.** A row could pass on an unrelated validation artifact. Fix: assert `type == "literal_error"` with `loc` ending at the injected reference's `kind` path via `errors()`, per combination. Test-only change.

The fix will be verified mechanically (focused run + assertion inspection) before merge; no further full review round is warranted for a test-assertion tightening.

pythagorakase:
Round-2 P3 fix verified mechanically: all 14 combinations now assert the structured (literal_error, kind-loc-suffix) pair; wire module runs 86 passed locally at the PR head. Gate satisfied: P2 + both P3s fixed, round-1 clean areas reconfirmed in round 2, live decode green on the slimmed schema. Merging on green CI.

Issue #465, opened 2026-07-11
In-UI Local Model Management: Pick, Activate, Download, and Register Local GGUF Models


pythagorakase:
Builds on the llama-server migration (#464). Turn local inference from a config-file affair into a user-facing capability in the settings pane: pick a local model, manage which quant is downloaded/active, and point at existing GGUFs on disk.

## Decisions (settled with owner)

- **Download mechanism:** `huggingface_hub` (resumable, progress, one dependency).
- **Storage:** BOTH — a default NEXUS models directory (config) for downloads, AND a "point at an existing file" flow via a **server-side directory browser** (a browser `<input type=file>` yields content not a path, and the Tauri shell has no IPC, so a small FS-browse endpoint under allowed roots is the web-native mechanism; Access-gated).
- **File verification is trivial:** GGUF self-describes — 4-byte `GGUF` magic + a KV metadata header (`general.architecture`, name, param count, quant type). The `gguf` python package reads it without loading the model, so a user-pointed file is recognized/validated in ~20 lines. No hashing/re-download.
- **Phasing:** split into three (below).

## Architecture

- **Curated catalog** (`nexus.toml`): supported families → HF repo → blessed quants (Q4_K_M, Q6_K, …) with sizes + RAM-fit hints. Drives the UI and keeps us from exposing every quant on HF.
- **Default models dir** (config), plus registered external paths.
- **A local-inference process manager** (the load-bearing piece): owns the `llama-server` subprocess — start with a chosen GGUF (`--alias` = registry id), stop, and **swap** (activate a different quant = restart pointed at its path). The `@local` provider serves whatever is active. This supersedes the static `[runtime.services.llama_server]` path from #464 for the dynamic case.
- **Endpoints** `/api/local-models/*`: status (installed/active/size per catalog quant + registered files), activate (swap llama-server), download (Phase 2), delete (Phase 2), browse-fs + register-path (Phase 1), gguf-verify.
- **UI:** local families appear in the settings model picker; selecting reveals a management panel — per quant: installed (delete), not installed (download + size + RAM-fit), active indicator; plus "point at existing file". House-idiom (dim/glow, transient progress); shadcn crawl for progress/download affordances.

## Phase 1 — Lifecycle & activation (backbone, no downloads)
- Local-inference process manager (start/stop/swap llama-server by GGUF path).
- Catalog + scan the default dir + register external paths via the FS-browse + GGUF-header verify.
[...]

PR #656, opened 2026-08-03
Card Intertitle and LocalModelRows; fix grid overflow on 65 cards

pythagorakase:
Design-sync of the `ui/` component library to the NEXUS Iris project on claude.ai/design. **95 → 97 components.** Already uploaded and verified; this PR lands the sync *inputs* so future runs reuse them.

## New cards

`Intertitle` (#456) and `LocalModelRows` (#465 Phase 1b) were both riding the bundle uncarded — importable by the design agent but with no preview, no prop doc, no card. Both are now pinned in `componentSrcMap` with authored previews, which clears the standing deferral NOTES.md was tracking.

`ds-provider.tsx` gains a GET stub for `/api/local-models/status` and `/download`, built from the real `nexus.toml` Hermes catalog. This is **required, not cosmetic**: `LocalModelRows` does `if (statusQuery.error && !status) throw statusQuery.error`, so an unmocked endpoint throws and blanks every pane that mounts it — and `SettingsPane` has embedded it since #482, i.e. since the last sync.

One constraint shaped the preview: a card renders all its cells in a single page load against one React Query cache, so every cell necessarily sees the same payload — per-cell mock scenarios are impossible. The stub therefore uses one scenario tuned to show the maximum number of row states at once (`system_ram_gb: 48` puts two 70B quants over the memory ceiling; one quant serving, one on disk, one downloadable, one mid-download at 37%).

## Card layout — 65 cards were cropping

Validate flags `[GRID_OVERFLOW]` when stories crop or escape their cell in the design pane's grid view. **65 of 97 were flagged and had never been triaged** — `config.json` carried no `overrides` key at all, so this has been shipping since the first sync. All 65 now carry the remedy validate names: 51 `cardMode: column` (keeps every story, one per full-width row) and 14 `cardMode: single` for the portal/overlay components, with a primary chosen to favour the open state showing real content (`Select`→`ModelPicker` over `Closed`, `WaitScreen`→`Generating`, `AlertDialog`→`WipeSlot`).

No grades were invalidated: `configSlicesFor` deliberately strips `cardMode`/`primaryStory` from the graded component slice, since they arrange the default card view rather than any solo-captured story.

## NOTES.md

Four things a future sync needs:

- **Never symlink `node_modules` when syncing from a worktree.** esbuild inlines each resolved module's path as a comment inside `_preview/<Name>.js`; a symlink resolves to its realpath and emits `../../../../ui/node_modules/lucide-react/…`, churning render hashes for every preview that imports a bundled dep. That made 10 components look `changed` purely because of *where* the sync ran, and inflated the upload set from 2 to 12. Hardlink-copy (`cp -Rl`) instead.
- The cached `remote-sync.json` anchor is always one generation stale — it's fetched before the run that then uploads a new one.
- The cardMode triage above, with the reasoning behind each `single` pick.
- **Open gap:** all 97 `.d.ts` files are permissive stubs (`{ [key: string]: unknown }`) because the phantom `--entry` points at a Vite lib build with no declaration tree (`[DTS] parsed 0 .d.ts files`). Filed separately.

## Verification

- Render check clean: 97 components, `bad 0`, `thin 0`, `variantsIdentical 0`, zero floor cards
- **Zero warn lines** on the final driver run (was 65 `[GRID_OVERFLOW]`)
- Both new cards graded `good` against the absolute rubric from their review sheets
- Conventions header re-validated against the fresh build — every token family, font utility, theme class and component name it names still resolves
- 514 files uploaded, 0 deletes, anchor written last

https://claude.ai/code/session_019697mhq4W45PA4Srr1AFib

### DIFF
--- a/ui/.design-sync/NOTES.md
+++ b/ui/.design-sync/NOTES.md
@@ -191,9 +191,15 @@ state diversity into one rich stub scenario instead.
   `[aria-expanded="false"]` so it is idempotent and never toggles an open family shut.
   Same principle as the ContextMenu preview: drive the real trigger, never hand-write
   the open state.
-- It renders bare `<li>`s — wrap in `<ul className="model-list">`, as `SettingsPane` does.
-  `.lm-*` and `.intertitle` rules are unscoped, so no `.nexus-shell` wrapper is needed
-  (unlike `SettingsPane`'s footer buttons).
+- It renders bare `<li>`s — wrap in `<ul className="model-list">` inside a
+  **`.nexus-shell` ancestor**, as `SettingsPane` does. Checking that the `.lm-*` rules
+  are unscoped is NOT sufficient: the component also emits `.caption`, `.btn-soft`, and
+  `.btn-primary`, and nexus-layout.css defines all three ONLY under `.nexus-shell`, so a
+  bare `<ul>` renders the quant labels and the EJECT / APPLY buttons as browser defaults
+  (caught in Codex review of PR #656). Override the shell's `100vh` and its `52px 1fr`
+  row grid locally. **When auditing a preview's wrapper, grep every class the component
+  emits — not just its own namespace.** `Intertitle` was checked the same way and is
+  genuinely clean (only unscoped `.intertitle*` rules).
 
 ## Component-type recipes (folded from wave-2 authoring)
 
--- a/ui/.design-sync/NOTES.md
+++ b/ui/.design-sync/NOTES.md
@@ -61,11 +61,41 @@ around it:
 - Review sheet renders each cell as a populated top band + an empty band below —
   the empty band is sheet layout, NOT a missing render.
 
+## OPEN GAP: every `.d.ts` is a permissive stub
+
+**All 97 components ship `interface <Name>Props { [key: string]: unknown }`** — no real
+prop contract. The `.d.ts` is what the claude.ai/design agent codes against, so today it
+learns each component's API only from the `.prompt.md` and the preview card, never from
+types. Pre-existing (it predates the 2026-07-30 sync; not a regression).
+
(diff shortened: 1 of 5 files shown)


pythagorakase:
The `.d.ts` gap noted under Verification is filed as #657 — all 97 prop contracts are permissive stubs because the phantom `--entry` has no declaration tree. Out of scope here; candidate fixes ranked in the issue.

pythagorakase:
Fixed in 46059f58 — the finding was correct and I verified it against the CSS rather than taking it on faith.

All three of `.nexus-shell .caption`, `.nexus-shell .btn-soft`, and `.nexus-shell .btn-primary` are shell-scoped, and `LocalModelRows` emits all three (`caption` ×5, `btn-soft` on EJECT, `btn-primary` on APPLY). Checking that its own `.lm-*` rules were unscoped was not sufficient — that was my error.

Audited the other 96 previews for the same class of mistake: `LocalModelRows` was the only one emitting a shell-scoped class without a wrapper. `SettingsPane` already had one; `Intertitle` emits only unscoped `.intertitle*` rules and is genuinely clean.

Fixed on the `SettingsPane` precedent (keep `nexus-shell`, override its `100vh` and 52px grid row, size to content). Visible confirmation from the fresh capture: quant labels moved from plain lowercase `q4_k_m 42.5 gb` to letterspaced small caps `Q4_K_M 42.5 GB`, and the card render grew 48KB → 140KB. It also stopped being flagged `wide` by the grid-overflow check.

NOTES.md's claim that no wrapper was needed here is corrected, with the general rule: when auditing a preview's wrapper, grep every class the component emits, not just its own namespace.

Re-verified end to end — 97 components, `bad 0`, `gridOverflow 0`, zero warn lines; both cells re-graded good; bundle re-uploaded to the project.

PR #659, opened 2026-08-03
Emit Real Declaration Files from the Design-Sync Lib Build

### LINKED ISSUE #657: Design-sync ships empty prop contracts: all 97 .d.ts files are permissive stubs
## What

Every component the `ui/` design-sync uploads to the NEXUS Iris project carries a `.d.ts` with no real API:

```ts
export interface LocalModelRowsProps {
  [key: string]: unknown;
}
export declare const LocalModelRows: React.ComponentType<LocalModelRowsProps>;
```

All **97 of 97** are like this. Confirmed by `grep -l "\[key: string\]: unknown"` across `ds-bundle/components/**/*.d.ts` → 97 hits, 0 files with real props.

## Why it matters

The `.d.ts` is the contract the claude.ai/design agent codes against — the skill calls it "the API contract it codes against", and the app's self-check reads each `<Name>Props` as the component's public interface. Today the agent learns our components' APIs only from `.prompt.md` and the rendered preview card, never from types. So it has no idea that `Intertitle` needs `season`/`episode`/`scene`/`worldLayer`/`worldTime`, or that `LocalModelRows` needs `selected`/`onPickLocal`/`knobs` — it has to infer everything from a picture. That degrades *every* design built with the library.

This is long-standing, not a regression — it predates the 2026-07-30 re-sync (PR #656), which is where it was noticed.

## Cause

The driver logs:

```
[DTS] parsed 0 .d.ts files from …/ui
```

`nexus-ui` is a Vite PWA app, not a publishable component library, so the sync uses a phantom `--entry` pointing at a Vite **library build** (`.design-sync/.cache/lib-dist/index.js`). That build emits only `index.js` + `style.css` — no declaration tree. With nothing for ts-morph to parse, prop extraction falls back to the permissive stub.

Not the `[DTS_REACT]` case: `@types/react` is installed and resolvable.

## Candidate fixes (cheapest first)

1. **Emit declarations from the lib build.** Add `vite-plugin-dts` to `.design-sync/vite.lib.config.mts`, or a `tsc --emitDeclarationOnly --outDir .design-sync/.cache/lib-dist` pass in `.design-sync/build.mjs` after the Vite step. The converter would then parse a real tree and all 97 contracts improve at once.
   - Main unknown: this codebase has **never been type-built as a library**, so the `tsc` error count is unknown and config work should be expected.
2. **Per-component `cfg.dtsPropsFor.<Name>`** with hand-written props bodies. Exact, but 97 entries that rot as the app evolves. Reserve for stragglers after (1).

The props interfaces themselves are well-formed in source (`IntertitleProps`, `LocalModelRowsProps`, etc.), so (1) should be largely mechanical once declarations emit.

## Verifying a fix

After rebuilding, `ds-bundle/components/**/*.d.ts` should contain real prop bodies, and `grep -c "\[key: string\]: unknown"` should drop toward 0. Re-run the driver and re-upload; `.d.ts`-only changes still ship because the upload partition is `sourceHashes`-based, and they don't invalidate any grade.

Context recorded under "OPEN GAP" in `ui/.design-sync/NOTES.md` (PR #656).

https://claude.ai/code/session_019697mhq4W45PA4Srr1AFib

pythagorakase:
Closes #657.

## What

The design-sync lib build now emits a real TypeScript declaration tree, so the converter extracts genuine prop contracts instead of falling back to `interface <Name>Props { [key: string]: unknown }` for all 97 components.

- **`build.mjs`** gains a declaration-only `tsc` pass (`tsconfig.lib.json`, strict, `noEmitOnError`) after the Vite lib build, then generates a curated top-level `index.d.ts` barrel from the exact `componentSrcMap` surface — failing loudly on any missing declaration. The barrel matters twice: the converter's glob ignores hidden directories (so nested output alone would still parse 0 files), and the runtime entry's hundreds of compound exports must not become standalone cards.
- **`package.json#types`** points the converter at the declaration barrel.
- **`config.json`** adds `dtsPropsFor` overrides (`[key: string]: never`) for the 13 genuinely prop-less shell/page components — a restrictive empty contract rather than the old permissive stub.
- **NOTES.md** OPEN GAP section replaced with the FIXED record and its load-bearing details.

## Verification

- `node .design-sync/build.mjs` completes: 129 `.d.ts` files emitted, 97/97 configured components resolve real props through the converter.
- `[key: string]: unknown` occurrences across emitted declarations: **1** — `NarrativeProgress.data`, an intentionally open server payload, not component props.
- Zero component-source changes were needed: the first strict declaration build passed clean. `npm run check` passes.
- JS bundle outputs (index.js + style.css + font-url rewrite) unchanged in shape.

Spot-checks: `IntertitleProps` carries `season/episode/scene/worldLayer/worldTime`; `LocalModelRowsProps` carries `selected/onPickLocal/knobs` with doc comments preserved.

## Post-Merge

The re-sync to claude.ai/design is a separate session-side step: `.d.ts`-only changes ship without invalidating any grade (the upload partition is `sourceHashes`-based).

### DIFF
--- a/ui/.design-sync/NOTES.md
+++ b/ui/.design-sync/NOTES.md
@@ -73,9 +73,9 @@ The converter resolves real props for **97/97** configured components instead of
 back to `interface <Name>Props { [key: string]: unknown }` for all of them.
 
 Two details are load-bearing:
-- The converter's package-wide glob ignores hidden directories. The top-level
-  `index.d.ts` makes `lib-dist/` the declaration root; emitting only the nested tsc
-  output would still report `[DTS] parsed 0 .d.ts files`.
+- Without the top-level barrel, the converter parses the 126 nested declaration files
+  but finds zero exported entry symbols. The curated `index.d.ts` supplies those entry
+  symbols from the package root.
 - The runtime entry intentionally exports hundreds of compound pieces. The declaration
   barrel exports only the 97 committed `componentSrcMap` roots so declaration discovery
   does not turn every `DialogTrigger`/`TableRow`-style subpart into a new card.
--- a/ui/.design-sync/NOTES.md
+++ b/ui/.design-sync/NOTES.md
@@ -10,23 +10,27 @@ converter expects an installed library with a `dist/` + `.d.ts` tree, so we rout
 around it:
 
 1. **`cfg.buildCmd` = `node .design-sync/build.mjs`** which: (a) runs
-   `gen-entry.mjs` to regenerate `.cache/lib-entry.tsx` (`export *` of all 95
+   `gen-entry.mjs` to regenerate `.cache/lib-entry.tsx` (`export *` of all 102
    component modules + the default re-exports + `DesignThemeRoot`) and
    `.cache/componentSrcMap.json`; (b) runs a **Vite library build**
    (`vite.lib.config.mts`) → clean ESM `.cache/lib-dist/index.js` + extracted
-   `style.css` (React externalized); (c) rewrites the brand `@font-face`
-   `/fonts/` urls so the converter can copy the TTFs.
+   `style.css` (React externalized); (c) runs the dedicated declaration-only
+   `tsconfig.lib.json` → a real `.cache/lib-dist/**/*.d.ts` tree plus a curated
+   top-level `index.d.ts`; (d) rewrites the brand `@font-face` `/fonts/` urls so
+   the converter can copy the TTFs.
 2. **Converter is run with a phantom `--entry`:**
    `node .ds-sync/package-build.mjs --config .design-sync/config.json --node-modules ./node_modules --entry ./.design-sync/.cache/lib-dist/index.js --out ./ds-bundle`
    The `--entry` points at the Vite dist (real → bundles the clean, pre-resolved
    JS, no app-isms). Vite absorbs the `@/` barrels, CSS side-effects, and font
    assets that esbuild (the converter's bundler) chokes on.
-3. **Discovery is via `componentSrcMap`** (95 entries) because there's no `.d.ts`
-   tree. This is a deliberate full enumeration (the skill's usual "sparse only"
-   rule doesn't apply — discovery depends on it). `gen-entry.mjs` regenerates
-   `componentSrcMap.json`; **on a component add/remove, re-merge it into
-   `config.json`** (it's static there). The `general` group = the shadcn `ui/`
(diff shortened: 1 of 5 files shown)


pythagorakase:

A fresh Sol (GPT-5.6-Codex) instance reviewed frozen commit `38ba6f17`, ran the build, `npm run check`, and the converter itself, and probed the barrel generator with synthetic declarations, an injected type error (confirming `noEmitOnError` emits zero files), and a temporarily dangling `package.json#types` (all consumers exit 0). Verdict: **CHANGES_REQUIRED** — one P1, one P2, one P3.

### [P1] Extracted per-card contracts carry dangling type references — REBUTTED as out-of-diff; filed as #662

Real finding (113 errors across 24 of 97 extracted cards, e.g. `LocalModelRowsProps.knobs` referencing an undeclared `LocalModelsUiKnobs`), but the defect lives in the converter's extraction pass under `ui/.ds-sync/` — gitignored session-harness tooling outside this PR's diff, which the reviewer itself could only exercise via the locally staged copy. This PR strictly improves all 97 contracts (73 fully self-contained; the remaining 24 now carry informative named type references instead of `[key: string]: unknown`), and the primary consumer is the design agent reading contracts as documentation, not a compiler. The transitive-type inlining fix is tracked as **#662** with the reviewer's evidence and a verification recipe.

### [P2] Regex-based default-export selection — CONFIRMED and fixed (`5cea43fb`)

The reviewer proved all three synthetic failure modes (named-plus-unrelated-default, default-not-the-component, doc-comment decoy) while AST comparison showed all 97 *current* entries resolve correctly — a latent trap, not a live defect. The barrel generator now resolves exports by symbol via the TypeScript checker (`getExportsOfModule`): named export preferred, genuine default as fallback, loud error when neither exists. Regenerated barrel is byte-identical for all 97 components (SHA-256 verified); all synthetic cases now resolve correctly, including the loud missing-export throw.

### [P3] NOTES.md degradation claim — CONFIRMED and fixed (`5cea43fb`)

The reviewer measured the true no-barrel behavior: 126 declarations parsed with *zero exported entry symbols* (not zero parsed files). The note now records the measured mechanism.

**Round-2 note:** both fixes are mechanical with reviewer-designed verification (empty barrel diff; the five synthetic cases); I verified the diff and outputs directly rather than commissioning another round.

Merging under delegated authority. Post-merge, the re-sync to claude.ai/design ships the improved contracts without invalidating grades.

PR #660, opened 2026-08-03
Fail-Closed QA Guard While Detached Maturation Is In Flight

### LINKED ISSUE #653: Night QA post-check misses detached Retrograde maturation usage
## Summary

The Night QA exact-usage protocol can return `continue` from the required
post-command check while a detached Retrograde maturation provider call caused
by that command is still in flight. The late event is charged to a later
pre-check instead of the generating command; an immediately following probe can
therefore overlap an unaccounted provider response.

## Public reproduction

Environment:

- clean `main` at `3751f739ce071f7645b8df22334db970f7a2e3e9`
- isolated slot 4 / port 8012
- every provider route pinned to `gpt-5.6-terra`
- seeded 42-chunk campaign with no carried Orrery jobs

1. Run the guard pre-check.
2. Use public `nexus --json continue --slot 4 --model gpt-5.6-terra
   --user-text ...` to introduce a genuinely new named character.
3. Use a second guarded public continue to accept that pending turn, which
   persists the stub and enqueues runtime Retrograde maturation.
4. Immediately run the required `qa_shift.py check ARCHIVE --expect-call`.
5. Before another model-generating command, run the normal pre-check again.

Observed usage:

- The immediate post-check returned `status=continue` with 82,100 tokens from
  writer, `retrograde_seed_candidates`, `retrograde_seed_selection`, and Gaia.
- The public CLI had already returned at `23:40:44.091` local time.
- At `23:40:54.776`, 10.7 seconds later, the detached
  `retrograde_expansion` response recorded another 14,532 tokens.
- The next *pre-check* reported that event as its own delta. The true command
  cost was 96,632 tokens, but the shift's `max_command_delta` remained 82,100.

PostgreSQL independently showed maturation job 15 for new entity `Nika Rel`,
requesting chunk 47, `created_at=00:40:08.936907-04`, and
`updated_at=00:40:55.410343-04` with `state=succeeded` and a persisted result
manifest. No carried job existed before the probe.

## Expected versus actual

Expected: once the post-command check returns `continue`, every provider call
caused by that public command is settled and represented in the command delta,
or the guard fails closed because provider-capable work remains queued/running.

Actual: the guard inspects only completed ledger events. It has no signal for
the still-running detached maturation thread, so it can authorize another
remote request before the late response is recorded.

Impact:

- breaks the mission's no-unaccounted-response safety boundary;
- misattributes per-command usage and understates the largest command delta;
- makes a bounded concurrency group's event set ambiguous if a prior
  maturation finishes during it; and
- can permit another request closer to the token fence than the guard knows.

## Likely cause

`nexus/api/narrative.py:1172-1195` deliberately launches
`drain_maturation_jobs_sync` on a daemon thread so gameplay does not wait for
multi-minute frontier calls. That product behavior is intentional.

`scripts/qa_shift/qa_shift.py:525-659` compares completed usage-event counts and
totals only. It does not consult provider in-flight state or the durable
`orrery_maturation_jobs` queue before returning `continue`.

## Acceptance criteria

- A post-check for a command that enqueues maturation either waits until all
  provider-capable work attributable to the command settles, or returns a
  fail-closed/pending status that forbids the next remote request.
- The command delta includes seed-candidate, seed-selection, expansion, writer,
  and Gaia events that the command caused, including delayed completions.
- A pre-check cannot authorize a new probe while the QA slot has an in-flight
  provider response or a provider-capable queued/leased maturation job.
- Detached maturation remains non-blocking for ordinary gameplay.
- A deterministic test delays `retrograde_expansion` until after the public
  continue returns and proves the guard neither loses nor misattributes it.

Deduplication: all 163 issues and 100 recent merged PRs were reviewed. #626/#627
cover completed-response accounting, #552 covers replay placement of delayed
maturation writes, and #647 covers route leakage; none covers this in-flight
post-check gap.

Codex — GPT-5

pythagorakase:
Closes #653.

## What

The Night QA usage guard could return `continue` from a post-command check while a detached Retrograde maturation call caused by that command was still in flight; the late usage event then poisoned the next pre-check's delta (observed twice: 14,532 and 13,566 late tokens, understating `max_command_delta` by ~15%).

The durable in-flight signal already existed — `orrery_maturation_jobs` with non-terminal states `queued`/`leased` in the per-slot database — the guard just never consulted it. Now it does, via a new public CLI readout, preserving the script's deliberate subprocess-only posture:

- **`nexus jobs --slot N --json`**: per-state counts plus a diagnosable non-terminal job list, built on the extended `load_maturation_status_sync` (one atomic query; the Orrery status consumer updated for the richer shape).
- **`JobsReader` seam** in `qa_shift.py`, mirroring `UsageReader`.
- **Third check status `pending` (exit code 3)**: any non-terminal job → the check reports the jobs and does NOT advance the usage watermark, does not update `max_command_delta`, and authorizes nothing. Re-running after settlement computes the delta from the original watermark, so late maturation usage lands in the causal command's delta — attribution is fixed by *refusing to move the goalposts*, not by event correlation.
- **Read ordering closes the race**: the queue is read *before* the usage snapshot. A job reaches a terminal state only after its usage event is durably recorded, so "settled queue, then usage" cannot miss a late event (regression-tested).
- **`begin` fails closed** on a dirty queue; **`finish` never blocks** but reports the jobs snapshot and `usage_settled: false` when unsettled.
- **Mission prompt** gains the pending protocol block (retry cadence and the twenty-minute plateau live there as operator guidance, not code constants).

Detached maturation itself is untouched — gameplay stays non-blocking (spec decision 10).

## Called-Out Drive-Bys (Mypy on Touched Files)

- `run_continue` in `nexus/cli.py` had an implicit-`None` fall-through when slot state was neither wizard nor narrative mode; now returns a loud structured error.
- `requests` import annotated `# type: ignore[import-untyped]`.

## Tests

- `tests/test_qa_shift.py` (+13 unit tests through the reader seams): pending without watermark advance; the issue's full timing scenario end-to-end (pre-check → leased job → `pending` → settle + late `retrograde_expansion` event → `continue` with the true 350-token command delta); queue-before-usage race ordering; only non-terminal states block; dirty-queue begin refusal; `usage_settled` both ways; exit-code contract.
- `tests/test_jobs_cli_pg.py` (new, `requires_postgres`): disposable `qa653_*` template clone, job rows across all four states, counts + non-terminal list shape.

## Gates

Black / flake8 / mypy clean on touched files; guard suite 27 passed; pg test passed; full suite **2147 passed, 501 skipped**.

### DIFF
--- a/nexus/agents/orrery/retrograde_maturation.py
+++ b/nexus/agents/orrery/retrograde_maturation.py
@@ -1142,20 +1142,60 @@ def namespace_expansion_event_refs(
     return payload
 
 
-def load_maturation_status_sync(cur: Any) -> dict[str, int]:
-    """Return compact maturation queue counts for status snapshots."""
+def load_maturation_status_sync(cur: Any) -> dict[str, Any]:
+    """Return maturation queue counts and diagnosable non-terminal jobs."""
 
     cur.execute(
         """
-        SELECT state::text AS state, count(*) AS count
+        SELECT
+            jsonb_build_object(
+                'queued', count(*) FILTER (WHERE state = 'queued'),
+                'leased', count(*) FILTER (WHERE state = 'leased'),
+                'succeeded', count(*) FILTER (WHERE state = 'succeeded'),
+                'failed', count(*) FILTER (WHERE state = 'failed')
+            ) AS counts,
+            COALESCE(
+                jsonb_agg(
+                    jsonb_build_object(
+                        'id', id,
+                        'state', state::text,
+                        'entity_kind', entity_kind,
+                        'entity_name', entity_name,
+                        'requesting_chunk_id', requesting_chunk_id,
+                        'attempts', attempts,
+                        'available_at', available_at,
+                        'lease_until', lease_until
+                    ) ORDER BY id
+                ) FILTER (WHERE state IN ('queued', 'leased')),
+                '[]'::jsonb
+            ) AS non_terminal_jobs
         FROM orrery_maturation_jobs
-        GROUP BY state
         """
     )
-    counts = {"queued": 0, "leased": 0, "succeeded": 0, "failed": 0}
-    for row in cur.fetchall():
-        counts[str(_row_value(row, "state", 0))] = int(_row_value(row, "count", 1))
-    return counts
+    row = cur.fetchone()
+    if row is None:
+        raise RuntimeError("Maturation status query returned no row")
+    raw_counts = _row_value(row, "counts", 0)
+    raw_jobs = _row_value(row, "non_terminal_jobs", 1)
+    if not isinstance(raw_counts, Mapping) or not isinstance(raw_jobs, list):
+        raise RuntimeError("Maturation status query returned an invalid shape")
+    counts = {
+        state: int(raw_counts.get(state, 0))
+        for state in ("queued", "leased", "succeeded", "failed")
+    }
+    return {"counts": counts, "non_terminal_jobs": raw_jobs}
+
+
+def load_maturation_status_for_slot_sync(slot: int) -> dict[str, Any]:
+    """Return the durable maturation queue snapshot for one save slot."""
+
+    conn = _connect_for_slot(slot)
+    try:
+        with conn:
+            with conn.cursor(cursor_factory=RealDictCursor) as cur:
+                return load_maturation_status_sync(cur)
+    finally:
+        conn.close()
 
 
 # ============================================================================
--- a/nexus/agents/orrery/worker.py
+++ b/nexus/agents/orrery/worker.py
@@ -326,10 +326,10 @@ def load_orrery_status_sync(
             with conn.cursor(cursor_factory=RealDictCursor) as cur:
                 maturation_counts = load_maturation_status_sync(cur)
                 return OrreryStatus(
-                    queued_maturation_jobs=maturation_counts["queued"],
-                    leased_maturation_jobs=maturation_counts["leased"],
-                    succeeded_maturation_jobs=maturation_counts["succeeded"],
-                    failed_maturation_jobs=maturation_counts["failed"],
+                    queued_maturation_jobs=maturation_counts["counts"]["queued"],
+                    leased_maturation_jobs=maturation_counts["counts"]["leased"],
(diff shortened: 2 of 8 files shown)


pythagorakase:

A fresh Sol (GPT-5.6-Codex) instance reviewed frozen commit `99be417f` with a brief that specifically attacked the settlement-race claim, pending-wedge edges, snapshot strictness, the status-helper shape change, the CLI drive-bys, and test honesty. It ran the guard suite (27 passed), the pg test, and the full suite (2147 passed) itself.

**Verdict: CHANGES_REQUIRED — one P1, zero P2/P3.**

### [P1] Terminal `failed` does not guarantee durable usage — CONFIRMED and fixed (`459e75d1`)

Verified against code, both legs: `record_usage_event` raises `UsageWriteError` when the JSONL append fails *after* a successful provider response (`nexus/telemetry/usage.py:230-244`), and the drain's blanket handler routes that into `_mark_maturation_failed`, which at the attempt cap sets the job terminally `failed` (`retrograde_maturation.py:576, 1517-1528`). The jobs snapshot would then report settlement while the ledger is missing a response — `continue` with unaccounted usage. Terminal-`succeeded` remains safe (usage is durably appended before the success mark); the hole was exclusively failed-at-cap.

Fix (guard-side, conservative): `begin` stores `baseline_failed_jobs`; any increase in the settled path appends a new fail-closed stop reason `maturation_job_failed`; `finish`'s `usage_settled` now also requires no new failures; the mission prompt documents the reason. Four new deterministic regressions cover: failure-above-baseline → stop; pre-existing failures at begin → normal continues; failure during pending → still pending, then the settled re-run stops *and* retains correct late-event attribution; finish honesty. Guard suite 31 passed; full suite 2151 passed, 501 skipped.

**Round-2 note:** the fix is mechanical counting logic mapped 1:1 to frozen test scenarios; I verified the diff line-by-line against the fix order and ran the suites rather than commissioning another full review round. The settlement invariant is now: `continue` ⇔ no non-terminal jobs AND no new terminal failures — terminal-`succeeded` is the only state trusted to imply durable usage, per the reviewer's own analysis.

Merging under delegated authority.

PR #661, opened 2026-08-03
Normalize Same-Character Enter/Exit to Net End-of-Turn State

### LINKED ISSUE #639: Tighten SkaldWriterWire Grammar: Split PresenceRef by Kind, Revisit Choices Bounds

pythagorakase:
Closes #639 (the reopened same-character enter/exit branch).

## What

PR #652 deliberately retained the app-side enter/exit overlap rejection because a strict JSON grammar cannot express cross-array disjointness. The first post-#652 roster-transition probe hit exactly that branch: a grammar-valid out-and-back move (new character enters the scene, then exits through the doors) was rejected with `presence cannot enter and exit the same character: nika rel` at a cost of 23,465 tokens for the writer retry.

A `mode="before"` model validator on `PresenceDelta` now normalizes the payload before the after-validator runs: every character in the casefolded enter∩exit intersection is removed from both lists and collapsed to one `kind="character"` mention (id carried from whichever side had one; de-duplicated against existing mentions). Each normalization logs the grep-stable marker `presence out-and-back normalized to mention: <name>` for QA-night rejection mining.

## Why This Is Correct Without Consulting the Prior Roster

The hydration algebra is roster = dedupe(baseline.present + enter) − exit. Working all four prior-state × intent combinations:

- **Absent + out-and-back** (the observed case): dropping both nets to absent-at-end; the mention preserves the junction row.
- **Present + leave-and-return**: dropping both leaves the baseline roster intact — net present; the extra mention collapses into the present row via the commit handlers' present-wins `ON CONFLICT` clause. (Raw hydration of the un-normalized payload would have gotten this case *wrong* — exit subtracts after enter adds, marking a still-present character absent.)

Loud rejections are preserved where the contradiction is genuine: `scene_reset` alongside enter/exit still raises, and the after-validator's overlap check stays as a defense-in-depth tripwire behind the normalizer.

No schema change (before-validators don't alter the emitted JSON schema — the writer schema budget test passes untouched) and no prompt change (normalization makes the encoding free; per-turn prompt tokens teaching avoidance would cost more than the thing they prevent).

## Tests

Seven new regressions in `tests/test_skald_wire.py`: casefolded normalization with id carry, idempotent re-validation (`model_validate(model_dump())` — the #651 boundary-coercion lesson), mention de-duplication, byte-identical passthrough of non-overlapping payloads, full raw-JSON wire decode, and both hydration end-states (absent-from-baseline → MENTIONED; present-on-baseline → PRESENT + mention pair). Warning-marker assertions via caplog. Also fixes the pre-existing mypy error at the old line 1973 (dict literal → constructed `OrreryReplacementStateDelta`; runtime-identical).

## Gates

Black / flake8 / mypy clean; focused suite 92 passed, 2 skipped; full suite **2142 passed, 500 skipped**.

Post-merge: the nightly QA lane's forced roster-transition family measures the class live; the reopen comment's acceptance metric (rejected attempts per writer attempt) comes from the shift's rejection ledger.

### DIFF
--- a/nexus/agents/logon/skald_wire.py
+++ b/nexus/agents/logon/skald_wire.py
@@ -167,28 +167,41 @@ class PresenceDelta(BaseModel):
         if not overlap:
             return data
 
+        removed = [*enter, *exit_references]
+        matching_by_name = {
+            name_key: [
+                reference
+                for reference in removed
+                if reference.name.casefold() == name_key
+            ]
+            for name_key in overlap
+        }
+        normalizable_overlap = {
+            name_key
+            for name_key, matching in matching_by_name.items()
+            if len({reference.id for reference in matching if reference.id is not None})
+            <= 1
+        }
+        if not normalizable_overlap:
+            return data
+
         normalized = dict(data)
         normalized["enter"] = [
             reference
             for reference, parsed in zip(enter_data, enter)
-            if parsed.name.casefold() not in overlap
+            if parsed.name.casefold() not in normalizable_overlap
         ]
         normalized["exit"] = [
             reference
             for reference, parsed in zip(exit_data, exit_references)
-            if parsed.name.casefold() not in overlap
+            if parsed.name.casefold() not in normalizable_overlap
         ]
         normalized_mentions = list(mentions_data)
         mention_keys = {
             (reference.kind, reference.name.casefold()) for reference in mentions
         }
-        removed = [*enter, *exit_references]
-        for name_key in sorted(overlap):
-            matching = [
-                reference
-                for reference in removed
-                if reference.name.casefold() == name_key
-            ]
+        for name_key in sorted(normalizable_overlap):
+            matching = matching_by_name[name_key]
             name = matching[0].name
             reference_id = next(
                 (reference.id for reference in matching if reference.id is not None),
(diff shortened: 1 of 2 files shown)


pythagorakase:

A fresh Sol (GPT-5.6-Codex) instance reviewed frozen commit `a8bde664` with a brief instructing it to attack the baseline-free correctness argument, hostile payload shapes, idempotency, casefold edges, and schema stability. It ran the focused and full suites itself and verified both wire schemas byte-identical to base by SHA-256 — stronger than the size-budget assertion the build was held to.

**Verdict: CHANGES_REQUIRED — one P1.**

### [P1] Casefold-name overlap conflates distinct canonical identities — CONFIRMED and fixed (`c4a384a4`)

Verified against code: the normalizer keyed the enter∩exit overlap on casefolded name only. Two different canonical characters with casefold-equal names — enter `Nika Rel (id=31)` / exit `NIKA REL (id=32)` with 32 present on the baseline — were conflated: both entries dropped, one mention with id 31. Hydration would leave 32 present (they actually exited) and demote 31 to mentioned; the commit conflict key is `(chunk_id, character_id)`, so nothing collapses downstream. The normalization had made the retained contradiction validator unreachable for exactly the case where its loud rejection is correct.

Fix: normalization now applies per-name only when identities are compatible (at most one distinct non-None id across both lists). Conflicting ids skip normalization and reach the retained after-validator, whose message now carries the ids to aid the provider's repair attempt (`presence cannot enter and exit the same character: nika rel (enter id=31, exit id=32)`). Mixed payloads normalize the compatible names and still reject on the conflict — the response needs re-encoding regardless.

Tests: the reviewer's exact scenario (baseline-present 32, enter 31 / exit 32, casefold variants) plus same-id, one-sided-id, and mixed-payload regressions; focused suite 95 passed, full suite 2145 passed, 500 skipped.

**Round-2 note:** the fix is a ~60-line refinement of the before-validator with 1:1 frozen-scenario coverage including the reviewer's own repro; I verified the diff line-by-line (the normalizable set and the conflict detector are exact complements) rather than commissioning another full round. Schema identity is structurally unaffected — before-validators do not alter the emitted schema.

Merging under delegated authority.

PR #671, opened 2026-08-06
Bind the Correct Summary ID for Every Retrograde Vector Upsert

### LINKED ISSUE #665: Retrograde batch embedding writes only the last summary vector
## Summary

Batch embedding of dedicated Retrograde summaries stores a vector only for the
last summary in each batch, but stamps `embedding_generated_at` on every
requested summary. The public history-sync command then trusts the stamps,
reports every row as `already_present`, and cannot repair the missing vectors.

This leaves most Retrograde history absent from vector retrieval even though
the operator surface reports a fully embedded corpus.

## Public reproduction

On isolated slot 4, a public narrative acceptance declared and matured a new
character. Maturation job 15 produced four Retrograde summaries (IDs 21-24)
and its result manifest reported successful embedding results for all four.

The public sync surfaces then both reported no pending work:

```text
$ nexus --json retrograde-embed-history --slot 4
"dry_run": true
"embedding_pending_summary_ids": []
summary IDs 21, 22, 23, 24: "status": "already_present"

$ nexus --json retrograde-embed-history --slot 4 --execute
"dry_run": false
"embedding_pending_summary_ids": []
"embedding_results": []
```

PostgreSQL contradicts that public result:

```sql
SELECT count(*) FROM retrograde_summaries;
-- 24

SELECT count(*) FROM retrograde_summaries
WHERE embedding_generated_at IS NOT NULL;
-- 24

SELECT count(*) FROM retrograde_summary_embeddings_2560d;
-- 8

SELECT count(*)
FROM retrograde_summaries s
WHERE s.embedding_generated_at IS NOT NULL
  AND NOT EXISTS (
    SELECT 1
    FROM retrograde_summary_embeddings_2560d e
    WHERE e.summary_id = s.id
  );
-- 16
```

For the current four-summary batch, all four rows share the same
`embedding_generated_at` timestamp, but only summary ID 24 has a vector row.
The older seed data shows the same last-row-per-batch pattern.

## Expected behavior

- Every requested `(summary_id, active_model)` pair receives its own vector.
- A summary is stamped embedded only after all of its active-model rows exist.
- `retrograde-embed-history --execute` detects and repairs stamped rows whose
  vector rows are missing.

## Actual behavior and impact

- Only the last summary ID in each batch is used for every vector upsert.
- All summary rows are stamped successfully, making the loss look complete.
- Dedicated vector search joins on the embedding table, so the 16 missing rows
  cannot participate in semantic/vector retrieval. Text search may still find
  literal matches, but the intended Retrograde semantic recall is degraded.

## Likely cause

`nexus/agents/orrery/retrograde_embedding.py` builds `generated` by summary ID,
but the write loop iterates `generated.values()` and passes the stale
`summary_id` left behind by the preceding generation loop to every INSERT:

```python
for model_embeddings in generated.values():
    ...
    cursor.execute(..., (summary_id, model_name, embedding_value))
```

The loop should bind each key, for example
`for summary_id, model_embeddings in generated.items()`.

The repair surface also derives pending work only from
`retrograde_summaries.embedding_generated_at IS NULL` in
`plan_retrograde_summaries`, so it never notices stamped-but-missing rows.

## Acceptance criteria

- Bind the correct summary ID for every vector upsert in a multi-summary batch.
- Add a PostgreSQL-backed regression covering at least two summaries and all
  active models; assert one vector row per `(summary_id, model)` before stamps
  are committed.
- Make `retrograde-embed-history` detect and repair stamped summaries missing
  any configured active-model vector.
- Provide a safe backfill path for existing stamped-but-missing rows.
- Verify every repaired summary can participate in dedicated vector search.

## Deduplication

Checked all 169 open/closed repository issues and 100 recent merged PRs, plus
issue searches for `retrograde summary embedding`, `embedding_generated_at`,
and `retrograde_summary_embeddings`; no matching report was found.

Codex — GPT-5

pythagorakase:
Closes #665.

## Root Cause (Verified)
Classic leaked loop variable: the generation loop binds `summary_id` per summary, then the vector write loop iterated `generated.values()` and reused the name — frozen at the LAST requested id — so every INSERT upserted onto one row while ALL requested summaries were stamped `embedding_generated_at`. The stamp-trusting repair surface then reported every damaged row `already_present`, making the loss self-concealing. Live damage on slot 4: 16 of 24 summaries stamped with no vector row.

## Fix
- **Binding**: `for summary_id, model_embeddings in generated.items():` — the one-line correct change.
- **Detection**: `plan_retrograde_summaries` now checks each stamped summary for a vector row per configured active MEMNON model (per-dimension table via `to_regclass` — a lazily absent table counts as missing without being created during dry-run). Damaged rows report `status: "stamped_missing_vectors"` with an explicit `missing_embedding_models` list, join `embedding_pending_summary_ids`, and get a `summaries_stamped_missing_vectors` counter.
- **Repair**: `retrograde-embed-history --execute` re-embeds deficient summaries through the same genuine path (idempotent upsert) and refreshes stamps only after all rows exist. One deliberate semantics change, recorded: `--execute` no longer gates on `retrieval_settings.embed_after_apply` — that toggle governs *automatic* post-apply embedding, and gating the dedicated repair command on it would make repair silently impossible with the toggle off.
- This is the backfill path; the live backfill of slot 4's 16 damaged rows runs post-merge and will be reported on #665.

## Tests
New `tests/test_orrery/test_retrograde_embedding_pg.py` (requires_postgres, disposable `qa665_*` clone of NEXUS_template, dropped after; real configured MEMNON models — no mocks):
1. Genuine two-summary batch → one vector row per `(summary_id, model)` for every summary, stamps set.
2. Old-bug damage reconstruction → dry-run reports `stamped_missing_vectors` + stays strictly read-only (stamp unchanged, absent tables not created) → `--execute` repairs every active-model row and bumps the stamp.
3. Repaired summary visible through MEMNON's genuine dedicated vector-search join.

Existing fake-cursor coverage extended additively for the two new tagged read queries. Suites: 51 passed (unit) + 3 passed (pg, 5.5s). mypy/black/flake8 clean on all touched files.

https://claude.ai/code/session_01JyugzKYfkonxthNs1EHRZD

### DIFF
--- a/nexus/agents/orrery/retrograde_embedding.py
+++ b/nexus/agents/orrery/retrograde_embedding.py
@@ -33,6 +33,19 @@ def _load_memnon_settings() -> dict[str, Any]:
     return settings
 
 
+def active_memnon_embedding_model_dimensions() -> dict[str, int]:
+    """Return configured active MEMNON model names and vector dimensions."""
+    configured_models = _load_memnon_settings().get("models", {})
+    active_models = {
+        str(name): int(config["dimensions"])
+        for name, config in configured_models.items()
+        if config.get("is_active", True)
+    }
+    if not active_models:
+        raise RuntimeError("No active MEMNON embedding models are configured")
+    return active_models
+
+
 def embed_retrograde_summaries(
     dbname: str,
     summary_ids: Sequence[int],
@@ -86,14 +99,7 @@ def embed_retrograde_summaries(
         raise RuntimeError(f"Retrograde summaries not found in {dbname}: {missing_ids}")
 
     memnon_settings = _load_memnon_settings()
-    configured_models = memnon_settings.get("models", {})
-    configured_active_models = [
-        name
-        for name, config in configured_models.items()
-        if config.get("is_active", True)
-    ]
-    if not configured_active_models:
-        raise RuntimeError("No active MEMNON embedding models are configured")
+    configured_active_models = list(active_memnon_embedding_model_dimensions())
 
     embedding_manager = EmbeddingManager(settings=memnon_settings)
     model_names = embedding_manager.get_available_models()
@@ -123,7 +129,7 @@ def embed_retrograde_summaries(
     with get_connection(dbname, dict_cursor=True) as conn:
         with conn.cursor() as cursor:
             ensured_tables: dict[int, str] = {}
-            for model_embeddings in generated.values():
+            for summary_id, model_embeddings in generated.items():
                 for model_name, embedding in model_embeddings:
                     dimensions = len(embedding)
                     table_name = ensured_tables.get(dimensions)
--- a/nexus/agents/orrery/retrograde_persistence.py
+++ b/nexus/agents/orrery/retrograde_persistence.py
@@ -524,6 +524,7 @@ def _build_plan(
         "summaries_would_insert",
         "summaries_inserted",
         "summaries_already_present",
+        "summaries_stamped_missing_vectors",
         "summaries_blocked",
     ):
         counters.setdefault(key, 0)
@@ -1851,6 +1852,7 @@ def plan_retrograde_summaries(
         event_sources = _load_persisted_retrograde_event_sources(cur)
 
     rows: list[dict[str, Any]] = []
+    active_embedding_model_dimensions: Optional[dict[str, int]] = None
     for source in event_sources:
         source_status = str(source.get("source_status") or "persisted")
         world_event_id = _optional_int(source.get("world_event_id"))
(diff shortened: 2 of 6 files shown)


pythagorakase:
Independent Sol review verdict: **CHANGES_REQUIRED**, one P2 — confirmed against code and fixed.

**[P2] planner `RuntimeError` bypassed the CLI's structured-error boundary** — with the loadable all-models-inactive configuration, the new detection call raised `active_memnon_embedding_model_dimensions()`'s RuntimeError inside `plan_retrograde_summaries`, outside the `except ValueError` plan boundary, so even a `--json` dry-run emitted a traceback — while the pre-existing execute path catches the same error class structurally. **Fixed**: the plan boundary now catches `(ValueError, RuntimeError)`, mirroring the execute boundary in the same function, with a regression driving `run_retrograde_embed_history` through the established FakeApplyConnection pattern and asserting the exact structured payload.

Suites re-run: 52 passed (unit) + 3 passed (pg, genuine models). Reviewer independently confirmed both acceptance suites green. Merging.

pythagorakase:
Merging with CI noted as degraded-infra, not red: this branch's drift job died at **action-download resolution (Service Unavailable, three retries)** before any code ran — its logic already executed locally via the validate-config pre-commit hook on both commits — and today's Actions queue is starving downstream runs generally. Operative gates passed: independent Sol review (P2 confirmed and fixed above), 52-test unit suite + 3 genuine-model pg regressions green.

PR #687, opened 2026-08-09
Delete Stamped Vectorless Project-Start Retrograde Summary Orphans

### LINKED ISSUE #672: Legacy Project-Start Retrograde Summaries Are Invisible to the Embed-History Planning Surface
## What

Slots 4 and 5 each carry one `retrograde_summaries` row (summary 9 → world_event 23, `court_patron_started`, identical lineage) that is stamped `embedding_generated_at` but has no vector row and can never be seen by `retrograde-embed-history`: `_load_persisted_retrograde_event_sources` excludes `PROJECT_STARTED_EVENT_TYPES` from the planning surface by documented design (mechanical projection provenance, not generated history prose), and this row predates that exclusion.

Found during #665's live backfill (PR #671): the in-surface damage class is now zero fleet-wide, but this row is structurally outside the repair surface.

## Why It's a Decision, Not a Bug

Current code would never create a summary for a project-start event, so the row is policy-noncompliant legacy data. Two clean resolutions:

1. **Delete** the orphan rows (a one-time data-hygiene migration matching current policy) — summaries for mechanical provenance events simply should not exist; or
2. **Include** stamped legacy rows in the planning surface regardless of event-type exclusion (the exclusion would then gate only *new* summary creation, not repair of existing rows).

Option 1 matches the exclusion's stated rationale; option 2 preserves whatever retrieval value the legacy prose has. Owner's call — the rows were deliberately left untouched.

## Evidence

```sql
SELECT s.id, s.world_event_id FROM retrograde_summaries s
WHERE s.embedding_generated_at IS NOT NULL
  AND NOT EXISTS (SELECT 1 FROM retrograde_summary_embeddings_2560d e WHERE e.summary_id = s.id);
-- save_04: (9, 23)   save_05: (9, 23)
```

world_event 23: `source='retrograde'`, `event_type='court_patron_started'`, payload carries `retrograde_event_ref`.

https://claude.ai/code/session_01JyugzKYfkonxthNs1EHRZD

pythagorakase:
Closes #672.

One-time hygiene migration (101) deleting `retrograde_summaries` rows that are (a) anchored to project-start provenance events, (b) ironman-stamped, and (c) vectorless in every `retrograde_summary_embeddings_*` dimension table — the legacy pre-exclusion rows found during #665's backfill (summary 9 → world_event 23 in save_04/save_05).

**Decision rationale** (recorded per the issue's "owner's call" framing, decided under delegated authority): deletion matches the exclusion's stated policy — and the rows are stamped-but-vectorless, so they were never reachable by retrieval; the "preserve retrieval value" argument is moot.

- Dynamic discovery of all current embedding dimension tables; idempotent rerun deletes zero.
- Planning-surface comment updated at the exclusion site.
- PG test covers deletion, protected controls (vector-backed / non-project-start / unstamped), and idempotency.

Fleet application via `scripts/migrate.py --all` post-merge.

### DIFF
--- a/migrations/101_delete_project_start_summary_orphans.sql
+++ b/migrations/101_delete_project_start_summary_orphans.sql
@@ -1,0 +1,51 @@
+-- Remove stamped, vectorless summaries for excluded project-start provenance.
+
+DO $$
+DECLARE
+    embedding_table record;
+    deleted_count bigint := 0;
+BEGIN
+    CREATE TEMPORARY TABLE migration_101_summary_candidates (
+        summary_id bigint PRIMARY KEY
+    ) ON COMMIT DROP;
+
+    INSERT INTO migration_101_summary_candidates (summary_id)
+    SELECT rs.id
+    FROM retrograde_summaries AS rs
+    JOIN world_events AS we ON we.id = rs.world_event_id
+    WHERE we.event_type IN (
+        'relocation_plan_started',
+        'recruit_ally_started',
+        'build_venture_started',
+        'pursue_romance_started',
+        'court_patron_started',
+        'seek_redemption_started'
+    )
+      AND rs.embedding_generated_at IS NOT NULL;
+
+    FOR embedding_table IN
+        SELECT table_schema, table_name
+        FROM information_schema.tables
+        WHERE table_schema = current_schema()
+          AND table_type = 'BASE TABLE'
+          AND table_name ~ '^retrograde_summary_embeddings_[0-9]+d$'
+        ORDER BY table_name
+    LOOP
+        EXECUTE format(
+            'DELETE FROM migration_101_summary_candidates AS candidate '
+            'USING %I.%I AS embedding '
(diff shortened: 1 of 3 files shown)


pythagorakase:
**Sol review verdict: CHANGES_REQUIRED (1×P2) — rebutted, merging.** The P2 ([tests gated behind NEXUS_RUN_POSTGRES] skip in default runs) is accurate as an observation but is the repo's enforced convention, not this PR's defect: tests/conftest.py:37 requires the `requires_postgres` mark on any test touching psycopg2, and all sibling `*_pg.py` suites carry the same gate. The migration's acceptance gate was the opt-in PG run, executed twice independently (worker + reviewer + coordinator): deletion, protected controls, and idempotency all pass. Sol verified the event-type set matches `PROJECT_STARTED_EVENT_TYPES`, the dimension-table regex covers future dims, and no scope strays. Repo-wide PG-in-CI is a real but separate conversation.

Issue #677, opened 2026-08-08
Add actor-owned experiential memory above Epistemics


pythagorakase:
## Summary

Add a durable, actor-owned representation of subjective experience without mixing it with canonical events, claims, or awareness.

NEXUS can record what happened and which account an entity possesses. It cannot yet record how a character experienced an event.

## Current evidence

- Closed #495 intentionally established claims and awareness first and identified episodic memory as an additive later layer.
- Closed #477 made knowledge mechanically consequential without introducing subjective memory.
- migrations/076_claims_awareness.sql, 080_epistemics_knowers.sql, 090_claim_accounts.sql, and 092_claim_distortion_depth.sql provide the canonical-event, account, possession, and distortion foundation.
- docs/npc_knowledge_mechanics_research.md separates incident, account, awareness, recall/disclosure, and consequences.
- MEMNON is authorial narrative retrieval, not an actor-private store.
- The MGO 4.0 RC2 architecture review identified actor-relative experience as the largest useful gap.

## Scope

Define and implement an actor-owned experience primitive that:

- Has stable actor and source provenance.
- Distinguishes direct participation or verified observation from being told an account.
- Can carry subjective properties such as emotion, salience, location, type, and controlled tags.
- Can optionally produce a concise subjective rendering without changing canonical truth.
- Survives replay and exposes enough version information to explain regenerated material.
- Fails safely if optional rendering is unavailable.

The issue does not prescribe a table, event projection, rendering model, embedding system, or synchronous versus asynchronous implementation.

## Design questions

- What source granularity is correct: one event, related events, one scene, or a bounded episode?
- Which role or perception evidence qualifies an actor for a first-hand experience?
- Should told or granted acquisition experiences be automatic or created only when useful?
- Which fields must be deterministic at commit, and which can be derived later?
- Are subjective renderings immutable records, versioned projections, or replaceable caches?
- When is semantic indexing justified?
- How do rollback, branch replacement, and model-version changes affect derived text?
- Can a first slice land without any model rendering and prove the ownership/provenance substrate first?

## Acceptance criteria

- [ ] An absent and uninformed actor receives no experience.
- [ ] A verified participant or witness can receive a first-hand experience linked to its source.
- [ ] A told actor can receive an acquisition experience linked to the exact delivered account, never a false witness memory.
- [ ] Two witnesses can hold different subjective experiences without changing the canonical event.
- [ ] Generated text cannot mint claims, awareness, participants, or world state.
[...]

PR #689, opened 2026-08-09
Retire the Observations Wire Arm

### LINKED ISSUE #674: Stop silently dropping character extra_observations at commit
## Problem

The structured-output contract accepts named character observations and carries them through hydration and the incubator. Both accepted-turn commit implementations then ignore the field.

A turn can therefore commit successfully while silently losing part of its declared character-state update.

## Evidence

Verified on main at 956f22d2:

- nexus/agents/logon/skald_wire.py:296-334 defines observations as a substantive character update.
- nexus/agents/logon/skald_wire.py:801-814 hydrates it into CharacterStateUpdate.extra_observations.
- nexus/agents/logon/apex_schema.py:379-418 describes the field as observations for characters.extra_data.
- nexus/api/lore_adapter.py:263-267 serializes the complete state-update model into the incubator.
- tests/test_api/test_lore_adapter.py:149-230 proves that observations survive that boundary.
- nexus/api/commit_handler.py:353-434 applies character scalars and Orrery tags but never reads extra_observations.
- nexus/api/commit_handler_sync.py:801-869 has the same omission.

No production persistence consumer was found.

Closed issue #555 deliberately restored observations to the authorial wire. The missing step is therefore downstream persistence or an explicit retirement decision.

## Impact

- The provider-facing contract promises a state update that accepted commits do not preserve.
- A caller cannot distinguish a durable observation from discarded output.
- Future character-memory work could mistakenly build on a non-durable field.
- Fixing only one commit path would preserve inconsistent behavior.

## Design space

Choose and document one explicit contract. Possible approaches include:

1. Persist observations under a defined namespace in characters.extra_data.
2. Persist them in a dedicated observation, event, or future character-experience store with accepted-chunk provenance.
3. Retire the field from the wire and canonical schema and reject non-empty observations before incubation.

If observations remain supported, the design must decide whether duplicate keys replace, append, or conflict; whether observations are canonical facts or subjective reports; and how replay and provenance represent them.

This issue does not require a particular storage model.

## Acceptance criteria

- [ ] A non-empty observation cannot pass an accepted commit and then disappear silently.
- [ ] Sync and async commit paths implement the same contract.
- [ ] If persisted, the write is atomic with chunk acceptance and records source provenance.
- [ ] If persisted, retry and duplicate semantics are deterministic.
- [ ] If persisted, rejection or rollback leaves no observation state behind.
- [ ] If persisted, replay or reconstruction behavior is defined and tested.
- [ ] If retired, provider and canonical schemas no longer advertise the field.
- [ ] If retired, non-empty legacy input fails explicitly before commit.
- [ ] Tests cover wire → incubator → accepted commit, not only serialization.
- [ ] Tests cover both commit implementations and rollback behavior.
- [ ] Prompts, schema descriptions, and developer documentation match the selected contract.

## Non-goals

- Designing the complete actor-owned episodic-memory system.
- Recovering previously discarded observations when no durable source remains.
- Treating generated observations as claims or canonical truth without a separate policy decision.

— Codex (GPT-5)

pythagorakase:
Closes #674.

Verified premise (independently by both reviews): `observations` → `extra_observations` was hydrated (`skald_wire.py:809`) and then ignored by **both** commit paths — a silent drop inviting false confidence that NPC observations became durable.

Resolution per the Arachne seq-24 ruling: with the full actor-experience layer (#677) approved as the proper home, the arm is **retired** rather than given persistence. Removal spans wire schema, `NamedObservation`, hydration, tag validation, and dead tests; `extra="forbid"` grammar kept strict throughout. No DB column existed, so no migration.

Wire-weight side effect: strict Gaia schema now 13,428 B / 3,050 tok.

### DIFF
--- a/nexus/agents/logon/apex_schema.py
+++ b/nexus/agents/logon/apex_schema.py
@@ -59,15 +59,6 @@ class Coordinates(BaseModel):
     model_config = ConfigDict(extra="forbid")
 
 
-class NamedObservation(BaseModel):
-    """Strict key/value observation entry for JSONB-style side notes."""
-
-    key: str = Field(min_length=1, description="Observation key or label")
-    value: str = Field(description="Observation value as concise prose")
-
-    model_config = ConfigDict(str_strip_whitespace=True, extra="forbid")
-
-
 class FactionStanceChange(BaseModel):
     """Strict faction stance change entry."""
 
@@ -78,7 +69,7 @@ class FactionStanceChange(BaseModel):
 
 
 def _stringify_structured_value(value: Any) -> str:
-    """Render arbitrary legacy observation values into strict-schema prose."""
+    """Render arbitrary legacy structured values into strict-schema prose."""
 
     if value is None:
         return ""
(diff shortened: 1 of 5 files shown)


pythagorakase:
**Sol review verdict: MERGE_READY, no findings.** Reviewer independently re-verified the no-consumer premise, confirmed zero dangling references across production/prompts/docs, and confirmed strict schemas now reject the retired fields with no fixture shims. 152 passed.

Issue #686, opened 2026-08-09
Move Bleed Density to Config and Measure Bleed Uptake at Commit


pythagorakase:
## Summary

Two coupled changes to the Orrery bleed channel:

1. Replace the prompt-prose density rule ("Roughly 10–20% of off-screen activity should bleed into the narrative") with a config float in `nexus.toml`, drawn at selection time.
2. Measure **uptake** — whether an offered bleed actually reached the accepted prose — with the deterministic oracle already latent in the design.

From the MGO 4.0 RC2 review (`temp/skyrimnet_mgo40_report.md` R5; `temp/skyrimnet_findings/cmp_gm_vs_gaia.md` §8.2).

## Evidence

- SkyrimNet moved pacing out of the model into a template-time dice roll (silence hidden from the model at a configured probability; verified firing at 52% across 52 production prompts). Its cautionary counter-result: a prose-level density rule ("roughly 1 in 4") was measured at 56% actual — prose rules lose to prompt gravity.
- NEXUS's density rule is a sentence in a prompt: untunable and unmeasured, against the standing directive that adjustable values belong in `nexus.toml`.
- The uptake oracle is free: the bleed instruction already requires "include that exact name at least once in the prose," so offered-vs-used is a substring check against the accepted chunk.
- The offer side is instrumented (`record_bleed_offers`, `bleed.py:362`, gated by `offer_count < 3`); **uptake** is the unmeasured half.

## Landing Spots

- Density: `nexus.toml [orrery.bleed]` float; selection-time draw in `load_bleed_candidates`.
- Uptake: post-commit step in `nexus/api/commit_handler_sync.py` (near the Step 8.5 Orrery hook); migration adding `used_chunk_id` / `use_count` to `orrery_resolutions`.
- Consumers: bleed-uptake rate in the QA kit (`scripts/qa_shift`); deprioritize high-offer/zero-uptake candidates in `load_bleed_candidates`.
- Extension: score n-gram overlap between offered stub text and accepted prose to test the register-amplification hypothesis (drift-study metrics already exist) before adopting direction-form (2–6 word) stubs in the bleed menu.

pythagorakase:
**Loom ruling (Arachne seq 24): density default = 0.20.** Ship the config float at `[orrery.bleed]` with 0.20 as the committed default (was prompt-prose "10–20%"). Uptake measurement proceeds as scoped.

PR #696, opened 2026-08-09
Persist the Pass-1/Pass-2 Baseline Across Turns via Incubator Staging

### LINKED ISSUE #675: Make the LORE Pass-1/Pass-2 baseline effective across narrative turns
## Problem

The main narrative route creates and closes a fresh LORE instance for every turn. Pass 2 runs at the start of that instance, while Pass 1 stores its baseline only after generation in instance memory.

The baseline produced by turn N therefore cannot be available to Pass 2 on turn N+1 through /api/narrative/continue.

## Evidence

Verified on main at 956f22d2:

- nexus/api/narrative_generation.py:195-257 constructs one LORE instance for the request and closes it in finally.
- This lifecycle is intentional resource cleanup after #401.
- nexus/agents/lore/utils/turn_cycle.py:328-343 invokes Pass 2 during user-input processing.
- nexus/agents/lore/utils/turn_cycle.py:1534-1578 creates the Pass-1 baseline only during response integration.
- nexus/memory/context_state.py:89-126 stores the baseline in the ContextStateManager instance.
- nexus/memory/manager.py:640-673 reports baseline_available=False when a fresh manager has no context and transition.
- nexus/api/lore_adapter.py:123-185 does not place the baseline or transition in the incubator.
- The separate storyteller API caches a LORE singleton in nexus/api/storyteller.py:252-271, but that is not the main route.

Existing memory-manager tests reuse one manager. They do not exercise the main production lifecycle.

## Impact

- Cross-turn divergence retrieval is effectively unavailable on the main route.
- Diagnostics can describe a Pass-1 baseline that is discarded before it can support the next turn.
- Unit tests can pass while normal requests receive baseline_available=False.
- A process restart is indistinguishable from normal request teardown because neither path restores a baseline.

The first continuation after bootstrap can legitimately lack a prior baseline. Later accepted turns should not lose one only because the API uses a fresh object.

## Design space

Possible approaches include:

1. Persist a compact, versioned baseline and transition against the accepted chunk.
2. Store provisional baseline material with the incubator and promote it only when the turn is accepted.
3. Reconstruct the required baseline deterministically from accepted canonical data at request start.
4. Redesign or retire the cross-turn Pass-1/Pass-2 contract if a simpler per-turn retrieval model provides the intended behavior.

Do not solve this by keeping the main-route LORE object alive indefinitely. That conflicts with the current resource-lifecycle boundary and does not handle restart or slot isolation.

The design should determine the minimum safe persisted material. It need not store the complete assembled prompt.

## Acceptance criteria

- [ ] A route-level test uses distinct LORE instances for consecutive main-route turns.
- [ ] After turn N is accepted, turn N+1 loads or rebuilds the correct baseline before handle_user_input runs.
- [ ] The test observes baseline_available=True and verifies meaningful Pass-2 behavior.
- [ ] The first turn without an eligible predecessor reports baseline absence explicitly.
- [ ] A rejected or regenerated provisional turn cannot become the next accepted baseline.
- [ ] Baselines cannot cross slot, parent-chunk, or canonical-history boundaries.
- [ ] A process restart does not change the result for the same accepted parent.
- [ ] Missing, incompatible, or stale baseline data produces an observable rebuild or explicit unavailable state.
- [ ] Provider or model changes cannot restore invalid token-budget assumptions.
- [ ] Tests retain coverage for simple-choice skipping, divergence retrieval, and token accounting.
- [ ] The separate singleton storyteller route is documented or tested independently; its lifecycle is not assumed to solve the main route.

## Non-goals

- Keeping one global LORE singleton for all slots or sessions.
- Changing MEMNON ranking or embedding models.
- Making provisional storyteller output canonical.
- Unifying the two API routes unless the selected implementation requires it.

— Codex (GPT-5)

pythagorakase:
Closes #675.

Design settled by independent consult (Sol, read-only) with coordinator concurrence: **option 2 — incubator staging with promote-on-accept**. The consult also proved option 3 (deterministic reconstruction) impossible: the exact prior retrieval selection can't be reproduced later because the corpus itself changes as chunks become embedded/ironman.

- `Pass2BaselineV1` persists only the cross-turn operational contract: exact post-trim memory identities, token accounting/remaining budget, config fingerprint. Assembled context is never serialized; prose reloads from the accepted chunk.
- Staged baselines are **deliberately unbound** (`parent_chunk_id=None`); acceptance binds them to the actual `RETURNING id` inside the commit transaction — the consult's named sharpest failure mode (sparse ids under regeneration) is structurally excluded.
- One hydration path for restart and ordinary teardown; `baseline_available=True` from the first post-migration accepted turn.
- Fail-loud everywhere: invalid staged state raises; historical tails get an administrative error naming the explicit stamp helper (`scripts/stamp_lore_pass_baseline.py`) — no inferred retrieval sets, no silent bootstrap inference.
- Rollback invariants tested: rejection, regeneration replace, acceptance failure, FK cascade.

96 unit + 7 PG tests green through the genuine fresh-instance-per-turn lifecycle.

### DIFF
--- a/migrations/107_lore_pass_baselines.sql
+++ b/migrations/107_lore_pass_baselines.sql
@@ -1,0 +1,26 @@
+-- Migration 107: durable Pass-1/Pass-2 baselines staged with provisional prose.
+
+ALTER TABLE incubator
+    ADD COLUMN IF NOT EXISTS lore_pass_baseline JSONB;
+
+COMMENT ON COLUMN incubator.lore_pass_baseline IS
+    'Unbound versioned Pass-2 baseline staged atomically with provisional storyteller prose and discarded on rejection.';
+
+CREATE TABLE IF NOT EXISTS lore_pass_baselines (
+    chunk_id BIGINT PRIMARY KEY
+        REFERENCES narrative_chunks(id) ON DELETE CASCADE,
+    schema_version INTEGER NOT NULL CHECK (schema_version > 0),
+    payload JSONB NOT NULL CHECK (jsonb_typeof(payload) = 'object'),
+    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
+);
+
+COMMENT ON TABLE lore_pass_baselines IS
+    'Accepted versioned Pass-2 baselines, promoted in the same transaction as their narrative chunk.';
+COMMENT ON COLUMN lore_pass_baselines.chunk_id IS
+    'Actual accepted narrative_chunks id returned by insertion; also the parent identity required during next-turn hydration.';
+COMMENT ON COLUMN lore_pass_baselines.schema_version IS
+    'Validated baseline wire-schema version duplicated from payload for indexed administrative inspection.';
+COMMENT ON COLUMN lore_pass_baselines.payload IS
+    'Validated bound Pass-2 baseline containing exact typed memory identities, token accounting, and compatibility fingerprint.';
+COMMENT ON COLUMN lore_pass_baselines.created_at IS
+    'Database time at accepted-chunk promotion.';
--- a/nexus/agents/lore/lore.py
+++ b/nexus/agents/lore/lore.py
@@ -403,6 +403,13 @@ class LORE:
         )
 
         try:
+            if parent_chunk_id is not None:
+                if self.memory_manager is None:
+                    raise RuntimeError(
+                        "Pass-2 baseline hydration requires the memory manager"
+                    )
+                self.memory_manager.restore_pass2_baseline(parent_chunk_id)
+
             # Phase 1: User Input Processing
             self.current_phase = TurnPhase.USER_INPUT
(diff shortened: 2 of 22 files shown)


pythagorakase:
**Review round addressed, point by point:** (P2 bypassed route) the lifecycle test now drives `generate_narrative_async()` twice — two genuine fresh LORE instances with only retrieval and the provider stubbed — proving hydration, Pass-2 exclusion, export, staging, acceptance, and restoration on the real continuation route. (P2 async parity) async failure injection after chunk+baseline insertion proves atomic rollback with the replacement incubator row surviving; async regeneration-replace covered. Also carries a cross-branch semantic fix (the #686 manifest test now supplies the baseline this branch makes required). Coordinator verified the combined suites green post-rebase onto the repaired main. Merging under the straightforward-fix-commit rule.

Issue #680, opened 2026-08-08
Add an end-to-end character cognition inspector


pythagorakase:
## Summary

Extend the existing Orrery developer surface into an anchor-aware character-cognition inspector.

For one character and scene anchor, provide one trace from source event through possession, recall, disclosure, and prompt exposure or rejection.

This is a focused child of the broader developer-mode work in #625, not a second unrelated dashboard.

## Current evidence

- nexus/agents/orrery/audit.py already exposes possessed claims and acquisition provenance.
- ui/client/src/pages/dev-orrery/EntityAudit.tsx already renders claim summaries, provenance, and acquisition time.
- nexus/agents/orrery/knowledge_surfacing.py has a strong spoiler boundary but no unified explanation of selection and truncation.
- Closed #477 added knower-gated audit information.
- Closed #495 made provenance and queryable ignorance core requirements.
- #677, #678, and #679 add experience, recall/disclosure, and broader producer coverage that need equivalent diagnostics.
- The MGO review found a unified event, memory, retrieval, prompt, request, and timing dashboard valuable for debugging information boundaries.

## Scope

For a selected character and anchor, expose:

- Exact possessed accounts and acquisition/source chains.
- Actor-owned experiences and source events when #677 exists.
- Recall candidates, score components, thresholds, and truncation when #678 exists.
- Disclosure results and blocking reasons.
- Material actually exposed to the Storyteller or an ambient renderer.
- Relevant async job state, timeline or generation identity, retries, and stale-result rejection.
- Effective configuration, caps, and producer coverage.

The implementation may extend the current audit endpoint and UI or add a dedicated cognition read model. It need not persist every decision if the same explanation can be reconstructed safely at an anchor.

## Design questions

- Should this extend /api/dev/orrery or use a dedicated cognition-trace endpoint?
- Which decisions need durable traces, and which can be recomputed?
- How can a privileged canonical comparison remain separate from actor-facing data?
- What redaction and access boundary protects latent secrets?
- How should large candidate sets and long source chains be paged?
- Which view best explains shared-budget starvation or truncation?
- How should model-dependent decisions display model, prompt version, and inputs without exposing chain of thought?
- Which panels can land before the underlying cognition features?

## Acceptance criteria

- [ ] A developer can select an actor and anchor and trace an account through immediate and root provenance.
- [ ] A fixture with contradictory sibling accounts shows only the possessed account in the actor-view path.
- [ ] Every recall candidate has an included or excluded result and machine-readable reason.
[...]

PR #702, opened 2026-08-09
Add Entitlement-First Recall and a Mechanical Disclosure Gate

### LINKED ISSUE #678: Add entitlement-first NPC recall and disclosure
## Summary

Separate three questions for a speaking actor:

1. What information does the actor possess?
2. What does the actor recall in this scene?
3. What is the actor willing or permitted to disclose to this audience?

Possession remains the hard boundary. Contextual retrieval can rank only material already owned by that actor.

## Current evidence

- Closed #495 defined recall and disclosure as a separate layer above awareness.
- nexus/agents/orrery/knowledge_surfacing.py exposes only exact possessed accounts and excludes latent secrets and unpossessed siblings.
- That query currently selects recent eligible rows under one shared max_entries cap for all present characters.
- nexus.toml sets the shared cap to 12.
- MEMNON performs global narrative retrieval and must not determine actor entitlement.
- docs/npc_knowledge_mechanics_research.md states the governing distinction: retrieval is not knowledge.
- #677 adds actor-owned subjective experience as another possible source, but recall can begin against existing claims.

## Scope

Add an explainable actor-scoped pipeline that:

- Applies possession, anchor, and timeline eligibility before ranking.
- Can rank claims and actor-owned experiences for current relevance.
- Supports mandatory or reserved treatment for critical newly acquired material.
- Represents recall strength separately from permanent awareness.
- Makes disclosure a separate audience- and context-dependent decision.
- Records why each candidate was included, excluded, or suppressed.
- Prevents one actor from consuming the complete shared prompt budget.

This issue does not select a ranking formula, embedding model, generated-query model, disclosure rules engine, or prompt shape.

## Design questions

- Which signals should contribute to recall: current input, scene semantics, recency, severity, involvement, emotion, place, or authored priority?
- Should ranking use raw input, a generated query, or both?
- How should per-actor budgets and shared-scene material interact?
- Is recall decay computed, persisted, or event-driven?
- Should disclosure filter material, annotate it as unavailable for speech, or produce typed behavioral constraints?
- Which disclosure decisions must be deterministic, and where can model judgment help?
- How should silence, deception, partial answers, and “I remember but will not tell you” be represented?
- What measurement must justify new embeddings or an extra query-generation model?

## Acceptance criteria

- [ ] No unpossessed claim, sibling account, experience, or latent secret enters an actor's eligible set.
- [ ] A distorted-account recipient receives only the delivered account.
- [ ] Per-actor budgeting prevents one character's recent history from starving all others.
- [ ] Recall strength can decline without deleting or altering awareness and provenance.
- [ ] A character can recall a secret while disclosure prevents or constrains revelation.
- [ ] Repeated evaluation is deterministic or records every random or model-dependent input.
- [ ] Rank components, truncation, and disclosure reasons are available to diagnostics.
- [ ] Global MEMNON output cannot bypass the ownership filter.
- [ ] Tests cover absent witnesses, contradictory accounts, time skips, acquisition sources, and possession-with-refusal.
- [ ] A retrieval bakeoff compares simple raw-input ranking with any proposed query-generation or reranking call before adding model latency.
- [ ] The prompt boundary preserves exact delivered-account and latent-secret protections.

## Non-goals

- Using similarity search to establish knowledge.
- Deleting facts to simulate forgetting.
- General output censorship.
- Universal cognitive or credibility simulation.
- Selecting a new vector database before measurement.

## Dependencies

This can begin against existing claims and awareness. #677 can add actor-owned experiences as another eligible source later.

The cognition inspector planned under #625 should consume these explanations before broad rollout.

— Codex (GPT-5)

pythagorakase:
Closes #678. Loom-ruled: MECHANICAL — both layers, with logging.

Three questions, three layers, in strict order:

1. **Eligibility (hard boundary)**: possessed accounts + owned experiences only, anchor/timeline-valid. Global MEMNON structurally cannot enter (semantic fit computes from stored chunk/experience vectors only; a missing vector gets a configured trace-visible score rather than expanded entitlement).
2. **Recall ranking (deterministic)**: configurable component weights (`[orrery.recall]`), recall-strength decay as a score modifier that never mutates possession, a reserved mandatory slice for critical newly-acquired accounts, and a per-character budget preventing any actor from consuming the shared knowledge cap.
3. **Disclosure (separate, audience-aware)**: claim scope + directed valence + faction status edges/obligations; suppression is logged with reasons and excludes from the prompt while the character demonstrably still knows.

Migration 106: retention-bounded recall trace (FIFO-pruned per config) — every candidate's include/exclude/suppress decision with score components, the substrate #680's inspector reads. `claim_id` deliberately FK-free (audit-only; replay-safe).

Existing knowledge-surfacing spoiler tests pass unmodified. Zero LLM calls anywhere in the pipeline.

### DIFF
--- a/migrations/106_recall_trace.sql
+++ b/migrations/106_recall_trace.sql
@@ -1,0 +1,55 @@
+-- 106_recall_trace.sql
+-- Per-turn entitlement-first recall and disclosure decisions.
+
+CREATE TABLE IF NOT EXISTS orrery_recall_trace (
+    id                    bigserial PRIMARY KEY,
+    turn_id               text NOT NULL CHECK (btrim(turn_id) <> ''),
+    anchor_chunk_id       bigint NOT NULL REFERENCES narrative_chunks(id),
+    character_entity_id   bigint NOT NULL REFERENCES entities(id),
+    candidate_kind        text NOT NULL
+                              CHECK (candidate_kind IN ('claim', 'experience')),
+    candidate_id          bigint NOT NULL CHECK (candidate_id > 0),
+    claim_id              bigint,
+    decision              text NOT NULL
+                              CHECK (decision IN (
+                                  'included', 'excluded', 'suppressed'
+                              )),
+    reason                text NOT NULL CHECK (btrim(reason) <> ''),
+    mandatory             boolean NOT NULL,
+    score                 double precision NOT NULL,
+    score_components      jsonb NOT NULL,
+    created_at            timestamptz NOT NULL DEFAULT now(),
+    UNIQUE (turn_id, character_entity_id, candidate_kind, candidate_id)
+);
+
+CREATE INDEX IF NOT EXISTS ix_orrery_recall_trace_character_fifo
+    ON orrery_recall_trace (character_entity_id, id DESC);
+
+COMMENT ON TABLE orrery_recall_trace IS
+    'Compact per-turn audit of every eligible actor-owned recall candidate and its post-ranking disclosure decision; rows are retention-bounded per character by [orrery.recall] trace_rows_per_character and pruned oldest-id-first after each turn.';
+COMMENT ON COLUMN orrery_recall_trace.id IS
+    'Monotonic identity and FIFO ordering key for retention pruning.';
+COMMENT ON COLUMN orrery_recall_trace.turn_id IS
+    'LORE turn identity; retries upsert the same candidate decision.';
+COMMENT ON COLUMN orrery_recall_trace.anchor_chunk_id IS
+    'Accepted narrative anchor whose world clock, timeline, place, and audience governed the decision.';
+COMMENT ON COLUMN orrery_recall_trace.character_entity_id IS
+    'Speaking-eligible present character who owns the candidate material.';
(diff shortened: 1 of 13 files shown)


pythagorakase:
**Review round addressed, point by point:** (P1 embedding side channel) ruled and implemented at the granularity level — claims carry no vectors and now never touch chunk vectors; they rank on deterministic components with the semantic slot trace-marked `not_applicable_claim`; experiences keep semantic fit strictly via their own embeddings. The reviewer's scenario is a regression test: adding an unpossessed sibling secret to a shared source chunk (which changes that chunk's vector) leaves the possessed account's serialized rank, score, and components byte-identical. Claim-text embeddings noted in config comments as future work. (P2 anchor-as-query) recall consumes the single query representation hybrid retrieval already computes — now verified to include the actual current input — threaded through TurnContext with zero additional embedding calls, no anchor fallback (trace-visible skip), and the genuine-entry-point test varies only user input and observes ranking changes; the planted-anchor idiom is removed. Coordinator verified 87 green post-rebase and independently confirmed the disclosed test-fixture near-miss left save_05 untouched (no trace table, chunk count unchanged). Merging under the straightforward-fix-commit rule.

